PR #13, opened 2015-10-19
Issue 3 new

AbfrmBlr:


AbfrmBlr:
This takes care of everything on Server.
I shall create an issue for pass through integration tests.

coordinator-entity/server/src/main/java/org/terracotta/consensus/entity/server/LeaderElector.java
```diff
@@ -28,39 +34,147 @@
 public class LeaderElector<K, V> {
 
   private final PermitFactory<V> factory;
-  private final ConcurrentMap<K, BlockingDeque<V>> leaderQueues = new ConcurrentHashMap<K, BlockingDeque<V>>();
+  private final ConcurrentMap<K, ElectionQueue<V>> leaderQueues = new ConcurrentHashMap<K, ElectionQueue<V>>();
+  private final ConcurrentMap<K, ReentrantLock> locks = new ConcurrentHashMap<K, ReentrantLock>();
+  private DelistListener<K, V> listener;
```
alexsnaps: Didn't we say you'd refactor this to be a single `ConcurrentMap<K, SomeSmartDS>` ?

alexsnaps: Right, merge these things, you never clean up the locks... I _think_ all these have all the same lifecycle.

AbfrmBlr: Locks are per namespace, which anyways is the number of entities, so right now we are not clearing that. 
It would mean last client leaving ends up in removing that namespace.

AbfrmBlr:
Done

coordinator-entity/common/src/main/java/org/terracotta/consensus/entity/Nomination.java
```diff
@@ -45,6 +45,11 @@ public Nomination(final long id) {
   public boolean awaitsElection() {
     return id == AWAITS_ELECTION;
   }
+  
+  @Override
+  public boolean equals(Object obj) {
+    return id == ((Nomination) obj).id;
```
ljacomet: Proper equals implementation so that it does not leave itself open to class cast exceptions

PR #16, opened 2015-10-27
Sync with API change

mkscott:
API for InvocationBuilder has changed to include an acknowledgement for an invoke being sent from the client to the server.


terracotta-org:
Can one of the admins verify this patch?

PR #18, opened 2015-11-02
Issue #17 only fire to other clients

alexsnaps:


alexsnaps:
@AbfrmBlr I think this address the issue you had, I think message types can be cleanup now then... let me know if I'm wrong...

Issue #15, opened 2015-10-21
Allow client to take-over nomination


alexsnaps:
In the case of a client crash, while that client was being nominated (i.e. didn't `.accept` the `Nomination` yet), another client "taking over" has to be informed that's the situation. Otherwise it couldn't know how to deal with say a "create another entity"-usecase, where it should be considered a failure to "recreate" if it exists already, but since the client crashed _before_ finishing its work, we can't say a clean situation ever existed. This "in-between" state can't be perceived by clients. 

Adding a `Nomination.isContinuing()` or some methods along those lines, where the Server would set it to true, should the Nomination be "forwarded" to the next client in line, would let the client act accordingly...

PR #24, opened 2015-11-26
Carry-over nomination #15

AbfrmBlr:


alexsnaps:
You totally lost me here... is this for issue #15 ? Why is `Nomination` both used as a return value for `org.terracotta.consensus.entity.CoordinationEntity#runForElection` as well as the server to clients message ? 
Also, I may be reading this wrong, but the "continuation" of the election of a crashed client is triggered by the server when that client "delists" (but how is that wired in the `.disconnect`?), but that can't be done that way... that approach requires a client to be connected for that happen, right? But this needs to work for the single connected client crashing, and only then a new client connecting and taking over then. Now I'm not asking to add code to cover this case though!

AbfrmBlr:
I merged both the messages since you suggested that, when we last discussed about #17.

Disconnect is already wired in to pass on to the delist of that client in the leader elector. Thus crashing of the client will lead to delist at the server side ending up in sending the continuing nomination to other in the queue.

For the second point, as of now if there was only one client which ran for election and crashed, since the queue is empty, nothing will be send to any of the clients. The new client joining will actually get a elected Nomination rather that a continuing one. I will code this through.

AbfrmBlr:
Rebased

AbfrmBlr:
rebased

ljacomet:
Will not be merged as is, was fixed differently

PR #26, opened 2015-11-30
Made changes to management service producer/consumer interfaces 

rkavanap:
These changes are to allow for producing and consuming any management message as opaque simple byte arrays. This will help the management service to be a real broker without having to understand different management messages pushed from the entities to the management system. This will also help entities to avoid additional serialization/de-serialization and pass messages from the entities clients as is to the consuming management system.


rkavanap:
This is not required as of now as a few more changes may be required before finalizing. Will raise another pull request  in a couple of days after all the changes are refined,

PR #33, opened 2015-12-22
Fixed an occasional test hang issue in current management service build.

rkavanap:
The problem was in the test class that simulates Producer and Consumer of the ring buffer. ProducerOrConsumerSimulator.java, where the consumer was exiting before consuming all messages in rare scenarios.

Also, piggybacked a minor performance optimization that avoided a new every time an item is inserted. This changed some noticeable reduction some small CPU spikes (during GC) during long perf runs.

If this PR is merged, issue-32 and issue-21 can be closed.


mathieucarbou:
test this please

rkavanap:
test this please

PR #40, opened 2016-02-06
Added ctor to make codec optional

mathieucarbou:
To match the client-side:

    ClientProxyFactory.createEntityProxy(MyInterfaceEntity.class, MyInterface.class, endpoint, new SerializationCodec());


mathieucarbou:
@chrisdennis : Do we delete or merge this one, which is already in master ?

chrisdennis:
I would close it with a comment saying that the change has already been pushed.

mathieucarbou:
change has already been pushed.

PR #45, opened 2016-02-23
Simplification of voltron-proxy api

mathieucarbou:
- Simplification of voltron-proxy api
- Added some @Override
- Ensuring the project is built with 1.6 target (except if submodules override)

### DIFF
--- a/proxy/src/main/java/org/terracotta/voltron/proxy/server/ProxyServerEntityService.java
+++ b/proxy/src/main/java/org/terracotta/voltron/proxy/server/ProxyServerEntityService.java
@@ -1,0 +1,49 @@
+package org.terracotta.voltron.proxy.server;
+
+import org.terracotta.entity.ConcurrencyStrategy;
+import org.terracotta.entity.MessageCodec;
+import org.terracotta.entity.NoConcurrencyStrategy;
+import org.terracotta.entity.PassiveServerEntity;
+import org.terracotta.entity.ServerEntityService;
+import org.terracotta.entity.ServiceRegistry;
+import org.terracotta.voltron.proxy.Codec;
+import org.terracotta.voltron.proxy.ProxyMessageCodec;
+import org.terracotta.voltron.proxy.SerializationCodec;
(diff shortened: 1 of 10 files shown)


proxy/src/main/java/org/terracotta/voltron/proxy/server/ProxyServerEntityService.java
```diff
@@ -0,0 +1,49 @@
...
+    this(proxyType, new SerializationCodec());
+  }
+
+  public ProxyServerEntityService(Class<?> proxyType, Codec codec, Class<?> ... eventTypes) {
+    this.proxyType = proxyType;
+    this.codec = codec;
+    this.eventTypes = eventTypes;
+  }
+
+  @Override
```
myronkscott: Looks like this is already merged but it might make sense to have this method be abstract. What concurrency is used is implementation specific and outside the scope of what proxy should do. (Comms)

mathieucarbou: @chrisdennis : do you want me do remove the method ? I've put it on the top because I wanted to have a much more a skeleton (i.e. hence the method throwing an exception for the passive entity).

Let me know if we fix it :-) I'll send another PR then.

chrisdennis: NoConcurrency is always a functional default... it could be overridden if necessary.  I say leave it as is.

myronkscott: That would be fine.  Still seems to me more appropriate to force a selection.  Also, NoConcurrencyStrategy returns MANAGEMENT_KEY as the concurrency.  This may not actually be a reasonable default as it is special.  It forces a flush of all the queues and blocks execution on any other queues until completion.  It might be better to return 1.  The actual result from the entity's viewpoint is the same, just less machinery working.

PR #46, opened 2016-03-04
Move UuidLeaderOffer to common as it must be deserialized on the clie…

chrisdennis:
…nt side


coordinator-entity/common/src/main/java/org/terracotta/consensus/entity/UuidLeaderOffer.java
```diff
@@ -0,0 +1,33 @@
+/*
+ * To change this license header, choose License Headers in Project Properties.
+ * To change this template file, choose Tools | Templates
+ * and open the template in the editor.
+ */
```
anthonydahanne: fix header

PR #47, opened 2016-03-07
several bugfixes for @Async usage

mathieucarbou:
- fix: strip header of bytes read for call with timeout
- async mode now requires @Async (and a Future to be returned)  …
- fixed generic type support for returned Future
- Introduced a MethodDescriptor class to better handle the specifics of async and method calls
- Client and Server entities should use the same interface

### DIFF
--- a/proxy/src/main/java/org/terracotta/voltron/proxy/CommonProxyFactory.java
+++ b/proxy/src/main/java/org/terracotta/voltron/proxy/CommonProxyFactory.java
@@ -32,8 +32,8 @@ import static java.util.Collections.unmodifiableMap;
  * @author Alex Snaps
  */
 public class CommonProxyFactory {
-  private static final Comparator<Method> METHOD_COMPARATOR = new Comparator<Method>() {
-    public int compare(final Method m1, final Method m2) {
+  private static final Comparator<MethodDescriptor> METHOD_COMPARATOR = new Comparator<MethodDescriptor>() {
+    public int compare(final MethodDescriptor m1, final MethodDescriptor m2) {
       return m1.toGenericString().compareTo(m2.toGenericString());
     }
   };
@@ -54,12 +54,12 @@ public class CommonProxyFactory {
     return unmodifiableMap(inversion);
   }
 
-  public static Map<Byte, Method> createMethodMappings(final Class<?> proxyType) {
-    SortedSet<Method> methods = getSortedMethods(proxyType);
+  public static Map<Byte, MethodDescriptor> createMethodMappings(final Class<?> proxyType) {
+    SortedSet<MethodDescriptor> methods = getSortedMethods(proxyType);
 
-    final HashMap<Byte, Method> map = new HashMap<Byte, Method>();
+    final HashMap<Byte, MethodDescriptor> map = new HashMap<Byte, MethodDescriptor>();
     byte index = 0;
-    for (final Method method : methods) {
+    for (final MethodDescriptor method : methods) {
       map.put(index++, method);
     }
     return map;
@@ -68,10 +68,10 @@ public class CommonProxyFactory {
   public static Map<Class<?>, Byte> createResponseTypeMappings(Class<?> proxyType, Class<?> ... events) {
     final HashMap<Class<?>, Byte> map = new HashMap<Class<?>, Byte>();
     byte index = 0;
-    for (Method m : getSortedMethods(proxyType)) {
-      Class<?> returnType = m.getReturnType();
-      if (!map.containsKey(returnType)) {
-        map.put(returnType, index++);
+    for (MethodDescriptor m : getSortedMethods(proxyType)) {
+      Class<?> responseType = m.getMessageType();
+      if (!map.containsKey(responseType)) {
+        map.put(responseType, index++);
       }
     }
     for (Class<?> eventType : getSortedTypes(events)) {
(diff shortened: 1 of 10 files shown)


proxy/src/main/java/org/terracotta/voltron/proxy/client/VoltronProxyInvocationHandler.java
```diff
@@ -125,13 +128,34 @@ public Object invoke(final Object proxy, final Method method, final Object[] arg
...
 
+  static Class<?> determineRawType(Type type) {
+    if (type instanceof Class<?>) {
+      return (Class<?>) type;
+    }
+    if (type instanceof ParameterizedType) {
+      return determineRawType(((ParameterizedType) type).getRawType());
+    }
+    if (type instanceof WildcardType) {
+      return Object.class;
```
chrisdennis: Shouldn't this be the first upper bound?

proxy/src/main/java/org/terracotta/voltron/proxy/client/VoltronProxyInvocationHandler.java
```diff
@@ -125,13 +128,34 @@ public Object invoke(final Object proxy, final Method method, final Object[] arg
...
+    }
+    if (type instanceof GenericArrayType) {
+      return determineRawType(((GenericArrayType) type).getGenericComponentType());
```
chrisdennis: This seems wrong... it's returning the raw component type and not the array type corresponding to that.

proxy/src/test/java/org/terracotta/voltron/proxy/client/VoltronProxyInvocationHandlerTest.java
```diff
@@ -75,10 +80,36 @@ public void testNullsClientIdAnnotatedParams() throws Throwable {
...
+  @Test
+  public void test_raw_type_determined_correctly() throws Throwable {
+    Type test1 = ((ParameterizedType) RawType.class.getMethod("test1").getGenericReturnType()).getActualTypeArguments()[0];
+    Type test2 = ((ParameterizedType) RawType.class.getMethod("test2").getGenericReturnType()).getActualTypeArguments()[0];
+    Type test3 = ((ParameterizedType) RawType.class.getMethod("test3").getGenericReturnType()).getActualTypeArguments()[0];
+    Type test4 = ((ParameterizedType) RawType.class.getMethod("test4").getGenericReturnType()).getActualTypeArguments()[0];
+    Type test5 = ((ParameterizedType) RawType.class.getMethod("test5").getGenericReturnType()).getActualTypeArguments()[0];
+    Type test6 = ((ParameterizedType) RawType.class.getMethod("test6").getGenericReturnType()).getActualTypeArguments()[0];
+
+    assertEquals(VoltronProxyInvocationHandler.determineRawType(test1), String.class);
```
chrisdennis: assertThat(...) is going to give better info here on failure.

mathieucarbou: I did not find a way to use assertThat because determineRawType() returns Class<?> and hamcrest matchers do not support matching Class<?> with for example String.class.

I.e.

`assertThat(VoltronProxyInvocationHandler.determineRawType(test6), CoreMatchers.equalTo(Collection.class));`

is incompatible :-(

proxy/src/main/java/org/terracotta/voltron/proxy/client/VoltronProxyInvocationHandler.java
mathieucarbou: right I add a test!

mathieucarbou:
@chrisdennis : PR updated :-)

proxy/src/test/java/org/terracotta/voltron/proxy/CommonProxyFactoryTest.java
```diff
@@ -0,0 +1,61 @@
...
+import java.util.List;
+import java.util.concurrent.Future;
+
+import static org.hamcrest.CoreMatchers.equalTo;
+import static org.junit.Assert.assertThat;
+
+/**
+ * @author Mathieu Carbou
+ */
+@RunWith(JUnit4.class)
```
chrisdennis: Why the @RunWith?

proxy/src/main/java/org/terracotta/voltron/proxy/CommonProxyFactory.java
```diff
@@ -32,12 +33,49 @@
...
+          dimensions++;
+          cl = cl.getComponentType();
+        }
+        StringBuilder sb = new StringBuilder();
+        sb.append(cl.getName());
+        for (int i = 0; i < dimensions; i++) {
+          sb.append("[]");
+        }
+        return sb.toString();
+      } catch (Throwable e) { /*FALLTHRU*/ }
```
chrisdennis: What Throwable is this catching, and why?

mathieucarbou: It comes from `Field.getTypeName()` and the method using that is a shortened version of `Method.toGenericString()`.
I frankly don't know why there should be a try / catch there either. I guess there are some reasons but I cannot tell. If you do not find one either I can remove it :-)

proxy/src/test/java/org/terracotta/voltron/proxy/CommonProxyFactoryTest.java
mathieucarbou: This annotation tells Junit to use the `BlockJUnit4ClassRunner ` behind which replaces the old default `JUnit4ClassRunner `, which is deprecated in favor of `BlockJUnit4ClassRunner`.
=> https://github.com/junit-team/junit/wiki/Test-runners

chrisdennis: The new default is the `BlockJunit4ClassRunner` though isn't it.  And even if it isn't *why* are we changing it.

mathieucarbou:
@chrisdennis : PR updated!

Issue #50, opened 2016-03-10
OffHeap Storage Reservation Service


chrisdennis:
Implement a very simple 'honor system' offheap reservation service.  Conceptually this is just a map of counters, where the key is the 'pool name' and the counter is the amount of available resource.  Reserve and release are then simple increment/decrements of the resource.

Issue #49, opened 2016-03-10
XML Configuration API for offheap storage reservation service


chrisdennis:
In line with the ideas I've been throwing around around configuration of resources for the Terracotta Server, I want to replace direct configuration of the "Core Storage Service" with a more abstract (and more TSA admin user focussed) off heap reservation service.  This would be a service that is configured to manage a set of "pools" of resource with fixed sizes that entities can access indirectly through a "hidden" service that actually allocates memory, and then implements the storage needs of the entity.

PR #51, opened 2016-03-18
Issue #49 and Issue #50 : Offheap Resource

chrisdennis:
I'm most interested in people's review of the XML configuration I'm proposing here.  The implementation is a little rougher around the edges... and will likely get cleaned up as I work with/on it.

@ljacomet @cljohnso @myronkscott @jd0-sag you guys seem like a good shortlist for reviewers.

### DIFF
--- a/offheap-resource/src/main/java/org/terracotta/offheapresource/OffHeapResource.java
+++ b/offheap-resource/src/main/java/org/terracotta/offheapresource/OffHeapResource.java
@@ -1,0 +1,93 @@
+/*
+ * The contents of this file are subject to the Terracotta Public License Version
+ * 2.0 (the "License"); You may not use this file except in compliance with the
+ * License. You may obtain a copy of the License at
+ *
+ * http://terracotta.org/legal/terracotta-public-license.
+ *
+ * Software distributed under the License is distributed on an "AS IS" basis,
+ * WITHOUT WARRANTY OF ANY KIND, either express or implied. See the License for
+ * the specific language governing rights and limitations under the License.
+ *
+ * The Covered Software is OffHeap Resource.
+ *
+ * The Initial Developer of the Covered Software is
+ * Terracotta, Inc., a Software AG company
+ */
+
+package org.terracotta.offheapresource;
+
+import java.util.concurrent.atomic.AtomicLong;
+
+/**
+ * Represents an offheap resource, providing a reservation system that can be
+ * used to control the combined memory usage of participating consumers.
+ * <p>
+ * Reservation and release calls perform no allocations, and therefore rely on
+ * the cooperation of callers to achieve control over the 'real' resource usage.
+ */
+public class OffHeapResource {
+
+  private final AtomicLong remaining;
+
+  /**
+   * Creates a resource of the given initial size.
+   *
+   * @param size size of the resource
+   * @throws IllegalArgumentException if the size is negative
+   */
+  OffHeapResource(long size) throws IllegalArgumentException {
+    if (size < 0) {
+      throw new IllegalArgumentException("Resource size cannot be negative");
+    } else {
+      this.remaining = new AtomicLong(size);
+    }
+  }
+
+  /**
+   * Reserves the given amount of this resource.
+   * <p>
+   * This method <em>performs no allocation</em>.  It is simply a reservation
+   * that the consumer agrees to bind by the result of.  A {@code false} return
+   * should mean the caller refrains from performing any associated allocation.
+   *
+   * @param size reservation size
+   * @return {@code true} if the reservation succeeded
+   * @throws IllegalArgumentException if the reservation size is negative
+   */
+  public boolean reserve(long size) throws IllegalArgumentException {
+    if (size < 0) {
+      throw new IllegalArgumentException("Reservation size cannot be negative");
+    } else {
+      for (long current = remaining.get(); current >= size; current = remaining.get()) {
+        if (remaining.compareAndSet(current, current - size)) {
+          return true;
+        }
+      }
+      return false;
+    }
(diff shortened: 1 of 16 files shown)


jd0-sag:
There is the usual concern I have with units:  saying "MB" when we actually interpret it as "MiB", etc.

Might want some doc at the top of OffHeapResource to mention that the service merely does accounting and the caller is expected to actually do the allocation/free.  The API makes that pretty clear, though, so it might not matter.

Otherwise, this sounds like exactly what you described.  The lack of a "default" pool is good (the named resource really _should_ be explicit).  I also like that this doesn't try to position itself as a generic service but makes it clear that this is only for off-heap.

I am happy with this.

offheap-resource/src/main/java/org/terracotta/offheapresource/OffHeapResource.java
```diff
@@ -0,0 +1,58 @@
...
+
+  OffHeapResource(long size) {
+    if (size < 0) {
+      throw new IllegalArgumentException("Resource size cannot be negative");
+    } else {
+      this.remaining = new AtomicLong(size);
+    }
+  }
+
+  public boolean allocate(long size) {
```
lorban: Can you just call this method `alloc`? I usuallu don't like shortened names but alloc/free is a common pattern so I personally prefer to use well-known names.

offheap-resource/src/main/java/org/terracotta/offheapresource/OffHeapResource.java
```diff
@@ -0,0 +1,58 @@
...
+
+  public void free(long size) {
+    if (size < 0) {
+      throw new IllegalArgumentException("Freed size cannot be negative");
+    } else {
+      remaining.addAndGet(size);
+    }
+  }
+
+  public long available() {
```
lorban: Shouldn't that one be called `remaining`?

offheap-resource/src/main/java/org/terracotta/offheapresource/OffHeapResource.java
```diff
@@ -0,0 +1,58 @@
...
+
+package org.terracotta.offheapresource;
+
+import java.util.concurrent.atomic.AtomicLong;
+
+public class OffHeapResource {
+
+  private final AtomicLong remaining;
+
+  OffHeapResource(long size) {
```
lorban: I would keep the `size` argument in a final variable with a getter for the sake of M&M.

ramsai1729:
any plans for validating against available/physical size?

offheap-resource/src/main/java/org/terracotta/offheapresource/OffHeapResource.java
chrisdennis: Except it's not actually "alloc" since it doesn't really allocate.  I need to add some javadoc here anyway... so I can make the equivalence more explicit.

offheap-resource/src/main/java/org/terracotta/offheapresource/OffHeapResource.java
chrisdennis: The distinction being?

offheap-resource/src/main/java/org/terracotta/offheapresource/OffHeapResource.java
chrisdennis: Would be interesting to get some insight from @mathieucarbou on how such a service should expose that metric.  I'd rather it not be exposed through the normal API if it's only for use by M&M... since I imagine M&M would be accessing the service differently?

chrisdennis:
I could easily add this... will look in to doing this and warning about it.  I'd rather not hard fail on this though since this service doesn't actually allocate anything at all... so it seems a bit too heavy handed to fail if we're short on physical memory.

offheap-resource/src/main/java/org/terracotta/offheapresource/OffHeapResource.java
mathieucarbou: @chrisdennis : if we re-use the management registry and provider approach like in ehcache, we'll have to call a sort of `registry.register(object)` and a provider will be used to expose the management metadata for its own objects. 

So basically someone should make these calls and the settings should be accessible (through getters)... Or at least the configuration object used to create the resource ? Because we can choose what is registered.

Anyway, perhaps we can try to make a little PoC next week, at the same time as for the ehcache server entity ?

offheap-resource/src/main/java/org/terracotta/offheapresource/OffHeapResource.java
lorban: If you want to differentiate this service from the traditional alloc/free -which I think makes sense- shouldn't you use other terms in the API?

offheap-resource/src/main/java/org/terracotta/offheapresource/OffHeapResource.java
lorban: I like `remaining` better. :)

offheap-resource/src/main/java/org/terracotta/offheapresource/OffHeapResource.java
chrisdennis: Makes sense.. switching to reserve and release.

ljacomet:
I would be in favour of failing fast if we know from the start the memory is not available. At least certainly if the total physical memory is less than configured. And if the free memory is less, I would still fail as well ...

chrisdennis:
I have create #53 for tracking the overcommit behavior.  Until we have something closer to a complete system I don't want to make any decisions here, since I believe this has to be approached in a holistic manner.

PR #52, opened 2016-03-22
Update management to new ServiceProvider API

chrisdennis:


management-service/src/main/java/org/terracotta/management/service/ManagementServiceProvider.java
```diff
@@ -27,6 +27,7 @@
 import java.io.IOException;
 import java.util.Collection;
 import java.util.Collections;
+import org.terracotta.entity.ServiceProviderCleanupException;
```
anthonydahanne: there are 2 unused imports in your commit, could you please fix them ?
import org.terracotta.entity.ServiceProviderCleanupException;

import java.io.IOException;

PR #54, opened 2016-03-24
Added accessors to be able to access create server entity instance…

mathieucarbou:
…in example to close it when the server entity is destroyed


mathieucarbou:
@anthonydahanne : FYI also.

mathieucarbou:
@chrisdennis : please do not merge directly, we need to discuss that because basically I would need a way to get the implementation reference, but not the interface.

I.e. I would need to do in my server entity service:

```
public class AgentServerEntity extends ProxiedServerEntity<Agent> {

  public AgentServerEntity(AgentConfig config, IMonitoringConsumer consumer) {
    super(Agent.class, new AgentImpl(config, consumer));
  }

  @Override
  public void destroy() {
    getTarget().close();
  }

}
```

The issue is that getTarget() based on `T` would return the proxified, interface, which is also the interface used on client-side...

chrisdennis:
This could be achieved without the cast by throwing some more generics at the problem... but I'd like to discuss if this is really the "right" way to solve this in this case.

mathieucarbou:
we can close ;-)

PR #55, opened 2016-03-30
Adapt to new core message codec API

chrisdennis:
So this passes all tests, but I haven't tested much more than that.  If @anthonydahanne or @mathieucarbou feel like testing the proxy code against their usecase that would be great.


mathieucarbou:
@chrisdennis : there is a test failure but I guess its not related to your changes ?
Can I still merge or you'd like to have a look / fix it ?

anthonydahanne:
all tests pass on my laptop

PR #60, opened 2016-04-19
Fix over allocation issue in ClientCommunicatorRequestCodec

ramsai1729:


client-communicator-support/src/main/java/org/terracotta/clientcommunicator/support/ClientCommunicatorRequestCodec.java
```diff
@@ -7,7 +7,7 @@
  */
 class ClientCommunicatorRequestCodec {
     public static byte[] serialize(ClientCommunicatorRequest request) {
-        int size = request.getMsgBytes().length + 12;
+        int size = request.getMsgBytes().length + 8;
```
jd0-sag: Add a comment to explain why this "magic number" is correct.

jd0-sag: Probably something like (2 * Integer.BYTES)?

PR #61, opened 2016-04-29
voltron-proxy: Support serialization of entity config through the installed codec

mathieucarbou:
Support serialization of entity config through the installed codec and supporting default Void config type.
This provides a default serialization behavior to help implementing entities easier.
Behavior can be overridden, as before.


mathieucarbou:
CC @chrisdennis - FYI

Issue #62, opened 2016-04-29
ServiceRegistry.getService(OffHeapResourceIdentifier) returns null


cljohnso:
An active Entity (EhcacheActiveEntity in the failing case) attempting to use the ServiceRegistry to locate an OffHeapResource identified by an OffHeapResourceIdentifier fails to find the resource even though defined.  

The underlying cause of this failure is the difference in ClassLoaders used for services versus entities -- 
service-defined objects are loaded using a ComponentURLClassLoader and objects instantiated by an entity are loaded using an AppClassLoader.  Configured in a <service> element of a tc-config.xml, OffHeapResource instances are created using the ComponentURLClassLoader (as are the OffHeapResourceIdentifier instances used to track them).  When an Entity uses ServiceRegistry.getService to attempt to look up an OffHeapResource, the OffHeapResourceIdentifier is 
constructed using the AppClassLoader (along with the OffHeapResource reference that class contains).  The OffHeapResource class reference, used as a map key within ServiceRegistry, fails to match when the key and lookup argument are loaded by different ClassLoaders.

To force the use of a common ClassLoader, the OffHeapResource and OffHeapResourceIdentifier classes need to be marked with the @CommonComponent annotation (in org.terracotta:packaging-support).

cljohnso:
Closed by PR #63

PR #63, opened 2016-04-29
Issue #62 Make OffHeapResource a CommonComponent

cljohnso:
When performing a org.terracotta.entity.ServiceRegistry.getService call
against OffHeapResource, the configured resources can not be found
due to a difference in the ClassLoaders being used.  Adding the
@CommonComponent annotation to OffHeapResource causes the common
ClassLoader to be used for OffHeapResource so the service-created
instances can be found.  Since OffHeapResourceIdentifier class
references are also used during OffHeapResource lookup (instanceof),
this class is also marked with @CommonComponent.

This commit separates the OffHeapResource interface from its
implementation to avoid an issue during instantiation after application
of the @CommonComponent annotation.

### DIFF
--- a/offheap-resource/pom.xml
+++ b/offheap-resource/pom.xml
@@ -24,6 +24,11 @@
       <version>10.0-SNAPSHOT</version>
     </dependency>
     <dependency>
+      <groupId>org.terracotta</groupId>
+      <artifactId>packaging-support</artifactId>
+      <version>${terracotta-apis.version}</version>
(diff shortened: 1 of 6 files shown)


cljohnso:
@chrisdennis ... you need to at least be aware of this one.

chrisdennis:
This is fine, I have the same effective changes sitting in a local branch from when I was working on this stuff.  Apologies for not pushing it myself.

PR #64, opened 2016-05-03
Update to account for EndpointDelegate API change

jd0-sag:
-EndpointDelegate.handleMessage() now takes EntityResponse instead of byte[]

**Note:**  This depends on Terracotta-OSS/terracotta-apis#78

### DIFF
--- a/proxy/src/main/java/org/terracotta/voltron/proxy/client/ProxyEndpointDelegate.java
+++ b/proxy/src/main/java/org/terracotta/voltron/proxy/client/ProxyEndpointDelegate.java
@@ -18,8 +18,10 @@
 package org.terracotta.voltron.proxy.client;
 
 import org.terracotta.entity.EndpointDelegate;
+import org.terracotta.entity.EntityResponse;
 import org.terracotta.voltron.proxy.Codec;
 import org.terracotta.voltron.proxy.client.messages.MessageListener;
+import org.terracotta.voltron.proxy.server.messages.ProxyEntityResponse;
 
 import java.util.Arrays;
 import java.util.Map;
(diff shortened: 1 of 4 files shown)


proxy/src/main/java/org/terracotta/voltron/proxy/client/ProxyEndpointDelegate.java
```diff
@@ -44,11 +46,11 @@ public ProxyEndpointDelegate(final Codec codec,
   }
 
   @Override
-  public void handleMessage(final byte[] bytes) {
-    final Object message = codec.decode(Arrays.copyOfRange(bytes, 1, bytes.length), eventMappings.get(bytes[0]));
-    final Class<?> aClass = message.getClass();
+  public void handleMessage(EntityResponse messageFromServer) {
+    ProxyEntityResponse response = (ProxyEntityResponse)messageFromServer;
+    final Class<?> aClass = response.getResponseType();
```
mathieucarbou: @jd0-sag @chrisdennis : I think we could also remove all the eventMappings maps that was used to map the first byte in the message to the event type, right ?

mathieucarbou: Update: I see the cleanup has not been done for the previous type mapping also.

```
public VoltronProxyInvocationHandler(final Map<MethodDescriptor, Byte> mappings,
                                       final EntityClientEndpoint entityClientEndpoint,
                                       final Codec codec, Map<Byte, Class<?>> eventMappings) {
```

Both `mappings` and `eventMappings` would  be / are useless and we can event do a further cleanup in:

```
        new VoltronProxyInvocationHandler(
                invert(createMethodMappings(type)),
                entityClientEndpoint, codec,
                invert(createResponseTypeMappings(type, messageTypes))
```

Do you want me do file an issue to remember to do this cleanup out of this PR ?

jd0-sag: If that cleanup is required in order for this PR to be correct, then talk to me about it so I can fix it, here.

If it is just a simplification post-pass (which is what it sounds like), then create a new issue which references this PR and someone can pick it up, in the future.

mathieucarbou: Yes its a post-cleanup. Will do it after. Thanks :-)

PR #72, opened 2016-06-14
add a connection healthchecker entity

myronkscott:
Add an entity that monitors the health of the underlying connection and closes it if connectivity is in question

### DIFF
--- a/healthchecker-entity/api/src/main/java/org/terracotta/healthchecker/HealthCheckerFactory.java
+++ b/healthchecker-entity/api/src/main/java/org/terracotta/healthchecker/HealthCheckerFactory.java
@@ -1,0 +1,133 @@
+/*
+ * All content copyright (c) 2014 Terracotta, Inc., except as may otherwise
+ * be noted in a separate copyright notice. All rights reserved.
+ */
+package org.terracotta.healthchecker;
+
+import java.io.IOException;
+import java.util.Collections;
+import java.util.LinkedHashSet;
+import java.util.Properties;
+import java.util.Set;
+import java.util.Timer;
+import java.util.TimerTask;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+import org.terracotta.connection.Connection;
+import org.terracotta.connection.entity.EntityRef;
+import org.terracotta.exception.EntityNotFoundException;
+import org.terracotta.exception.EntityNotProvidedException;
+import org.terracotta.exception.EntityVersionMismatchException;
+
+/**
+ *
+ * @author mscott
+ */
+public class HealthCheckerFactory {
+  
+  private static final String NAME = "staticHealthChecker";
+  
+  public static TimeoutManager startHealthChecker(Connection connection, int probeFrequencyPerMinute, long probeTimeoutInMillis) {
+    try {
+      EntityRef<HealthCheck, Properties> check = connection.getEntityRef(HealthCheck.class, HealthCheck.VERSION, NAME);
+      HealthCheck hc = check.fetchEntity();
+      if (probeFrequencyPerMinute < 1 || probeFrequencyPerMinute > 120) {
+        throw new IllegalArgumentException("probe frequency must be greater than zero and less than 120");
+      }
+      return new HealthCheckTimeoutManager(connection, hc).start(probeTimeoutInMillis, probeFrequencyPerMinute);
+    } catch (EntityNotProvidedException notvalid) {
+      throw new IllegalStateException("healthchecker entity is not installed with the name " + NAME, notvalid);
+    } catch (EntityNotFoundException notfound) {
+      throw new IllegalStateException("healthchecker entity is not installed with the name " + NAME, notfound);
+    } catch (EntityVersionMismatchException version) {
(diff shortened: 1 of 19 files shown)


healthchecker-entity/api/src/main/java/org/terracotta/healthchecker/HealthCheckerFactory.java
```diff
@@ -0,0 +1,146 @@
...
+
+    @Override
+    public <T extends Entity, C> EntityRef<T, C> getEntityRef(Class<T> cls, long version, String name) throws EntityNotProvidedException {
+      return root.getEntityRef(cls, version, name);
+    }
+
+    @Override
+    public void close() throws IOException {
+      driver.cancel();
+      checker.close();
```
jd0-sag: Do we want to explicitly close the entity, here, or are we worried about that "hang on close in the case of disconnected server" such that we should just strong-arm the connection into a closed state and handle the unexpected disconnect, from within the entity, in concert with this?

healthchecker-entity/client-impl/src/main/java/org/terracotta/healthchecker/HealthCheckerClient.java
```diff
@@ -0,0 +1,96 @@
...
+  private final EntityClientEndpoint<HealthCheckReq, HealthCheckRsp> endpoint;
+
+  public HealthCheckerClient(EntityClientEndpoint<HealthCheckReq, HealthCheckRsp> endpoint) {
+    this.endpoint = endpoint;
+  }
+
+  @Override
+  public Future<String> ping(String message) {
+    try {
+      return wrapFuture(endpoint.beginInvoke().message(new HealthCheckReq(message)).replicate(false).invoke());
```
jd0-sag: Do we want to add a comment here to state that we explicitly CANNOT register for any acks here, just to make sure we don't create a problem, in the future?

healthchecker-entity/client-impl/src/main/java/org/terracotta/healthchecker/HealthCheckerClient.java
```diff
@@ -0,0 +1,96 @@
...
+
+      @Override
+      public boolean isDone() {
+        return invoke.isDone();
+      }
+
+      @Override
+      public String get() throws InterruptedException, ExecutionException {
+        try {
+          return invoke.get().toString();
```
jd0-sag: Should we implement this version of the get(), in this path, or throw an unsupported operation since we don't expect it to be used that way?

healthchecker-entity/client-impl/src/main/java/org/terracotta/healthchecker/HealthCheckerClient.java
```diff
@@ -0,0 +1,96 @@
...
+
+  @Override
+  public byte[] createExtendedReconnectData() {
+// do nothing
+    return new byte[0];
+  }
+
+  @Override
+  public void didDisconnectUnexpectedly() {
+//  do nothing
```
jd0-sag: Is it worth trying to break out of any timeout-limited get, in this case, or do we assume that the extra path isn't worth it?

healthchecker-entity/server-impl/src/main/java/org/terracotta/healthchecker/HealthCheckServerEntityService.java
```diff
@@ -0,0 +1,74 @@
...
+    @Override
+    public Set getKeysForSynchronization() {
+      return Collections.emptySet();
+    }
+
+  };
+
+  @Override
+  public long getVersion() {
+    return 1L;
```
jd0-sag: Can we reference the version constant, from here, or is it not visible?

myronkscott: Not visible.  I don't like puling in API just for these statics

healthchecker-entity/api/src/main/java/org/terracotta/healthchecker/HealthCheckerFactory.java
myronkscott: I think that makes better sense.

healthchecker-entity/client-impl/src/main/java/org/terracotta/healthchecker/HealthCheckerClient.java
myronkscott: The entity is wholly internal at the moment.  Let me just comment and leave it.

healthchecker-entity/client-impl/src/main/java/org/terracotta/healthchecker/HealthCheckerClient.java
myronkscott: i don't think it is worth it.  the normal path should clean itself up eventually.

healthchecker-entity/client-impl/src/main/resources/META-INF/services/org.terracotta.entity.EntityClientService
```diff
@@ -0,0 +1 @@
+org.terracotta.healthchecker.HealthCheckClientService
```
jd0-sag: Does this need the same license as org.terracotta.entity.ServerEntityService?

PR #78, opened 2016-06-17
Issue 77 - wrap user types

ljacomet:


concurrent-map-entity/client/src/main/java/org/terracotta/entity/map/TerracottaClusteredMap.java
```diff
@@ -126,36 +165,62 @@ public void clear() {
...
+    for (Object o : response.getValues()) {
+      result.add(valueValueCodec.decode(o));
+    }
+    return result;
   }
 
   @Override
   public Set<Map.Entry<K, V>> entrySet() {
-    return (Set<Map.Entry<K, V>>)(Object)((EntrySetResponse)invokeWithReturn(new EntrySetOperation())).getEntrySet();
+    HashSet<Entry<K, V>> result = new HashSet<Entry<K, V>>();
```
albinsuresh: Is there a requirement to preserve the order of the entries in the map? If not then this should be fine.

albinsuresh: Scratch that. The return type is a set. So shouldn't matter.

concurrent-map-entity/client/src/main/java/org/terracotta/entity/map/TerracottaClusteredMap.java
```diff
@@ -79,27 +99,42 @@ public boolean isEmpty() {
...
+      return null;
+    }
+    MapValueResponse response = (MapValueResponse) invokeWithReturn(new GetOperation(keyValueCodec.encode((K) key)));
+    return valueValueCodec.decode(response.getValue());
   }
 
   @Override
   public V put(K key, V value) {
-    return (V) ((MapValueResponse)invokeWithReturn(new PutOperation(key, value))).getValue();
+    MapValueResponse response = (MapValueResponse) invokeWithReturn(new PutOperation(keyValueCodec.encode(key), valueValueCodec.encode(value)));
```
AbfrmBlr: Can we add `isAssignableFrom` check in put, putifabsent and replace operations as well ?

ljacomet: Well, I am relying on the generic type for that. So you would want an early `ClassCastException`?

AbfrmBlr: Oh yes

PR #81, opened 2016-06-20
Moved org.terracotta.entity.map.common to org/terracotta/entity/map/common

jd0-sag:
-some tools complain if the source and class package location don't agree and we already use the directory shape for the other modules

Let me know if you think this isn't worth changing since it isn't "technically" wrong.


jd0-sag:
(removed the resurrected MapEntry file - should be good to merge, now, if you think this makes sense)

ljacomet:
Wow, sorry for that, will blame the IDE! No idea I had messed it up that way.

PR #91, opened 2016-06-28
Moving management in

anthonydahanne:
In order to ease the release process, we want to manage less git repositories.
We came to think that moving terracotta-management to terracotta-platform makes sense, since modules in terracotta-management only depend on terracotta-platform and terracotta-apis
History seems to be preserved, everything builds fine; can you please verify this move is harmless ?
Thank you


mathieucarbou:
Looks good to me.

@ljacomet : we had yesterday the idea to do the migration through a PR to only keep all the commit history from master and not impact the current history / files that exist. 

If it's fine, we can merge.

PR #96, opened 2016-06-28
Added support to lookup all available offheap resource identifiers

albinsuresh:


terracotta-org:
Can one of the admins verify this patch?

ljacomet:
This seems to be missing an addition to `org.terracotta.offheapresource.OffHeapResourcesProvider#getProvidedServiceTypes`

offheap-resource/src/main/java/org/terracotta/offheapresource/OffHeapResourceIdentifier.java
```diff
@@ -52,6 +52,11 @@ public boolean equals(Object obj) {
   }
 
   @Override
+  public String toString() {
```
ljacomet: If we are going to return this type instead of a simple `String`, I recommend it gets a `getName` method instead.

offheap-resource/src/main/java/org/terracotta/offheapresource/OffHeapResourcesProvider.java
```diff
@@ -75,14 +76,21 @@ public synchronized boolean initialize(ServiceProviderConfiguration unknownConfi
     if (unknownConfiguration instanceof OffHeapResourceIdentifier) {
       OffHeapResourceIdentifier identifier = (OffHeapResourceIdentifier) unknownConfiguration;
       return (T) identifier.getServiceType().cast(resources.get(identifier));
+    } else if (OffHeapResources.class.isAssignableFrom(unknownConfiguration.getServiceType())) {
+      return (T) new OffHeapResources() {
+        @Override
+        public Set<OffHeapResourceIdentifier> getAllIdentifiers() {
+          return resources.keySet();
```
ljacomet: Please, do not simply return the keySet - as it enables remove operations.
Return a read-only set of all the names instead.

offheap-resource/src/main/java/org/terracotta/offheapresource/OffHeapResourceIdentifier.java
albinsuresh: The sole reason why I added this here is so that I can use the returned `Set<OffHeapResourceIdentifier>` from OffHeapResourcesProvider.getAllIdentifiers` method as such in log messages. But if that's not allowed then I don't really see a point in adding a `getName` method here unless we plan on using it for something else.

offheap-resource/src/main/java/org/terracotta/offheapresource/OffHeapResourcesProvider.java
albinsuresh: Sure

ljacomet:
Looks good to me now - @chrisdennis wdyt?

offheap-resource/src/main/java/org/terracotta/offheapresource/OffHeapResources.java
```diff
@@ -0,0 +1,15 @@
...
+
+import java.util.Set;
+
+/**
+ * Represents a collection of {@link OffHeapResource} instances
+ */
+@CommonComponent
+public interface OffHeapResources {
+
+  Set<String> getAllIdentifiers();
```
chrisdennis: Why is it a set of strings and not a set of OffHeapResourceIdentifier?

chrisdennis: I also hate the indirection here... but I guess thats required given the way the service registry works.

ljacomet: Let's keep it like that and revisit once we clarify `ServiceProvider` and `ServiceRegistry` contracts

ljacomet:
test this please

PR #97, opened 2016-06-28
Further platform clean

anthonydahanne:


pom.xml
```diff
@@ -147,6 +147,29 @@
         <groupId>org.sonatype.plugins</groupId>
         <artifactId>nexus-staging-maven-plugin</artifactId>
       </plugin>
+      <plugin>
```
mathieucarbou: I think inherited is missing there because using such config means the plugin is inherited to all sub-modules.
And header.txt only exists at the root level.

So the trick is to use inherited=false + the aggregate option of the plugin.
Otherwise, have a dependency jar deployed, containing the plugin, or read the file from a remote location.

I.e.: try to run `mvn clean verify` on the project to confirm if it works. The verify phase is triggered before the release:prepare to be sure every licenses are there.

anthonydahanne: just tried that
did not seem to work
I'll show you

anthonydahanne:
@mathieucarbou I've committed the changes

PR #99, opened 2016-06-29
make management rely on platform parent versions

anthonydahanne:


anthonydahanne:
@mathieucarbou please check if that's ok with you : I've verified in an intellij workspace only importing management that everything worked fine

pom.xml
```diff
@@ -28,6 +28,7 @@
     <maven-forge-plugin.version>1.2.16</maven-forge-plugin.version>
     <slf4j.version>1.7.7</slf4j.version>
     <terracotta-apis.version>1.0-SNAPSHOT</terracotta-apis.version>
+    <platform.version>1.0-SNAPSHOT</platform.version>
```
ljacomet: Should be `${project.version}`

PR #104, opened 2016-06-29
Clean up pom files

ljacomet:
* Get rid of versions for managed dependencies
* Remove platform.version property
* Misc cleanups


anthonydahanne:
LGTM

PR #107, opened 2016-07-04
Fixed header style that was in javadoc style, to avoid the IDE reformat

mathieucarbou:
... to avoid the IDE to reformat the header each type by adding `<p>`

The problem is that reformatting a class or modifying causes the IDE to reformat the javadoc. The issue is that the license headers were in javadoc style, so they all get reformatted also.

Added to `maven-license-plugin`:

```
          <mapping>
            <java>SLASHSTAR_STYLE</java>
          </mapping>
```


anthonydahanne:
LGTM

PR #111, opened 2016-07-06
Several bug-fixes for management entity

mathieucarbou:
- Sequence object is not sharable, must be converted to byte[]
 - Using serializable array instead of lists to store data into
 voltron tree
 - separating tree branches for client-side notifications and stats
 from server entity notifs and stats


anthonydahanne:
LGTM

PR #113, opened 2016-07-13
Preparing support for management calls - links a

mathieucarbou:
management registry directly with the management agent service

### DIFF
--- a/management/management-registry/src/main/java/org/terracotta/management/registry/ManagementProviderAdapter.java
+++ b/management/management-registry/src/main/java/org/terracotta/management/registry/ManagementProviderAdapter.java
@@ -1,0 +1,97 @@
+/*
+ * Copyright Terracotta, Inc.
+ *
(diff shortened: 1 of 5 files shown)


management/management-registry/src/main/java/org/terracotta/management/registry/ManagementProviderAdapter.java
```diff
@@ -0,0 +1,97 @@
...
+  @Override
+  public void register(Object managedObject) {
+  }
+
+  @Override
+  public void close() {
+  }
+
+  @Override
+  public void unregister(Object managedObject) {
```
anthonydahanne: abstract ?

mathieucarbou: no: it's an adapter.

anthonydahanne:
LGTM

PR #117, opened 2016-07-14
compatibility with latest tc apis and clustering testing libs

anthonydahanne:


pom.xml
```diff
@@ -27,9 +27,9 @@
   <properties>
     <maven-forge-plugin.version>1.2.16</maven-forge-plugin.version>
     <slf4j.version>1.7.7</slf4j.version>
-    <terracotta-apis.version>1.0.2.beta</terracotta-apis.version>
+    <terracotta-apis.version>1.0.3.beta</terracotta-apis.version>
     <tcconfig.version>10.0.2.beta</tcconfig.version>
```
jd0-sag: If you are updating this, you should also update the config version (although I do wonder what you have that depends on it).

anthonydahanne: @jd0-sag yeah that's what I figured, I sent another PR just after

PR #119, opened 2016-07-15
Renamed resource files

jd0-sag:
-org.terracotta.entity.ServerEntityService renamed to org.terracotta.entity.EntityServerService since the backing type name changed so this meant that the ServiceLoader wasn't finding these


ljacomet:
Needs a release once merged! Breaks ehcache in non obvious ways and I missed that.

anthonydahanne:
LGTM

PR #120, opened 2016-07-20
Issue #121 state service

AbfrmBlr:
Service added which will hold state for entity and enable them to share state when being promoted to active from passive

### DIFF
--- a/entity-state-service/src/main/java/org/terracotta/state/EntityStateRepository.java
+++ b/entity-state-service/src/main/java/org/terracotta/state/EntityStateRepository.java
@@ -28,7 +28,7 @@ import java.util.Map;
 @CommonComponent
 public interface EntityStateRepository {
 
-  < K, V> Map<K, V> getOrCreateState(String name, Class<K> kClass, Class<V> vClass);
+  <K, V> Map<K, V> getOrCreateState(String name, Class<K> kClass, Class<V> vClass);
 
   void destroyState(String name);
 }
--- a/entity-state-service/src/main/java/org/terracotta/state/EntityStateRepository.java
+++ b/entity-state-service/src/main/java/org/terracotta/state/EntityStateRepository.java
@@ -1,0 +1,34 @@
+/*
+ * Copyright Terracotta, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.terracotta.state;
+
+import com.tc.classloader.CommonComponent;
+
+import java.util.Map;
+
+/**
+ * Repository to hold state of entities.
+ * This class is not thread safe, since it's always expected
+ * that state changes only during lifecycle of entity under maintenance.
+ */
+@CommonComponent
+public interface EntityStateRepository {
+
+  < K, V> Map<K, V> getOrCreateState(String name, Class<K> kClass, Class<V> vClass);
+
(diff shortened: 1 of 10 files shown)


entity-state-service/src/main/java/org/terracotta/state/EntityStateRepository.java
```diff
@@ -0,0 +1,34 @@
...
+
+/**
+ * Repository to hold state of entities.
+ * This class is not thread safe, since it's always expected
+ * that state changes only during lifecycle of entity under maintenance.
+ */
+@CommonComponent
+public interface EntityStateRepository {
+
+  < K, V> Map<K, V> getOrCreateState(String name, Class<K> kClass, Class<V> vClass);
```
albinsuresh: Formatting of the generic parameters: < K, V> : <K, V>

pom.xml
```diff
@@ -42,6 +42,7 @@
     <module>client-communicator-support</module>
     <module>healthchecker-entity</module>
     <module>management</module>
+	  <module>entity-state-service</module>
```
albinsuresh: Formatting again.

entity-state-service/src/main/java/org/terracotta/state/config/TransientEntityStateRepositoryConfig.java
```diff
@@ -0,0 +1,29 @@
...
+ */
+
+package org.terracotta.state.config;
+
+import org.terracotta.entity.ServiceConfiguration;
+import org.terracotta.state.TransientEntityStateRepository;
+
+/**
+ */
+public class TransientEntityStateRepositoryConfig implements ServiceConfiguration<TransientEntityStateRepository> {
```
albinsuresh: Do we really need this new config type? From the `EntityStateServiceProvider` you can invoke _getServiceType()_ on the incoming configuration rather than doing an *instanceOf* check for this exact same type.

entity-state-service/src/test/java/org/terracotta/state/EntityStateServiceProviderTest.java
```diff
@@ -0,0 +1,97 @@
...
+import static org.hamcrest.Matchers.startsWith;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertNotSame;
+import static org.junit.Assert.assertSame;
+import static org.junit.Assert.assertThat;
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
+import static org.mockito.Mockito.mock;
+
+/**
```
albinsuresh: Get rid of those comment lines pls.

entity-state-service/src/test/java/org/terracotta/state/TransientEntityStateRepositoryTest.java
```diff
@@ -0,0 +1,69 @@
...
+import org.junit.Test;
+
+import java.util.Map;
+
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertNotSame;
+import static org.junit.Assert.assertSame;
+
+
+/**
```
albinsuresh: This too.

entity-state-service/src/main/java/org/terracotta/state/EntityStateRepository.java
```diff
@@ -0,0 +1,34 @@
...
+
+/**
+ * Repository to hold state of entities.
+ * This class is not thread safe, since it's always expected
+ * that state changes only during lifecycle of entity under maintenance.
+ */
+@CommonComponent
+public interface EntityStateRepository {
+
+  <K, V> Map<K, V> getOrCreateState(String name, Class<K> kClass, Class<V> vClass);
```
albinsuresh: Why not make this return a `ConcurrentMap`? That way we wouldn't have to restrict the usage of this service to non-concurrent usages. Currently the javadoc for this service says that this can be only used in life-cycle of entities under maintenance. Why restrict it like that? Making it concurrent is not very costly, is it?

chrisdennis:
I don't understand why this is a generalized service in `terracotta-platform` and not a specialized service in the ehcache codebase directly.

AbfrmBlr:
Do not merge it yet

AbfrmBlr:
Closing this as it is decided to move this service in ehcache and make it ehcache specific.

Issue #108, opened 2016-07-04
Replace bounded queues used to store notifications and statistics


mathieucarbou:
The server entity for management (`ManagementAgentServerEntity`) creates 2 bounded queues to store in the monitoring tree all the notifications and stats coming from clients.

=> use instead a better system, ring-buffer ? see old saved code in the project.

__Depends on:__

- [ ] Terracotta-OSS/terracotta-apis#134
- [ ] #128

PR #125, opened 2016-08-10
refactoring of the monitoring service

mathieucarbou:
refactoring of the monitoring service
- moved unread notification settings to consumer configuration (this
was a consumer option, not a service-level option)
- renamed recordMutation option in builder
- refactored how mutations are collected, per consumer

This is a preparation work for #108  and new IMonitoringProducer method that will record notifications and statistics


anthonydahanne:
LGTM

PR #129, opened 2016-08-16
Issues #128, #108, #109, #130 and compat' with Terracotta-OSS/terracotta-apis#134

### LINKED ISSUE #108: Replace bounded queues used to store notifications and statistics

### LINKED ISSUE #109: Replace priority queue used to store voltron tree mutation
Use a better system to store unread mutations per consumer in `MonitoringService`

Get rid of  `MonitoringServiceConfiguration.maximumUnreadMutationsPerConsumer` ?

__Depends on:__

- [x] Terracotta-OSS/terracotta-apis#134
- [x] #128

### LINKED ISSUE #127: Put the ManagementRegistry as a Voltron service
Idea:

Having a `VoltronManagementRegistryService` implementation of a voltron service providing the `ManagementRegistry` interface.

Then, each server entity could:

1. get a ref to the `IMonitoringProducer` interface
2. create a specific `ManagementProvider` for this server entity constructed with the entity name and type or id and the `IMonitoringProducer`.
3. call `ManagementRegistry.addManagementProvider()` => this call would add a new provider to the voltron service that will support exposing metadata into voltron monitoring tree
4. then call as wanted `ManagementRegistry.register()`

The `VoltronManagementRegistryService` implementation and the `ManagementProviderSkeleton` would provide the common mechanism to expose the metadata by using the `IMonitoringProducer` of each provider.

First use case would be for Ehcache: ehcache/ehcache3#1008

Depends on issues:

- Terracotta-OSS/tc-passthrough-testing#14
- Terracotta-OSS/terracotta-apis#140

### LINKED ISSUE #128: Add a best-efforts "event push" method to IMonitoringProducer
Implementation of Terracotta-OSS/terracotta-apis#134 in monitoring service

### LINKED ISSUE #130: Moving TMS Entity in this project and use the ImonitoringConsumer.getBuffer() method
- because the tms entity do not have any dependency on tms. but only from tc-platform libraries
- because tms entity just "reads" topology and messages, all objects in OSS already.

Also, the TMS Entity needs to use the new available methods for notifs and stats

mathieucarbou:
Part of the work required for ehcache/ehcache3#1008

__Content of this PR:__

- TMS Entity moved into tc-platform
- Buffer (ring buffer) support in monitoring service
- Buffers are per consumer (so several consumers can _listen_ to the same category name)
- `IMonitoringProducer.push()` added
- `IMonitoringConsumer.readMutations()` removed in favor of buffer usage
- `IMonitoringConsumer.getOrCreateBuffer()` added
- Configuration cleanup
- TMS Entity and Management Entity now uses the buffers
- `Version` classes renamed to `TmsAgentVersion` and `ManagementAgentVersion` to avoid conflict
- Forward compatibility with Terracotta-OSS/terracotta-apis#134

__Release to do after: 5.0.4.beta__

__For GitHub:__

- Closes #128
- Closes #108 
- Closes #109
- Closes #130


mathieucarbou:
@anthonydahanne : once the build will pass and that this PR will be merged, could you please release tc-platform, to version __5.0.4-beta__, to comply with the other version patterns ?

mathieucarbou:
test this please

mathieucarbou:
should pass tomorrow, when released versions will be deployed.

management/monitoring-service/src/main/java/org/terracotta/management/service/monitoring/MonitoringService.java
```diff
@@ -60,44 +54,129 @@
...
+        if (null != node) {
+          children = node.getChildValues();
+        }
+        return Optional.ofNullable(children);
+      }
+
+      @SuppressWarnings("unchecked")
+      @Override
+      public <V> ReadOnlyBuffer<V> getOrCreateBestEffortBuffer(String category, int maxBufferSize, Class<V> type) {
+        if ("monitoring-tree-mutations".equals(category) && Mutation.class != type) {
```
anthonydahanne: constant ? 
```
private static String MONITORING_TREE_MUTATIONS = "monitoring-tree-mutations";
```

management/monitoring-service/src/test/java/org/terracotta/management/service/monitoring/MonitoringServiceTest.java
```diff
@@ -258,8 +261,41 @@ public void test_topology_mutation_replace_branch() {
     assertEquals(asList(asList(), asList()), parentVals(mutations));
   }
 
+  @Test
+  public void test_push_inexisting_buffer() {
```
anthonydahanne: test_push_inexisting_then_existing_buffer

management/monitoring-service/src/main/java/org/terracotta/management/service/monitoring/MonitoringService.java
```diff
@@ -60,44 +54,129 @@
...
+          throw new IllegalArgumentException("Protected buffer name: monitoring-tree-mutations");
+        }
+        return (ReadWriteBuffer<V>) consumers.get(callerConsumerID).computeIfAbsent(category, s -> new RingBuffer<V>(maxBufferSize) {
+          @Override
+          public void put(V value) {
+            if (!type.isInstance(value)) {
```
anthonydahanne: that seems like a reasonable check to have on all WriteRingBuffer
Why didn't you add it directly to RingBuffer.java in the first place ?

mathieucarbou: I hesitated in putting this check. I didn't do it because I see the ring biffer like a collection, and a collection is java just has a parameterized type but does not take a class as argument and do not do any check.

I really do not know if this check is something right to do... Also because the type could be in another classloader... So I even do not know if it is safe because of the way voltron isolate the classloaders between services and entities...

management/monitoring-service/src/test/java/org/terracotta/management/service/monitoring/MonitoringServiceTest.java
mathieucarbou: test is much more: `test_push_into_inexisting_buffer_do_not_fail`

management/monitoring-service/src/main/java/org/terracotta/management/service/monitoring/MonitoringServiceProvider.java
```diff
@@ -33,7 +32,8 @@
 
   private static final Collection<Class<?>> providedServiceTypes = Arrays.asList(
       IMonitoringConsumer.class,
-      IMonitoringProducer.class
+      IMonitoringProducer.class,
```
anthonydahanne: this  one does not seem to be used anymore ?
in getService() it is not returned anymore

mathieucarbou: Yes it is! Its the part where you declare which type of services your provider provides. And now we provide the consumer, or own producer (which extends voltron's one) and also voltron's producer. They have all to be there because the interface type is checked, not the "instance"

management/monitoring-service/src/test/java/org/terracotta/management/service/monitoring/RingBufferTest.java
```diff
@@ -0,0 +1,53 @@
...
+
+    buffer.put(3);
+    assertThat(buffer.size(), equalTo(2));
+
+    assertThat(buffer.read(), equalTo(2));
+    assertThat(buffer.size(), equalTo(1));
+    assertThat(buffer.read(), equalTo(3));
+    assertThat(buffer.isEmpty(), equalTo(true));
+    assertThat(buffer.size(), equalTo(0));
+
```
anthonydahanne:
```
    buffer.put(1);
    buffer.put(2);
    assertThat(buffer.size(), equalTo(2));
    List<Integer> collect = buffer.stream().collect(Collectors.toList());
    assertThat(collect.get(0), equalTo(1));
    assertThat(collect.get(1), equalTo(2));
```

anthonydahanne: I know stream() is used and tested elsewhere, but I thought they were missing here...

PR #144, opened 2016-09-01
Missing cluster topology when client notifications come and no topology change

mathieucarbou:
Missing cluster topology when client notifications come and no topology change


anthonydahanne:
LGTM

PR #147, opened 2016-09-08
#127 : ManagementRegistry as a Voltron service

### LINKED ISSUE #127: Put the ManagementRegistry as a Voltron service

mathieucarbou:
Content of this PR:

- Implementation of a voltron service which mimics the management registry, but for entities (closes #127 ). __No stats or management call supported, just management metadata information.__
- Fixes some missing serialVersionUid fields
- Better handling of path cleanup and update when a registry is exposed through the management entity
- Moving `management-registry` and `sequence-generator` modules at a common level, like `management-module` to facilitate the push of data containing Sequence, to avoid having to deal with Serializable[] objects and byte arrays representing the sequence


anthonydahanne:
LGTM

PR #148, opened 2016-09-14
marshalling lib with basic feature set

lorban:


runnel/src/main/java/org/terracotta/runnel/decoding/StructArrayDecoder.java
```diff
@@ -0,0 +1,185 @@
...
+/**
+ * @author Ludovic Orban
+ */
+public class StructArrayDecoder {
+  private final Field structField;
+  private final List<? extends Field> metadata;
+  private final ByteBuffer byteBuffer;
+  private final StructDecoder parent;
+  private final int size;
+  private int maxSize;
```
ChrisGreenaway: This could be final, I think.

runnel/src/main/java/org/terracotta/runnel/decoding/StructArrayDecoder.java
```diff
@@ -0,0 +1,185 @@
...
+      return null;
+    } else if (index != field.index()) {
+      return null;
+    } else {
+      return field;
+    }
+  }
+
+  private Field findMetadataFor(int index) {
+    for (Field field : metadata) {
```
ChrisGreenaway: I don't like that for each call to, for example, int64(String name) you iterate through all the fields - kind of O(n^2).

There should probably be methods like int64(int index) so that there's no need to call findMetadataFor - just use metadata.get(index) instead.

runnel/src/main/java/org/terracotta/runnel/decoding/StructArrayDecoder.java
```diff
@@ -0,0 +1,185 @@
...
+      if (field.index() == index) {
```
ChrisGreenaway: I think the API isn't quite right - because you've got string names for fields, it means you have to map between names and indexes. If your API just took an index everywhere there's a name, then the code would be simpler and faster. You could argue that it isn't as easy to use and API, but I think that with a few:

```
private static final int FIRST_NAME = 0;
private static final int LAST_NAME = 1;
```

declarations then the API is still just as readable.

runnel/src/main/java/org/terracotta/runnel/metadata/Field.java
```diff
@@ -0,0 +1,34 @@
...
+
+import org.terracotta.runnel.utils.ReadBuffer;
+
+import java.nio.ByteBuffer;
+import java.util.List;
+
+/**
+ * @author Ludovic Orban
+ */
+public interface Field {
```
mathieucarbou: Is it possible to explain a little what a `Field` is/represents and why it can have sub fields ? I.e. how an `Int64Field` can have some sub fields ? Thanks :-)

runnel/src/test/java/org/terracotta/runnel/ByteBufferTest.java
```diff
@@ -0,0 +1,82 @@
...
+  public void testStructWithByteBuffer() throws Exception {
+    Struct struct = StructBuilder.newStructBuilder()
+        .string("name", 5)
+        .byteBuffer("blob", 10)
+        .build();
+
+    ByteBuffer encoded = struct.encoder()
+        .string("name", "john doe")
+        .byteBuffer("blob", buffer(128))
+        .encode();
```
mathieucarbou: Is there a way to encode "null" ? i.e. if name is null, something like: `.nul("name")` ?

runnel/src/test/java/org/terracotta/runnel/ByteBufferTest.java
```diff
@@ -0,0 +1,82 @@
...
+ * @author Ludovic Orban
+ */
+public class ByteBufferTest {
+
+  @Test
+  public void testStructWithByteBuffer() throws Exception {
+    Struct struct = StructBuilder.newStructBuilder()
+        .string("name", 5)
+        .byteBuffer("blob", 10)
+        .build();
```
mathieucarbou: I really prefer the way Kryo deals with indexes because it does not force the user to declare indexes. Indexes are internal things. True, they should be overridable for backward compat, but from a user perspective, it would be easier to generate indexes, like kryo does, and provide a mechanism to override them if needed.

So it could end up like a protobuf file, declaring a structure medatata as a file, json or else:

`Struct struct = StructBuilder.newStruct(File yaml)`

and the yaml file:

```
name: string
blob: byteBuffer
```

etc. This yaml file can even be sent across the wire so that the client knows how to decode.

It's ordered, so indexes are easy to generate, and overridable thanks to the extensible yaml format if some indexes need to be forced for backward compat.

... some ideas ;-)

runnel/src/test/java/org/terracotta/runnel/PrimitiveArrayStructBuilderTest.java
```diff
@@ -0,0 +1,151 @@
...
+      .int64("age", 5)
+      .build();
+
+  @Test
+  public void testReadAll() throws Exception {
+    ByteBuffer bb = ByteBuffer.allocate(1024);
+
+    struct.encoder()
+        .string("name", "joe")
+        .int64s("ids")
```
mathieucarbou: I would prefer" `.int64s("ids", 1, 2, 3, 4)`, strings("colors", "red", "green", "blue")

mathieucarbou:
Global note: it would be nice to have an object mapper with some custom codec extensions (i.e. to serialize / deserialize enums as string for example)

runnel/src/main/java/org/terracotta/runnel/decoding/StructArrayDecoder.java
```diff
@@ -0,0 +1,185 @@
...
+    currentlyRead += (after - before);
+    return decoded;
+  }
+
+  public Long int64(String name) {
+    Int64Field field = findField(name, Int64Field.class);
+    if (field == null) {
+      return null;
+    }
+    int before = byteBuffer.position();
```
ChrisGreenaway: Can you spin these four lines of code out and reuse them across all these decoding methods?

runnel/src/main/java/org/terracotta/runnel/decoding/StructDecoder.java
```diff
@@ -0,0 +1,209 @@
...
+    this.parent = parent;
+    currentlyRead += 4;
+  }
+
+  public Integer int32(String name) {
+    Int32Field field = findField(name, Int32Field.class);
+    if (field == null) {
+      return null;
+    }
+    int before = byteBuffer.position();
```
ChrisGreenaway: You've got these same four lines of code again - please refactor.

It's a pain that it modifies currentlyRead because that means you'll need a base class.

runnel/src/main/java/org/terracotta/runnel/decoding/StructDecoder.java
```diff
@@ -0,0 +1,209 @@
...
+    if (parent == null) {
+      throw new RuntimeException("Cannot end root decoder");
+    }
+
+    byteBuffer.position(byteBuffer.position() + (maxSize - currentlyRead));
+
+    return parent;
+  }
+
+  private <F extends Field> F findField(String name, Class<F> fieldClazz) {
```
ChrisGreenaway: A lot of this code is in common with StructArrayDecoder - please refactor.

runnel/src/main/java/org/terracotta/runnel/encoding/StructEncoder.java
```diff
@@ -0,0 +1,181 @@
...
+  }
+
+  public StructEncoder end() {
+    if (parent == null) {
+      throw new RuntimeException("Cannot end root encoder");
+    }
+    return parent;
+  }
+
+  public ArrayEncoder<Integer> int32s(String name) {
```
ChrisGreenaway: Please refactor int32s and int64s etc to share common code.

runnel/src/main/java/org/terracotta/runnel/encoding/StructEncoder.java
```diff
@@ -0,0 +1,181 @@
...
+        if (field.getClass() != clazz) {
+          throw new RuntimeException("Invalid type for field '" + name + "', expected : '" + clazz.getSimpleName() + "' but was '" + field.getClass().getSimpleName() + "'");
+        }
+        return field;
+      }
+    }
+    throw new RuntimeException("No such field left : '" + name + "'");
+  }
+
+  public ByteBuffer encode(ByteBuffer bb) {
```
ChrisGreenaway: These two methods are identical except for where the ByteBuffer comes from. Think about using a strategy pattern (a bit like guava's Supplier) to factor out the commonality.

runnel/src/main/java/org/terracotta/runnel/metadata/ArrayField.java
```diff
@@ -0,0 +1,55 @@
...
+  }
+
+  @Override
+  public List<? extends Field> subFields() {
+    return Collections.singletonList(arrayedField);
+  }
+
+  @Override
+  public Object decode(ByteBuffer byteBuffer) {
+    throw new UnsupportedOperationException("todo");
```
ChrisGreenaway: You've got a todo here and one more later - are you intending to implement them before you merge?

runnel/src/main/java/org/terracotta/runnel/metadata/Field.java
```diff
@@ -0,0 +1,33 @@
...
+
+  String name();
+  int index();
+  boolean isComposite();
```
ChrisGreenaway: Minor: inconsistent spacing.

runnel/src/main/java/org/terracotta/runnel/metadata/Int32Field.java
```diff
@@ -0,0 +1,39 @@
...
+ * limitations under the License.
+ */
+package org.terracotta.runnel.metadata;
+
+import java.nio.ByteBuffer;
+
+/**
+ * @author Ludovic Orban
+ */
+public class Int32Field extends AbstractField {
```
ChrisGreenaway: It might be worth having a FixedWidthField base class for Int32Field and Int64Field. The skip() method is essentially the same. That fixed width abstraction can probably be applied other places too.

runnel/src/main/java/org/terracotta/runnel/dataholders/StringDataHolder.java
```diff
@@ -0,0 +1,46 @@
...
+
+  public int size(boolean withIndex) {
+    return value.length() * 2 + 4 + (withIndex ? 4 : 0);
+  }
+
+  public void encode(ByteBuffer byteBuffer, boolean withIndex) {
+    if (withIndex) {
+      byteBuffer.putInt(index);
+    }
+    byteBuffer.putInt(value.length() * 2);
```
ChrisGreenaway: At some point we will need to worry about chars that don't fit in two bytes.

runnel/src/main/java/org/terracotta/runnel/decoding/StructArrayDecoder.java
lorban: Yes, I know about this. It's one of the things I plan on reworking.

runnel/src/main/java/org/terracotta/runnel/metadata/Field.java
lorban: `Field` is the representation of a structure's field from the point of view of decoding. `Int64Field` cannot/should not have sub-fields, but `StructField` and `ArrayField` can.

runnel/src/test/java/org/terracotta/runnel/ByteBufferTest.java
lorban: Yes: you simply skip a field and it'll be decoded as null.

runnel/src/main/java/org/terracotta/runnel/metadata/Field.java
mathieucarbou: Ok, So it's not a field, its a `Type` right ? a Struct is a type, an int64 is a type, etc ?

runnel/src/test/java/org/terracotta/runnel/ByteBufferTest.java
lorban: Forcing to specify indices is really a core idea: this allows the encoded structures to be parsed even with only partial structure metadata.

Backward/forward compatibility between different versions of the encoding and decoding routines is critical, so forcing the indices to be specified *and stay constant over time* is the best way I've found to enforce that.

runnel/src/test/java/org/terracotta/runnel/PrimitiveArrayStructBuilderTest.java
lorban: I considered that but dropped the idea for a few reasons:
 - That would require building an array and filling it in before passing that to the encoder. That means more allocations and more garbage.
 - Most of the time, you won't know all the `id`'s upfront: it usually involves iterating some structure to get them. It's usually nicer to be able to fill them in one after the other.

runnel/src/main/java/org/terracotta/runnel/decoding/StructDecoder.java
lorban: Yes, that's on my todo list.

runnel/src/main/java/org/terracotta/runnel/encoding/StructEncoder.java
lorban: Yes, that's on my todo list.

runnel/src/main/java/org/terracotta/runnel/encoding/StructEncoder.java
lorban: Maybe. I'll see during my refactoring sessions what's the best way to get rid of this code duplication.

runnel/src/main/java/org/terracotta/runnel/metadata/Int32Field.java
lorban: I think I'm going to drop the `skip()` method, but we'll re-discuss that once I'm done with all my planned refactorings.

runnel/src/main/java/org/terracotta/runnel/dataholders/StringDataHolder.java
lorban: I coded this trivial-but-mediocre string encoding in a rush, I'll probably use UTF-8 encoding in the final code.

runnel/src/main/java/org/terracotta/runnel/decoding/StructArrayDecoder.java
lorban: With a pair of maps (name -> index and index -> name) we can lower the cost of using names to almost zero, so I'm not convinced this wouldn't hurt the usability of the API, especially since this would require defining a set of constants on top of the structures.

lorban:
@mathieucarbou an objet mapper is completely outside the scope of this lib. The scope is to provide a lib that allows writing performant *and* backward/forward compatible codecs for our clustered entities with minimal effort.

But support for other types is planned: enum, boolean, int8, int16, float and double are candidates to make it in.

runnel/src/main/java/org/terracotta/runnel/metadata/Field.java
lorban: `Int64Field` is a field of type int64. :-)

PR #151, opened 2016-09-16
marshalling lib with basic feature set

lorban:
This is the (un-)marshalling lib we're going to use to write codecs. At this moment, it supports the following types: int32, int64, ByteBuffer, String, array and structures but more will come (double, enums...) and can easily be added.

### DIFF
--- a/runnel/src/main/java/org/terracotta/runnel/Enm.java
+++ b/runnel/src/main/java/org/terracotta/runnel/Enm.java
@@ -37,7 +37,7 @@ public class Enm<E extends Enum<E>> {
     return integer;
   }
 
-  public Enum toEnum(int intValue) {
+  public E toEnum(int intValue) {
     return integerToEnum.get(intValue);
   }
 }
--- a/runnel/src/main/java/org/terracotta/runnel/Enm.java
+++ b/runnel/src/main/java/org/terracotta/runnel/Enm.java
@@ -29,10 +29,6 @@ public class Enm<E extends Enum<E>> {
     this.integerToEnum = integerToEnum;
   }
 
-  public boolean maps(E e) {
-    return enumToInteger.containsKey(e);
-  }
-
   public int toInt(E e) {
     Integer integer = enumToInteger.get(e);
     if (integer == null) {
--- a/runnel/src/main/java/org/terracotta/runnel/Enm.java
+++ b/runnel/src/main/java/org/terracotta/runnel/Enm.java
@@ -1,0 +1,47 @@
+/*
+ * Copyright Terracotta, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.terracotta.runnel;
+
+import java.util.Map;
+
+/**
+ * @author Ludovic Orban
+ */
+public class Enm<E extends Enum<E>> {
+  private final Map<E, Integer> enumToInteger;
+  private final Map<Integer, E> integerToEnum;
+
+  Enm(Map<E, Integer> enumToInteger, Map<Integer, E> integerToEnum) {
+    this.enumToInteger = enumToInteger;
+    this.integerToEnum = integerToEnum;
+  }
+
+  public boolean maps(E e) {
+    return enumToInteger.containsKey(e);
+  }
+
+  public int toInt(E e) {
+    Integer integer = enumToInteger.get(e);
+    if (integer == null) {
+      throw new IllegalArgumentException("Unmapped enum : " + e);
+    }
+    return integer;
+  }
+
+  public Enum toEnum(int intValue) {
+    return integerToEnum.get(intValue);
+  }
+}
--- a/runnel/src/main/java/org/terracotta/runnel/EnmBuilder.java
+++ b/runnel/src/main/java/org/terracotta/runnel/EnmBuilder.java
@@ -41,6 +41,9 @@ public class EnmBuilder<E extends Enum<E>> {
     if (enumToInteger.containsKey(e)) {
       throw new IllegalArgumentException("Duplicate enum value : " + e);
     }
+    if (value < 0) {
+      throw new IllegalArgumentException("Int value must be >= 0, got : " + value);
+    }
     if (integerToEnum.containsKey(value)) {
       throw new IllegalArgumentException("Duplicate int value : " + value);
     }
--- a/runnel/src/main/java/org/terracotta/runnel/EnmBuilder.java
+++ b/runnel/src/main/java/org/terracotta/runnel/EnmBuilder.java
@@ -15,7 +15,10 @@
  */
 package org.terracotta.runnel;
 
+import java.util.EnumMap;
+import java.util.EnumSet;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.Map;
 
 /**
@@ -23,17 +26,25 @@ import java.util.Map;
  */
 public class EnmBuilder<E extends Enum<E>> {
 
-  private final Map<E, Integer> enumToInteger = new HashMap<E, Integer>();
+  private final EnumMap<E, Integer> enumToInteger;
   private final Map<Integer, E> integerToEnum = new HashMap<Integer, E>();
+  private final Class<E> enumClass;
 
-  private EnmBuilder() {
+  private EnmBuilder(Class<E> enumClass) {
+    this.enumClass = enumClass;
+    this.enumToInteger = new EnumMap<E, Integer>(enumClass);
   }
 
-  public static <E extends Enum<E>> EnmBuilder<E> newEnumBuilder() {
-    return new EnmBuilder<E>();
+  public static <E extends Enum<E>> EnmBuilder<E> newEnumBuilder(Class<E> enumClass) {
+    return new EnmBuilder<E>(enumClass);
   }
 
   public Enm<E> build() {
+    HashSet<E> unregisteredEnums = new HashSet<E>(EnumSet.allOf(enumClass));
+    unregisteredEnums.removeAll(enumToInteger.keySet());
+    if (!unregisteredEnums.isEmpty()) {
+      throw new IllegalStateException("Missing enum mappings for : " + unregisteredEnums);
+    }
     return new Enm<E>(enumToInteger, integerToEnum);
   }
 
--- a/runnel/src/main/java/org/terracotta/runnel/EnmBuilder.java
+++ b/runnel/src/main/java/org/terracotta/runnel/EnmBuilder.java
@@ -1,0 +1,52 @@
+/*
+ * Copyright Terracotta, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.terracotta.runnel;
+
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * @author Ludovic Orban
+ */
+public class EnmBuilder<E extends Enum<E>> {
+
+  private final Map<E, Integer> enumToInteger = new HashMap<E, Integer>();
+  private final Map<Integer, E> integerToEnum = new HashMap<Integer, E>();
+
+  private EnmBuilder() {
+  }
+
+  public static <E extends Enum<E>> EnmBuilder<E> newEnumBuilder() {
+    return new EnmBuilder<E>();
+  }
+
+  public Enm<E> build() {
+    return new Enm<E>(enumToInteger, integerToEnum);
+  }
+
+  public EnmBuilder<E> mapping(E e, int value) {
+    if (enumToInteger.containsKey(e)) {
+      throw new IllegalArgumentException("Duplicate enum value : " + e);
+    }
+    if (integerToEnum.containsKey(value)) {
+      throw new IllegalArgumentException("Duplicate int value : " + value);
+    }
+    enumToInteger.put(e, value);
+    integerToEnum.put(value, e);
+    return this;
+  }
+
+}
--- a/runnel/src/main/java/org/terracotta/runnel/Struct.java
+++ b/runnel/src/main/java/org/terracotta/runnel/Struct.java
@@ -19,6 +19,7 @@ import org.terracotta.runnel.decoding.StructDecoder;
 import org.terracotta.runnel.decoding.fields.ArrayField;
 import org.terracotta.runnel.decoding.fields.Field;
 import org.terracotta.runnel.decoding.fields.StructField;
+import org.terracotta.runnel.decoding.fields.ValueField;
 import org.terracotta.runnel.encoding.StructEncoder;
 import org.terracotta.runnel.utils.ReadBuffer;
 import org.terracotta.runnel.utils.VLQ;
@@ -97,10 +98,11 @@ public class Struct {
         dumpField(subField, readBuffer, out, depth + 1);
       }
     } else if (field != null) {
+      ValueField valueField = (ValueField) field;
       out.print(" type: "); out.print(field.getClass().getSimpleName());
       out.print(" name: "); out.print(field.name());
       parentBuffer.rewind(VLQ.encodedSize(fieldSize));
-      Object decoded = field.decode(readBuffer);
+      Object decoded = valueField.decode(readBuffer);
       out.print(" decoded: ["); out.print(decoded); out.print("]");
     } else {
       out.print(" type: ???");
--- a/runnel/src/main/java/org/terracotta/runnel/Struct.java
+++ b/runnel/src/main/java/org/terracotta/runnel/Struct.java
@@ -1,0 +1,111 @@
+/*
+ * Copyright Terracotta, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.terracotta.runnel;
+
+import org.terracotta.runnel.decoding.StructDecoder;
+import org.terracotta.runnel.decoding.fields.ArrayField;
+import org.terracotta.runnel.decoding.fields.Field;
+import org.terracotta.runnel.decoding.fields.StructField;
+import org.terracotta.runnel.encoding.StructEncoder;
+import org.terracotta.runnel.utils.ReadBuffer;
+import org.terracotta.runnel.utils.VLQ;
+
+import java.io.PrintStream;
+import java.nio.ByteBuffer;
+import java.util.Map;
+
+/**
+ * @author Ludovic Orban
+ */
+public class Struct {
+  private final StructField root;
+
+  public Struct(StructField root) {
+    this.root = root;
+  }
+
+  StructField getRoot() {
+    return root;
+  }
+
+  public StructEncoder encoder() {
+    return new StructEncoder(root);
+  }
+
+  public StructDecoder decoder(ByteBuffer byteBuffer) {
+    return new StructDecoder(root, new ReadBuffer(byteBuffer));
+  }
+
+  public void dump(ByteBuffer byteBuffer, PrintStream out) {
+    Map<Integer, Field> fieldsByInteger = root.getMetadata().buildFieldsByIndexMap();
+
+    ReadBuffer readBuffer = new ReadBuffer(byteBuffer);
+    int totalSize = readBuffer.getVlqInt();
+    readBuffer = readBuffer.limit(totalSize);
+
+    while (!readBuffer.limitReached()) {
+      int index = readBuffer.getVlqInt();
+      out.print("index: "); out.print(index);
+      Field field = fieldsByInteger.get(index);
+      dumpField(field, readBuffer, out, 0);
+      out.print("\n");
+    }
+  }
+
+  private void dumpField(Field field, ReadBuffer parentBuffer, PrintStream out, int depth) {
+    int fieldSize = parentBuffer.getVlqInt();
+    out.print(" size: "); out.print(fieldSize);
+    ReadBuffer readBuffer = parentBuffer.limit(fieldSize);
+
+    if (field instanceof ArrayField) {
+      out.print(" type: "); out.print(field.getClass().getSimpleName());
+
+      int length = readBuffer.getVlqInt();
+      out.print(" length: "); out.print(length);
+
+      ArrayField arrayField = (ArrayField) field;
+      Field subField = arrayField.subFields().get(0);
+
+      for (int i = 0; i < length; i++) {
+        out.print("\n  "); for (int j = 0; j < depth; j++) out.print("  ");
+        dumpField(subField, readBuffer, out, depth + 1);
+      }
+    } else if (field instanceof StructField) {
+      out.print(" type: "); out.print(field.getClass().getSimpleName());
+      out.print(" name: "); out.print(field.name());
+
+      StructField structField = (StructField) field;
+      Map<Integer, Field> fieldsByInteger = structField.getMetadata().buildFieldsByIndexMap();
+      while (!readBuffer.limitReached()) {
+        out.print("\n  "); for (int j = 0; j < depth; j++) out.print("  ");
+        int index = readBuffer.getVlqInt();
+        out.print(" index: "); out.print(index);
+        Field subField = fieldsByInteger.get(index);
+        dumpField(subField, readBuffer, out, depth + 1);
+      }
+    } else if (field != null) {
+      out.print(" type: "); out.print(field.getClass().getSimpleName());
+      out.print(" name: "); out.print(field.name());
+      parentBuffer.rewind(VLQ.encodedSize(fieldSize));
+      Object decoded = field.decode(readBuffer);
+      out.print(" decoded: ["); out.print(decoded); out.print("]");
+    } else {
+      out.print(" type: ???");
+      readBuffer.skipAll();
+    }
+  }
+
+}
--- a/runnel/src/main/java/org/terracotta/runnel/Struct.java
+++ b/runnel/src/main/java/org/terracotta/runnel/Struct.java
@@ -77,7 +77,7 @@ public class Struct {
       out.print(" length: "); out.print(length);
 
       ArrayField arrayField = (ArrayField) field;
-      Field subField = arrayField.subFields().get(0);
+      Field subField = arrayField.subField();
 
       for (int i = 0; i < length; i++) {
         out.print("\n  "); for (int j = 0; j < depth; j++) out.print("  ");
--- a/runnel/src/main/java/org/terracotta/runnel/Struct.java
+++ b/runnel/src/main/java/org/terracotta/runnel/Struct.java
@@ -26,6 +26,7 @@ import org.terracotta.runnel.utils.VLQ;
 
 import java.io.PrintStream;
 import java.nio.ByteBuffer;
+import java.util.List;
 import java.util.Map;
 
 /**
@@ -38,8 +39,8 @@ public class Struct {
     this.root = root;
   }
 
-  StructField getRoot() {
-    return root;
+  List<? extends Field> getRootSubFields() {
+    return root.subFields();
   }
 
   public StructEncoder encoder() {
--- a/runnel/src/main/java/org/terracotta/runnel/Struct.java
+++ b/runnel/src/main/java/org/terracotta/runnel/Struct.java
@@ -16,13 +16,10 @@
 package org.terracotta.runnel;
 
 import org.terracotta.runnel.decoding.StructDecoder;
-import org.terracotta.runnel.decoding.fields.ArrayField;
 import org.terracotta.runnel.decoding.fields.Field;
 import org.terracotta.runnel.decoding.fields.StructField;
-import org.terracotta.runnel.decoding.fields.ValueField;
 import org.terracotta.runnel.encoding.StructEncoder;
 import org.terracotta.runnel.utils.ReadBuffer;
-import org.terracotta.runnel.utils.VLQ;
 
 import java.io.PrintStream;
 import java.nio.ByteBuffer;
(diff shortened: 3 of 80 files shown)


runnel/src/main/java/org/terracotta/runnel/decoding/StructDecoder.java
```diff
@@ -0,0 +1,153 @@
...
+
+  private StructDecoder(StructField structField, ReadBuffer readBuffer, StructDecoder parent) {
+    this.fieldSearcher = structField.getMetadata().fieldSearcher();
+    this.parent = parent;
+    int size = readBuffer.getVlqInt();
+    this.readBuffer = readBuffer.limit(size);
+  }
+
+  @Override
+  public Integer int32(String name) {
```
chrisdennis: Looks like with some generics on Field, we could eliminate all the redundant null checking and decode calling here.  Also no point in handing `null` and `readBuffer` from every method.

runnel/src/main/java/org/terracotta/runnel/decoding/fields/Int32Field.java
```diff
@@ -0,0 +1,35 @@
...
+ */
+public class Int32Field extends AbstractField {
+
+  public Int32Field(String name, int index) {
+    super(name, index);
+  }
+
+  @Override
+  public Object decode(ReadBuffer readBuffer) {
+    readBuffer.getVlqInt();
```
chrisdennis: If we're going to write lengths for all the primitives shouldn't we at least validate them on decode?  If not, could we make them optional?

runnel/src/main/java/org/terracotta/runnel/encoding/StructArrayEncoder.java
```diff
@@ -0,0 +1,104 @@
...
+
+  StructArrayEncoder(List<StructDataHolder> values, StructEncoder parent, StructField structField) {
+    this.values = values;
+    this.parent = parent;
+    this.fieldSearcher = structField.getMetadata().fieldSearcher();
+    this.currentData = new ArrayList<DataHolder>(ARRAY_INITIAL_SIZE);
+  }
+
+  @Override
+  public StructArrayEncoder int32(String name, int value) {
```
chrisdennis: Lots of duplicated code here that it feels like we should be able to eliminate.

runnel/src/main/java/org/terracotta/runnel/encoding/dataholders/ByteBufferDataHolder.java
```diff
@@ -0,0 +1,48 @@
...
+  }
+
+  public int size(boolean withIndex) {
+    int len = value.remaining();
+    return VLQ.encodedSize(len) + len + (withIndex ? VLQ.encodedSize(index) : 0);
+  }
+
+  public void encode(WriteBuffer writeBuffer, boolean withIndex) {
+    if (withIndex) {
+      writeBuffer.putVlqInt(index);
```
chrisdennis: This stanza appears all over the place (along with it's equivalent in the the size method).  Could we introduce a method in some superclass to represent the commonality?

runnel/src/main/java/org/terracotta/runnel/utils/BufferLimitReachedException.java
```diff
@@ -0,0 +1,22 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.terracotta.runnel.utils;
+
+/**
+ * @author Ludovic Orban
+ */
+public class BufferLimitReachedException extends RuntimeException {
```
chrisdennis: Could we just use `IndexOutOfBoundsException` instead?

runnel/src/main/java/org/terracotta/runnel/utils/ReadBuffer.java
```diff
@@ -0,0 +1,139 @@
...
+import java.nio.ByteBuffer;
+import java.nio.CharBuffer;
+import java.nio.charset.Charset;
+
+/**
+ * @author Ludovic Orban
+ */
+public class ReadBuffer {
+
+  private static final Charset UTF8 = Charset.forName("UTF-8");
```
chrisdennis: Given TCs rich history of stuffing weird things inside Strings we might want to consider using the same modified UTF-8 encoding we do in the Ehcache serializer.

runnel/src/main/java/org/terracotta/runnel/utils/VLQ.java
```diff
@@ -0,0 +1,62 @@
...
+ * limitations under the License.
+ */
+package org.terracotta.runnel.utils;
+
+import java.nio.ByteBuffer;
+
+/**
+ * @author Ludovic Orban
+ */
+public class VLQ {
```
chrisdennis: Any reason we're not using VLQ for all integers?  (maybe something with a more compact negative int representation for likely negatives)

ChrisGreenaway: For general data, there's no space saved using a VLQ. It only helps when you know that the integer is likely to be small. If you know something else about the structure of the data then you could use some other encoding that was more compact. The general solution to this, where the data has some structure, but you don't know what, is compression.

lorban: as @ChrisGreenaway said. :)

runnel/src/main/java/org/terracotta/runnel/utils/BufferLimitReachedException.java
lorban: `IndexOutOfBoundsException` would be misleading here, or at least hint about the internals.

At the moment, the implementation uses `ByteBuffer` but everything is designed to work with streams... in which an index wouldn't make sense.

runnel/src/main/java/org/terracotta/runnel/StructBuilder.java
```diff
@@ -0,0 +1,134 @@
...
+
+import java.util.ArrayList;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+
+/**
+ * @author Ludovic Orban
+ */
+public class StructBuilder {
```
lorban: Indeed, it looks like I started adding name checks in this class but forgot to finish the work. Let me clean that up.

runnel/src/main/java/org/terracotta/runnel/decoding/StructDecoder.java
lorban: What kind of generics wizardry do you have in mind?

runnel/src/main/java/org/terracotta/runnel/decoding/fields/Int32Field.java
lorban: I could add a check to assert the length, and maybe introduce a new `StreamCorruptedException`. But the length is never optional, even on primitive fields for which the length is fixed for forward/backward compatibility reasons.

runnel/src/main/java/org/terracotta/runnel/encoding/StructArrayEncoder.java
lorban: Like what?

runnel/src/main/java/org/terracotta/runnel/encoding/dataholders/ByteBufferDataHolder.java
lorban: I'm not convinced extracting the following bit of code:

```
    if (withIndex) {
      writeBuffer.putVlqInt(index);
    }
```

into an abstract super class is worth it. Are you?

lorban:
test this please

runnel/src/main/java/org/terracotta/runnel/utils/VLQ.java
```diff
@@ -0,0 +1,62 @@
...
+
+  public static void encode(int value, ByteBuffer out) {
+    if (value < 0) {
+      throw new IllegalArgumentException("Cannot encode negative values");
+    }
+    while ((value & 0xFFFFFF80) != 0L) {
```
ChrisGreenaway: Did you get this VLQ algorithm from somewhere? If so, can you reference the source in a comment?

runnel/src/main/java/org/terracotta/runnel/utils/WriteBuffer.java
```diff
@@ -0,0 +1,51 @@
...
+public class WriteBuffer {
+
+  private final ByteBuffer byteBuffer;
+
+  public WriteBuffer(ByteBuffer byteBuffer) {
+    this.byteBuffer = byteBuffer;
+  }
+
+  public void putDouble(double value) {
+    byteBuffer.putLong(Double.doubleToRawLongBits(value));
```
ChrisGreenaway: What's wrong with byteBuffer.putDouble()?

runnel/src/main/java/org/terracotta/runnel/Enm.java
```diff
@@ -0,0 +1,47 @@
...
+
+/**
+ * @author Ludovic Orban
+ */
+public class Enm<E extends Enum<E>> {
+  private final Map<E, Integer> enumToInteger;
+  private final Map<Integer, E> integerToEnum;
+
+  Enm(Map<E, Integer> enumToInteger, Map<Integer, E> integerToEnum) {
+    this.enumToInteger = enumToInteger;
```
ChrisGreenaway: Should we have a check here (or in the builder) that all the enum values are specified?

runnel/src/main/java/org/terracotta/runnel/EnmBuilder.java
```diff
@@ -0,0 +1,52 @@
...
+
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * @author Ludovic Orban
+ */
+public class EnmBuilder<E extends Enum<E>> {
+
+  private final Map<E, Integer> enumToInteger = new HashMap<E, Integer>();
```
ChrisGreenaway: EnumMap would be faster.

runnel/src/main/java/org/terracotta/runnel/Enm.java
```diff
@@ -0,0 +1,47 @@
...
+    this.integerToEnum = integerToEnum;
```
ChrisGreenaway: In theory you could test these maps to pick up the case where the mapping is enum<->ordinal and then use a faster implementation for that common case.

runnel/src/main/java/org/terracotta/runnel/encoding/dataholders/EnmDataHolder.java
```diff
@@ -24,28 +24,24 @@
...
 public class EnmDataHolder<E extends Enum<E>> implements DataHolder {
 
-  private final E value;
+  private final int value;
   private final int index;
-  private final Enm<E> enm;
 
   public EnmDataHolder(E value, int index, Enm<E> enm) {
-    this.value = value;
+    this.value = enm.toInt(value);
```
ChrisGreenaway: What happens if enum.toInt(value) returns null?

runnel/src/main/java/org/terracotta/runnel/utils/ReadBuffer.java
```diff
@@ -0,0 +1,139 @@
...
+      throw new LimitReachedException();
+    }
+  }
+
+  public Double getDouble() {
+    if (byteBuffer.position() + 8 > limit) {
+      throw new LimitReachedException();
+    }
+    long aLong = byteBuffer.getLong();
+    return Double.longBitsToDouble(aLong);
```
lorban: Nothing... I just missed it! Let me change that.

runnel/src/main/java/org/terracotta/runnel/metadata/Metadata.java
```diff
@@ -0,0 +1,61 @@
...
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * @author Ludovic Orban
+ */
+public class Metadata {
+
+  // yes, you need this weirdo holder class because every Field.index() call is megamorphic, hence very expensive
```
lorban: Yes, that trickery improves perf by about 15%.

runnel/src/main/java/org/terracotta/runnel/utils/WriteBuffer.java
lorban: Same as above, I missed that method. Let me change this here too.

runnel/src/main/java/org/terracotta/runnel/Enm.java
lorban: That might be desirable. Let me consider it.

runnel/src/main/java/org/terracotta/runnel/encoding/dataholders/EnmDataHolder.java
lorban: This cannot happens, `enm.toInt()` would throw `IllegalArgumentException` in that case.

runnel/src/main/java/org/terracotta/runnel/utils/VLQ.java
lorban: The encoding is clearly described in https://en.wikipedia.org/wiki/Variable-length_quantity so I just wrote this code myself.

runnel/src/main/java/org/terracotta/runnel/metadata/FieldSearcher.java
```diff
@@ -0,0 +1,94 @@
...
+      int fieldSize = readBuffer.getVlqInt();
+      readBuffer.skip(fieldSize);
+      if (readBuffer.limitReached()) {
+        return null;
+      }
+      index = readBuffer.getVlqInt();
+    }
+
+    if (index > fieldWithIndex.fieldIndex) {
+      readBuffer.rewind(VLQ.encodedSize(index));
```
lorban: I could certainly avoid the rewind by keeping the index around for the next `nextField()` call. I agree this would keep the streaming concept pure.

runnel/src/main/java/org/terracotta/runnel/utils/ReadBuffer.java
```diff
@@ -0,0 +1,139 @@
...
+  }
+
+  public String getString(int size) {
+    if (byteBuffer.position() + size > limit) {
+      throw new LimitReachedException();
+    }
+
+    ByteBuffer slice = byteBuffer.slice();
+    slice.limit(size);
+    CharBuffer decode = UTF8.decode(slice);
```
lorban: Or re-implement our special codec that can handle any char sequence, as @chrisdennis suggested?

runnel/src/main/java/org/terracotta/runnel/utils/ReadBuffer.java
```diff
@@ -0,0 +1,139 @@
...
+    this.origin = byteBuffer.position();
+    this.limit = origin + limit;
+    if (this.limit > byteBuffer.capacity()) {
+      throw new LimitReachedException();
+    }
+  }
+
+  public Double getDouble() {
+    if (byteBuffer.position() + 8 > limit) {
+      throw new LimitReachedException();
```
lorban: I want to abstract the fact that `ByteBuffer` is used internally as I suspect we will have to come up with a version that works with `InputStream` as well. That's why I want to throw a runnel-specific exception.

runnel/src/main/java/org/terracotta/runnel/utils/ReadBuffer.java
```diff
@@ -0,0 +1,139 @@
...
+  private final int origin;
+  private final int limit;
+
+  public ReadBuffer(ByteBuffer byteBuffer) {
+    this(byteBuffer, byteBuffer.position() + byteBuffer.capacity());
+  }
+
+  private ReadBuffer(ByteBuffer byteBuffer, int limit) {
+    this.byteBuffer = byteBuffer;
+    this.origin = byteBuffer.position();
```
lorban: Yes: when a structure contains a substructure it's very handy to carve a new `ReadBuffer` with a different limit that just allows working on the bytes of the substructure. You know where it ends when you're looking for an index in the stream and you know how many bytes to drop when you want to skip it.

runnel/src/main/java/org/terracotta/runnel/utils/ReadBuffer.java
```diff
@@ -0,0 +1,139 @@
...
+public class ReadBuffer {
+
+  private static final Charset UTF8 = Charset.forName("UTF-8");
+
+  private final ByteBuffer byteBuffer;
+  private final int origin;
+  private final int limit;
+
+  public ReadBuffer(ByteBuffer byteBuffer) {
+    this(byteBuffer, byteBuffer.position() + byteBuffer.capacity());
```
lorban: Good catch, this should be

```
this(byteBuffer, byteBuffer.remaining());
```

runnel/src/main/java/org/terracotta/runnel/Enm.java
lorban: This is very unlikely in practice so I don't think that's worth the effort.

runnel/src/main/java/org/terracotta/runnel/EnmBuilder.java
lorban: Agreed, let me change that.

runnel/src/main/java/org/terracotta/runnel/metadata/Metadata.java
```diff
@@ -0,0 +1,61 @@
...
+    FieldWithIndex(Field field) {
+      this.field = field;
+      this.fieldIndex = field.index();
+    }
+
+    Field field;
+    int fieldIndex;
+  }
+
+  final Map<String, FieldWithIndex> fieldsByName;
```
lorban: You're right.

runnel/src/main/java/org/terracotta/runnel/utils/ReadBuffer.java
ChrisGreenaway: If you use CharsetDecoder rather than Charset then there's more control. I don't think we should re-implement a special codec - we shouldn't be spending our time implementing extreme edge cases right now.

runnel/src/main/java/org/terracotta/runnel/utils/ReadBuffer.java
ChrisGreenaway: You misunderstand my comment. I wasn't questioning the value of creating a ReadBuffer with a limit. I was questioning what happens when the ByteBuffer passed to this constructor has a limit set.

runnel/src/main/java/org/terracotta/runnel/Enm.java
ChrisGreenaway: I think it's highly likely - isn't that the obvious mapping from enums to integers. But if you don't think it's worth the effort then I'm OK with that.

runnel/src/main/java/org/terracotta/runnel/decoding/fields/ArrayField.java
```diff
@@ -0,0 +1,45 @@
...
+  }
+
+  @Override
+  public List<? extends Field> subFields() {
+    return Collections.singletonList(arrayedField);
+  }
+
+  @Override
+  public Object decode(ReadBuffer readBuffer) {
+    throw new UnsupportedOperationException();
```
lorban: Agreed.

runnel/src/main/java/org/terracotta/runnel/decoding/fields/AbstractField.java
```diff
@@ -0,0 +1,47 @@
...
+    return name;
+  }
+
+  @Override
+  public int index() {
+    return index;
+  }
+
+  @Override
+  public List<? extends Field> subFields() {
```
lorban: Actually, only `StructField` has subfields, and `ArrayField` always has one subfield. With some generics use I could get rid of the `subFields()` method from the `Field` interface.

runnel/src/main/java/org/terracotta/runnel/utils/VLQ.java
ChrisGreenaway: Can you add a comment linking to that wikipedia page please?
Also, it looks to me as if the encode() method puts the least significant byte first whereas the wikipedia page says it's most significant byte first. Can you make this difference clear in the comment too please? Or, perhaps better, fix it to match the wikipedia description.
Also, add some tests - if you change the byte order then there are some test cases on the wikipedia page.

runnel/src/main/java/org/terracotta/runnel/decoding/StructArrayDecoder.java
```diff
@@ -0,0 +1,123 @@
...
+      int structSize = readBuffer.getVlqInt();
+      structReadBuffer = arrayReadBuffer.limit(structSize);
+    } else {
+      structReadBuffer = arrayReadBuffer.limit(0);
+    }
+  }
+
+  @Override
+  public Integer int32(String name) {
+    Int32Field field = fieldSearcher.nextField(name, Int32Field.class, null, structReadBuffer);
```
lorban: All that could move into the `FieldSearcher` class.

runnel/src/main/java/org/terracotta/runnel/Struct.java
```diff
@@ -0,0 +1,111 @@
...
+ */
+public class Struct {
+  private final StructField root;
+
+  public Struct(StructField root) {
+    this.root = root;
+  }
+
+  StructField getRoot() {
+    return root;
```
lorban: Ok, let me encapsulate that a bit better.

runnel/src/main/java/org/terracotta/runnel/Struct.java
```diff
@@ -0,0 +1,111 @@
...
+        out.print("\n  "); for (int j = 0; j < depth; j++) out.print("  ");
+        int index = readBuffer.getVlqInt();
+        out.print(" index: "); out.print(index);
+        Field subField = fieldsByInteger.get(index);
+        dumpField(subField, readBuffer, out, depth + 1);
+      }
+    } else if (field != null) {
+      out.print(" type: "); out.print(field.getClass().getSimpleName());
+      out.print(" name: "); out.print(field.name());
+      parentBuffer.rewind(VLQ.encodedSize(fieldSize));
```
lorban: Ok, sure.

runnel/src/main/java/org/terracotta/runnel/Struct.java
```diff
@@ -0,0 +1,111 @@
...
+    }
+  }
+
+  private void dumpField(Field field, ReadBuffer parentBuffer, PrintStream out, int depth) {
+    int fieldSize = parentBuffer.getVlqInt();
+    out.print(" size: "); out.print(fieldSize);
+    ReadBuffer readBuffer = parentBuffer.limit(fieldSize);
+
+    if (field instanceof ArrayField) {
+      out.print(" type: "); out.print(field.getClass().getSimpleName());
```
lorban: Ok, sure.

runnel/src/main/java/org/terracotta/runnel/encoding/StructEncoder.java
```diff
@@ -0,0 +1,208 @@
...
+
+  private StructEncoder(StructField structField, List<DataHolder> values, StructEncoder structEncoder) {
+    this.fieldSearcher = structField.getMetadata().fieldSearcher();
+    this.data = values;
+    this.parent = structEncoder;
+  }
+
+  @Override
+  public StructEncoder int32(String name, int value) {
+    int fieldIndex = fieldSearcher.findFieldIndex(name, Int32Field.class, null);
```
lorban: Done.

runnel/src/main/java/org/terracotta/runnel/encoding/dataholders/StructDataHolder.java
```diff
@@ -0,0 +1,70 @@
...
+  }
+
+  public void encode(WriteBuffer writeBuffer, boolean withIndex) {
+    if (withIndex) {
+      writeBuffer.putVlqInt(index);
+    }
+
+    int size = 0;
+    for (DataHolder value : values) {
+      size += value.size(true);
```
lorban: Ok, done.

runnel/src/main/java/org/terracotta/runnel/Struct.java
lorban: Let me get rid of `rewind()`.

runnel/src/main/java/org/terracotta/runnel/Enm.java
chrisdennis: Talked with @cschanck and @cljohnso and we are all in agreement that the ordinal pattern is so likely, that we should do away with custom enum mapping stuff entirely and just support using the ordinals.  It's faster, and everyone is used to having to keep ordinal values stable.

runnel/src/main/java/org/terracotta/runnel/Enm.java
```diff
@@ -0,0 +1,43 @@
...
+ * limitations under the License.
+ */
+package org.terracotta.runnel;
+
+import java.util.Map;
+
+/**
+ * @author Ludovic Orban
+ */
+public class Enm<E extends Enum<E>> {
```
chrisdennis: If we keep this class, then we should rename it - it's not an enum instance, it's an EnumMapping or an EnumEncoding object.

runnel/src/main/java/org/terracotta/runnel/EnmBuilder.java
```diff
@@ -0,0 +1,66 @@
...
+import java.util.EnumMap;
+import java.util.EnumSet;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Map;
+
+/**
+ * @author Ludovic Orban
+ */
+public class EnmBuilder<E extends Enum<E>> {
```
chrisdennis: This is an EnumMappingBuilder.

runnel/src/main/java/org/terracotta/runnel/StructBuilder.java
```diff
@@ -0,0 +1,145 @@
...
+    return new Struct(new StructField("root", -1, fields));
+  }
+
+  public static StructBuilder newStructBuilder() {
+    return new StructBuilder();
+  }
+
+
+  public StructBuilder enm(String name, int index, Enm enm) {
+    checkParams(name, index);
```
chrisdennis: We can reduce each one of these methods to a single line if we push all the commonality in to a single method.

runnel/src/main/java/org/terracotta/runnel/StructBuilder.java
```diff
@@ -0,0 +1,145 @@
...
+    fields.add(new ArrayField(name, index, new StructField(name, index, struct.getRootSubFields())));
+    return this;
+  }
+
+  private void checkParams(String name, int index) {
+    checkName(name);
+    checkIndex(index);
+  }
+
+  private void checkIndex(int index) {
```
chrisdennis: This is really `advanceIndex(int index)` since it's mutative.

runnel/src/main/java/org/terracotta/runnel/StructBuilder.java
```diff
@@ -0,0 +1,145 @@
...
+    if (index <= 0) {
+      throw new IllegalArgumentException("Index must be greater than zero : " + index);
+    }
+    if (index <= lastIndex) {
+      throw new IllegalArgumentException("Index must be registered in growing order : " + index + " (last registered one : " + lastIndex + ")");
+    }
+    lastIndex = index;
+  }
+
+  private void checkName(String name) {
```
chrisdennis: This should be called `addName(String name)` since it's mutative.  Also you can achieve this with a single `names.add(name)` call.

runnel/src/main/java/org/terracotta/runnel/decoding/fields/AbstractValueField.java
```diff
@@ -0,0 +1,38 @@
...
+ */
+public abstract class AbstractValueField<T> extends AbstractField implements ValueField<T> {
+
+  protected AbstractValueField(String name, int index) {
+    super(name, index);
+  }
+
+  @Override
+  public void dump(ReadBuffer readBuffer, PrintStream out, int depth) {
+    out.print(" type: "); out.print(getClass().getSimpleName());
```
chrisdennis: Looks like you wanted to use append here.

runnel/src/main/java/org/terracotta/runnel/encoding/StructArrayEncoder.java
```diff
@@ -0,0 +1,113 @@
...
+    this.values = values;
+    this.parent = parent;
+    this.fieldSearcher = structField.getMetadata().fieldSearcher();
+    this.currentData = new ArrayList<DataHolder>(ARRAY_INITIAL_SIZE);
+  }
+
+  @Override
+  public <E extends Enum<E>> StructArrayEncoder enm(String name, E value) {
+    EnmField<E> field = (EnmField<E>) fieldSearcher.findField(name, EnmField.class, null);
+    currentData.add(new EnmDataHolder<E>(value, field.index(), field.getEnm()));
```
chrisdennis: If data holder construction was encoded inside the field type then all the duplicated code in these methods could be removed (and most of the other methods in the other encoders).

runnel/src/main/java/org/terracotta/runnel/encoding/StructEncoder.java
```diff
@@ -0,0 +1,217 @@
...
+
+    return bb;
+  }
+
+  /**
+   * Size the structure and return a newly allocated byte buffer containing the encoded structure of the exact size.
+   * The returned byte buffer is heap-allocated, so it supports {@link ByteBuffer#array()}.
+   * @return the encoded structure in a new byte buffer.
+   */
+  public ByteBuffer encode() {
```
chrisdennis: This method should call the other encode method, right?

runnel/src/main/java/org/terracotta/runnel/metadata/Metadata.java
chrisdennis: It's not megamorphic any more, so we can eliminate this class.

runnel/src/main/java/org/terracotta/runnel/utils/ReadBuffer.java
chrisdennis: You accusing the lusers of being wise?

runnel/src/main/java/org/terracotta/runnel/StructBuilder.java
lorban: true

runnel/src/main/java/org/terracotta/runnel/StructBuilder.java
lorban: true

runnel/src/main/java/org/terracotta/runnel/decoding/fields/AbstractValueField.java
lorban: True, let me do that.

runnel/src/main/java/org/terracotta/runnel/encoding/StructArrayEncoder.java
lorban: That's probably not a big deal, so I'm tempted to leave everything the way it is, at least for now.

runnel/src/main/java/org/terracotta/runnel/Enm.java
lorban: True, let me do that.

runnel/src/main/java/org/terracotta/runnel/EnmBuilder.java
lorban: True, let me do that.

lorban:
I vehemently disagree with the removal of the enum mapping feature. While *some* enums will be API, I'm pretty certain others won't. And even if all enums were API, we're not safe from human mistakes.

Writing the enum mappings is a one-off task, is only executed once during init at runtime, and only costs one EnumMap lookup per enum encoding and one HashMap lookup per decoding.

It is my opinion that this is a small price to pay for the peace of mind that clients and servers will never, ever disagree on the enum <-> integer mappings without you being able to notice it with a quick glance at the code.

runnel/src/main/java/org/terracotta/runnel/encoding/StructEncoder.java
lorban: Not directly, but I could extract the common code. Let me do that.

runnel/src/main/java/org/terracotta/runnel/StructBuilder.java
ljacomet: I know this has been merged and I am late to the game, but would be great to rename this to `addName` as suggested.

Issue #149, opened 2016-09-14
High-level Voltron Monitoring Service


mathieucarbou:
- [x] Compat' with Terracotta-OSS/terracotta-apis#151
- [x] Get rid of all usage of `PlatformMonitoringConstants`
- [x] Get rid of all `consuemerId` usage
- [x] Get rid of the tree structure: we only need to expose and push messages. We do not need to provide an api for a tree structure.
- ~~[ ] Remove the `CLIENT_TAGS_UPDATED` notification~~
- [x] RingBuffer: add indexes to entries to support `LOST_NOTIFICATIONS` directly in monitoring buffers
- ~~[ ] Remove the newly added module: `management-registry-service`. Since the `expose()` methods will be in the monitoring service directly. So no need for an extra service that is based on the `consumerId`.~~
- [x] More high-level monitoring service: requires using the `cluster-topology`, `management-model` and `sequence-generator` modules. Sort of methods to add:
 - `Stripe getStripe()`
 - `Collection<Client> getClients()`
 - `expose(ServerEntityRef, ContextContainer, Capability[])`
 - `expose(ClientDescriptor (?), ContextContainer, Capability[])`
 - etc...
- [x] support stats, notifications and management metadata from server entities

General goal is to move to a high-level service and move what Tms entity and management entities do into this service.

- Management Entity will just delegate its calls to the `expose()` methods, and handle management calls.
- Tms entity will then just reads all the buffers and get the topology.

If the interface we need cannot be extracted easily as a common API, then the entities can directly depend on an adapter that will forward all the interface calls to an `IMonitoringProducer` handled by the platform, and we will re-translate these calls to what we need within the service.

PR #152, opened 2016-09-16
Close #150: Compat' with Terracotta-OSS/terracotta-apis#151

### LINKED ISSUE #50: OffHeap Storage Reservation Service

### LINKED ISSUE #150: Compat' with Terracotta-OSS/terracotta-apis#151
- [x] Compat' with Terracotta-OSS/terracotta-apis#151

mathieucarbou:
:construction: Close #150: Compat' with Terracotta-OSS/terracotta-apis#151 (and Terracotta-OSS/tc-passthrough-testing#26)

Quick and dirty changes to be compatible with next tc-apis changes.

Goal is to minimize the work to do since all the monitoring service will be correctly refactored into #149 to provide a much richer API that consumers and M&M will use directly.
The new implementation will be decoupled from the voltron api and internal voltron things like the tree structure. See #149 for details.


anthonydahanne:
LGTM

PR #156, opened 2016-09-23
Runnel

lorban:
Add some lambda-friendly sub-structure helper methods and a check to prevent mistaken sub-structure encoding, plus support for bool and char types.

### DIFF
--- a/runnel/src/main/java/org/terracotta/runnel/encoding/StructEncoder.java
+++ b/runnel/src/main/java/org/terracotta/runnel/encoding/StructEncoder.java
@@ -103,6 +103,16 @@ public class StructEncoder implements PrimitiveEncodingSupport<StructEncoder> {
     return this;
   }
 
+  public StructEncoder struct(String name, StructEncoderFunction function) {
+    StructField field = fieldSearcher.findField(name, StructField.class, null);
+    List<DataHolder> values = new ArrayList<DataHolder>();
+    data.add(new StructDataHolder(values, field.index()));
+    StructEncoder subStructEncoder = new StructEncoder(field, values, this);
+    function.encode(subStructEncoder);
+    subStructEncoder.end();
+    return this;
+  }
+
   public StructEncoder struct(String name) {
     StructField field = fieldSearcher.findField(name, StructField.class, null);
     List<DataHolder> values = new ArrayList<DataHolder>();
@@ -171,6 +181,17 @@ public class StructEncoder implements PrimitiveEncodingSupport<StructEncoder> {
     data.add(new ArrayDataHolder(values, field.index()));
     return new StructArrayEncoder(values, this, ((StructField) field.subField()));
   }
+
+  public StructEncoder structs(String name, StructArrayEncoderFunction function) {
+    final ArrayField field = fieldSearcher.findField(name, ArrayField.class, StructField.class);
+    List<StructDataHolder> values = new ArrayList<StructDataHolder>();
+    data.add(new ArrayDataHolder(values, field.index()));
+    StructArrayEncoder subStructArrayEncoder = new StructArrayEncoder(values, this, ((StructField) field.subField()));
+    function.encode(subStructArrayEncoder);
+    subStructArrayEncoder.end();
+    return this;
+  }
+
 
   /**
    * Encode the structure in the passed byte buffer.
--- a/runnel/src/main/java/org/terracotta/runnel/encoding/StructEncoder.java
+++ b/runnel/src/main/java/org/terracotta/runnel/encoding/StructEncoder.java
@@ -199,6 +199,9 @@ public class StructEncoder implements PrimitiveEncodingSupport<StructEncoder> {
    * @return the passed-in byte buffer.
    */
   public ByteBuffer encode(ByteBuffer bb) {
+    if (parent != null) {
+      throw new IllegalStateException("Cannot encode non-root encoder");
+    }
     int size = calculateSize();
     return performEncoding(bb, size);
   }
@@ -209,6 +212,9 @@ public class StructEncoder implements PrimitiveEncodingSupport<StructEncoder> {
    * @return the encoded structure in a new byte buffer.
    */
   public ByteBuffer encode() {
+    if (parent != null) {
+      throw new IllegalStateException("Cannot encode non-root encoder");
+    }
     int size = calculateSize();
     ByteBuffer bb = ByteBuffer.allocate(size + VLQ.encodedSize(size));
     return performEncoding(bb, size);
--- a/runnel/src/main/java/org/terracotta/runnel/encoding/StructEncoder.java
+++ b/runnel/src/main/java/org/terracotta/runnel/encoding/StructEncoder.java
@@ -18,6 +18,7 @@ package org.terracotta.runnel.encoding;
 import org.terracotta.runnel.decoding.fields.ArrayField;
 import org.terracotta.runnel.decoding.fields.BoolField;
 import org.terracotta.runnel.decoding.fields.ByteBufferField;
+import org.terracotta.runnel.decoding.fields.CharField;
 import org.terracotta.runnel.decoding.fields.EnumField;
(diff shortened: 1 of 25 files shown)


runnel/src/main/java/org/terracotta/runnel/encoding/StructEncoder.java
```diff
@@ -103,6 +103,16 @@ public StructEncoder byteBuffer(String name, ByteBuffer value) {
     return this;
   }
 
+  public StructEncoder struct(String name, StructEncoderFunction function) {
+    StructField field = fieldSearcher.findField(name, StructField.class, null);
+    List<DataHolder> values = new ArrayList<DataHolder>();
```
ChrisGreenaway: Could be  new ArrayList<>()

runnel/src/main/java/org/terracotta/runnel/encoding/StructEncoder.java
```diff
@@ -172,12 +182,26 @@ public StructArrayEncoder structs(String name) {
     return new StructArrayEncoder(values, this, ((StructField) field.subField()));
   }
 
+  public StructEncoder structs(String name, StructArrayEncoderFunction function) {
+    final ArrayField field = fieldSearcher.findField(name, ArrayField.class, StructField.class);
+    List<StructDataHolder> values = new ArrayList<StructDataHolder>();
```
ChrisGreenaway: Could be new ArrayList<>(

runnel/src/test/java/org/terracotta/runnel/StructStructBuilderTest.java
```diff
@@ -66,6 +77,34 @@ public void testReadAll() throws Exception {
   }
 
   @Test
+  public void testReadAll_withLambda() throws Exception {
+    ByteBuffer bb = struct.encoder()
+        .string("name", "joe")
+        .struct("mapEntry", new StructEncoderFunction() {
```
ChrisGreenaway: Can you change the test to actually use a lambda - there should be no difference, but the test is called testReadAll_withLambda

runnel/src/test/java/org/terracotta/runnel/StructArrayStructBuilderTest.java
```diff
@@ -76,6 +78,42 @@ public void testReadAll() throws Exception {
   }
 
   @Test
+  public void testReadAll_withLambda() throws Exception {
+    ByteBuffer bb = struct.encoder()
+        .string("name", "joe")
+        .structs("mapEntry", new StructArrayEncoderFunction() {
```
ChrisGreenaway: Or pass a StructEncoderFunction and an iterable and the structs() method can iterate calling the function for each element?

runnel/src/test/java/org/terracotta/runnel/StructStructBuilderTest.java
```diff
@@ -66,6 +77,34 @@ public void testReadAll() throws Exception {
...
+          @Override
+          public void encode(StructEncoder encoder) {
```
ChrisGreenaway: Can you explain why this is better than the non-lambda way? Why would I not just call struct("mapEntry").string("key", "1").string("value", "one") [not sure I've the calls exactly right]?

runnel/src/main/java/org/terracotta/runnel/encoding/StructEncoder.java
lorban: No, that code targets JDK 6.

runnel/src/main/java/org/terracotta/runnel/encoding/StructEncoder.java
lorban: No, that code targets JDK 6.

runnel/src/test/java/org/terracotta/runnel/StructStructBuilderTest.java
lorban: No, that code targets JDK 6.

runnel/src/main/java/org/terracotta/runnel/encoding/StructEncoder.java
ChrisGreenaway: Yeah, forgot you were in a different JDK (x3).

runnel/src/test/java/org/terracotta/runnel/StructStructBuilderTest.java
lorban: well, when filling-in the entries of the struct takes more than a simple call to a getter (for instance a switch statement), it's handy to be able to delegate that to a lambda that does the work and keep the call chain fluent instead of having to break it up to store a temp variable.

ChrisGreenaway: Ludovic went through an example of this with me earlier. Makes sense.

runnel/src/test/java/org/terracotta/runnel/StructArrayStructBuilderTest.java
lorban: That would require a bit of work to factor-out an interface from `StructEncoder`. This isn't a bad idea, but more work than I'm willing to put in at the moment.

ChrisGreenaway: Isn't it just StructEncoderFunction rather than StructArrayEncoderFunction?

ChrisGreenaway: Ah, no, I suppose you have to have the element to encode in the method arguments too.

I think you should do this though.

runnel/src/main/java/org/terracotta/runnel/utils/ReadBuffer.java
```diff
@@ -37,6 +37,13 @@ private ReadBuffer(ByteBuffer byteBuffer, int limit) {
     }
   }
 
+  public Boolean getBoolean() {
+    if (byteBuffer.position() + 1 > limit) {
+      throw new LimitReachedException();
+    }
+    return byteBuffer.get() == 0 ? Boolean.FALSE : Boolean.TRUE;
```
ChrisGreenaway: Don't you need a cast to byte: "byteBuffer.get() == (byte) 0"?

ChrisGreenaway:
You can't encode every unicode character in two bytes.

runnel/src/main/java/org/terracotta/runnel/utils/ReadBuffer.java
lorban: Nah, javac is smart enough to figure out that `byteBuffer.get()` returns a byte, hence `0` must be cast as a byte.

But I can add the explicit byte cast.

runnel/src/test/java/org/terracotta/runnel/StructArrayStructBuilderTest.java
lorban: Nah, not now.

lorban:
@ChrisGreenaway I know that, but I don't how to call what java stores in a `char` variable. I've updated the doc to explicitly call that a 'java char'.

runnel/src/test/java/org/terracotta/runnel/StructArrayStructBuilderTest.java
ChrisGreenaway: If not now then you'll implement a load of lambdas using this API which would need to change.

ChrisGreenaway:
Just leave the char as two bytes, as discussed.

runnel/src/main/java/org/terracotta/runnel/utils/ReadBuffer.java
ChrisGreenaway: I don't mind either way - if you want to add an extra cast that's fine, or not.

It was just that spent so much of the last two weeks typing "(byte)" that the lack of it jumped out at me and I couldn't think straight on how the compiler would react.

PR #158, opened 2016-10-05
Decoder reuse when sharing common message base

ljacomet:
This adds a method to transform a decoder into a different one while
 maintaining read position and reducing new objects creation.
 This is useful when messages share a common structure, such as
  beginning with an operation code for example.

### DIFF
--- a/runnel/src/test/java/org/terracotta/runnel/encoding/dataholders/StructTest.java
+++ b/runnel/src/test/java/org/terracotta/runnel/encoding/dataholders/StructTest.java
@@ -16,14 +16,19 @@
 package org.terracotta.runnel.encoding.dataholders;
 
 import org.junit.Test;
+import org.terracotta.runnel.Struct;
+import org.terracotta.runnel.StructBuilder;
+import org.terracotta.runnel.decoding.StructDecoder;
 import org.terracotta.runnel.utils.ReadBuffer;
 import org.terracotta.runnel.utils.WriteBuffer;
 
 import java.nio.ByteBuffer;
 import java.util.Arrays;
 
+import static org.hamcrest.CoreMatchers.containsString;
 import static org.hamcrest.core.Is.is;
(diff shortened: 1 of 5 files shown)


runnel/src/test/java/org/terracotta/runnel/encoding/dataholders/StructTest.java
```diff
@@ -77,4 +82,65 @@ public void testWithoutIndex() throws Exception {
     assertThat(readBuffer.getString(8), is("deuxieme"));
   }
 
+  @Test
+  public void testFromDecoder() throws Exception {
+    Struct messageBase = StructBuilder.newStructBuilder().int32("opcode", 10).build();
+    Struct messageCacheCreate = StructBuilder.newStructBuilder().int32("opcode", 10).string("cache", 20).build();
+    Struct messageGet = StructBuilder.newStructBuilder().int32("opcode", 10).int64("key", 20).build();
```
lorban: I don't like this pattern of having to repeat the opcode in every sub-structure.

ChrisGreenaway: Me neither.

ljacomet: understood that the repetition is not nice. But can't reuse builders, they are changing internal state. And like I said - that's the minimal PR to not have to encapsulate these different messages in a top level struct. Which would allow to have flat binary structure and provide better API in the future while staying binary compatible - important for Ehcache adoption.

ChrisGreenaway: You could use two encoders. One for the type a second one (selected from N) for the message.

ljacomet:
Shelving, will use a flat binary format and will resolve the fact that it means creating multiple decoders at a later stage.

PR #159, opened 2016-10-05
Close Issue #149 - High-level Voltron Monitoring Service, usable from Server Entities

### LINKED ISSUE #149: High-level Voltron Monitoring Service

### LINKED ISSUE #153: ManagementAgentService: log issues to warn on management call execution in threads
ManagementAgentService: log issues to warn on management call execution in threads

`onMessage` line 86

mathieucarbou:
Close #149 and  Close #153


mathieucarbou:
test this please

management/cluster-topology/src/main/java/org/terracotta/management/model/cluster/ServerEntityIdentifier.java
```diff
@@ -13,58 +13,61 @@
...
-public abstract class ManagementEvent implements Serializable {
+public class ServerEntityIdentifier implements Serializable {
 
-  private static final long serialVersionUID = 1;
+  private static final long serialVersionUID = 1L;
 
-  private final String id;
-  private final ClientIdentifier from;
+  private final String type; // type (service, client entity, server entity, etc)
+  private final String name; // type (entity type name)
```
GaryWKeim: Check comment.

management/management-entity/management-entity-client/src/main/java/org/terracotta/management/entity/management/client/ManagementAgentService.java
```diff
@@ -81,36 +85,44 @@ public void unregister(Object managedObject) {
...
+                      .build()
+                      .execute()
+                      .getSingleResult();
+                  // check again in case the management call takes some time
+                  if (bridging) {
+                    get(entity.callReturn(null, message.getFrom(), message.getManagementCallIdentifier(), aReturn), timeout);
+                  }
                 }
+              } catch (RuntimeException err) {
+                LOGGER.log(Level.WARNING, "Error on management call execution or result sending for " + contextualCall + ". Error: " + err.getMessage(), err);
```
GaryWKeim: Please avoid, possibly unnecessary, string concatenation by using parameter substitution in all logging.

mathieucarbou: Not available for the JDK logger. I'll wrap!

mathieucarbou:
@ljacomet @anthonydahanne : the build does not pass because CloudBee is using an outdated JDK 8 that has a bug. 


Yesterday I ran into a generic issue with my JDK 8:

    Stream.of(Optional.of(1), Optional.of(2)).map(n -> n.orElseThrow(IllegalStateException::new));

Gives this compile error:

    Error:(205, 69) java: unreported exception X; must be caught or declared to be thrown

I needed to write this to have no compilation errors:

    Stream.of(Optional.of(1), Optional.of(2)).map(n -> n.<IllegalStateException>orElseThrow(IllegalStateException::new));

__java 1.8.0_51__

IDE was smarter than the JDK, saying I do not need to specify the type ;-) And @chrisdennis  tested on his computer (__1.8.0_102__) and this works fine. After updated mine, it works fine also. 

Anyway! All of that to say that we have an outdated JDK on CloudBee that also has this bug. I am not able to make the PR build pass:

https://terracotta-oss.ci.cloudbees.com/job/terracotta-platform-pull-requests/293/consoleFull

```
[ERROR] /scratch/jenkins/workspace/terracotta-platform-pull-requests/management/monitoring-service/src/main/java/org/terracotta/management/service/monitoring/DefaultMonitoringService.java:[200,100] error: unreported exception X; must be caught or declared to be thrown
[ERROR] where X,T are type-variables:
[ERROR] X extends Throwable declared in method <X>orElseThrow(Supplier<? extends X>)
[ERROR] T extends Object declared in class Optional
[ERROR] /scratch/jenkins/workspace/terracotta-platform-pull-requests/management/monitoring-service/src/main/java/org/terracotta/management/service/monitoring/DefaultMonitoringService.java:[232,107] error: unreported exception X; must be caught or declared to be thrown
[ERROR] -> [Help 1]
```

management/cluster-topology/src/main/java/org/terracotta/management/model/cluster/Client.java
```diff
@@ -60,7 +61,12 @@ public Client setManagementRegistry(ManagementRegistry managementRegistry) {
     return this;
   }
 
-  public SortedSet<String> getTags() {
+
+  public boolean isManageable() {
```
GaryWKeim: I have questions about the thread-safety of this type.

mathieucarbou: Topology objects are not meant to be shared: no one is thread safe: the whole model is serialized on server entity and deserialized on consumer-side. 

I used some Concurrent maps because the simulator (__before__) had several threads modifying the topology, but this is no more the case. 

The right modification to do would be to remove all the concurrent maps I think, if we want to not have 2 different types of structures.

mathieucarbou: I added commit 43a3474 for that.

management/monitoring-service/src/main/java/org/terracotta/management/service/monitoring/DefaultMonitoringService.java
```diff
@@ -0,0 +1,317 @@
...
+  DefaultMonitoringService(DefaultStripeMonitoring stripeMonitoring, long consumerId, ClientCommunicator clientCommunicator) {
+    this.stripeMonitoring = stripeMonitoring;
+    this.consumerId = consumerId;
+    this.clientCommunicator = clientCommunicator;
+  }
+
+  @Override
+  public void close() {
+    fetches.clear();
+    buffer.clear();
```
GaryWKeim: Other places in this module handle buffer == null.

management/monitoring-service/src/main/java/org/terracotta/management/service/monitoring/DefaultStripeMonitoring.java
```diff
@@ -0,0 +1,381 @@
...
+        .setStartTime(platformServer.getStartTime())
+        .setHostAddress(platformServer.getHostAddress())
+        .setVersion(platformServer.getVersion())
+        .computeUpTime();
+
+    stripe.addServer(server);
+
+    entities.putIfAbsent(server.getServerName(), new HashMap<>());
+
+    fireNotification(SERVER_JOINED, server.getContext());
```
GaryWKeim: Isn't it a no-no to invoke event listeners under a lock?

mathieucarbou:
@GaryWKeim : i finished applying changes after your comments and added commit 08bba63 so that the build pass. There's a JDK issue, so I forced the type to overcome it.

management/monitoring-service/src/test/java/org/terracotta/management/service/monitoring/VoltronMonitoringServiceTest.java
```diff
@@ -57,136 +30,15 @@
...
-    assertThat(buffer.read(), equalTo(null));
-
-    producer.pushBestEffortsData(null, "test_push_buffer", 1);
-    producer.pushBestEffortsData(null, "test_push_buffer", 2);
-    producer.pushBestEffortsData(null, "test_push_buffer", 3);
-
-    assertThat(buffer.size(), equalTo(2));
-    assertThat(buffer.read(), equalTo(2));
-    assertThat(buffer.read(), equalTo(3));
+  public void test() {
```
anthonydahanne: maybe delete the test class altogether , since VoltronMonitoringService does not exist anymore

management/samples/hello-world-entity/src/test/java/org/terracotta/management/entity/helloworld/HelloWorldTest.java
```diff
@@ -22,29 +22,28 @@
...
-import org.terracotta.management.entity.management.client.ManagementAgentEntityClientService;
-import org.terracotta.management.entity.management.client.ManagementAgentEntityFactory;
-import org.terracotta.management.entity.management.client.ManagementAgentService;
 import org.terracotta.management.entity.helloworld.client.HelloWorldEntity;
 import org.terracotta.management.entity.helloworld.client.HelloWorldEntityClientService;
 import org.terracotta.management.entity.helloworld.client.HelloWorldEntityFactory;
 import org.terracotta.management.entity.helloworld.client.management.HelloWorldManagementRegistry;
 import org.terracotta.management.entity.helloworld.server.HelloWorldEntityServerService;
-import org.terracotta.management.entity.monitoring.client.MonitoringServiceEntity;
+import org.terracotta.management.entity.management.ManagementAgentConfig;
```
mathieucarbou: I would prefer not to correct this since we cannot rebase master to get rid of this change... so now that it's done, it's done, and we should avoid adding another reformat commit.

management/management-registry/src/main/java/org/terracotta/management/registry/AbstractManagementProvider.java
```diff
@@ -21,10 +21,10 @@
 import org.terracotta.management.model.capabilities.context.CapabilityContext;
 import org.terracotta.management.model.capabilities.descriptors.Descriptor;
 import org.terracotta.management.model.context.Context;
+import org.terracotta.management.model.stats.Statistic;
 import org.terracotta.management.registry.action.ExposedObject;
 import org.terracotta.management.registry.action.Named;
 import org.terracotta.management.registry.action.RequiredContext;
-import org.terracotta.management.model.stats.Statistic;
```
mathieucarbou: I would prefer not to correct this since we cannot rebase master to get rid of this change... so now that it's done, it's done, and we should avoid adding another reformat commit.

management/monitoring-service/src/test/java/org/terracotta/management/service/monitoring/VoltronMonitoringServiceTest.java
mathieucarbou: No: I wanted to test the monitoring service. I wanted to complete and add some tests in this class :-)
This is ongoing ;-)

PR #160, opened 2016-10-12
Fix to dependency scopes

ljacomet:
Move all voltron libraries to provided


mathieucarbou:
@ljacomet, @anthonydahanne : I propose we release tomorrow with also #161 included (yet to be merged)

PR #166, opened 2016-10-24
Workaround for Terracotta-OSS/terracotta-core#334

mathieucarbou:
Workaround for Terracotta-OSS/terracotta-core#334
Required for ehcache/ehcache3#1555


anthonydahanne:
LGTM

myronkscott:
You understand that while this change does not cause NPE, it also means that the value will not be stored.  Correct?

myronkscott:
Disregard my last comment.  This change should work.

PR #169, opened 2016-10-26
upgrade to latest dependencies

anthonydahanne:


anthonydahanne:
test this please
weird failure

mathieucarbou:
test this please

mathieucarbou:
@jd0-sag @myronkscott : the voltron-proxy module does not compile anymore. 

See failure at the end: https://terracotta-oss.ci.cloudbees.com/job/terracotta-platform-pull-requests/337/console

This is because this module is jdk6 and compiled with a jdk6 on cloudbee. I guess the client communicator api is now compiled jdk8 ?

@anthonydahanne : was able to reproduce locally: doing a `mvn clean install` with jdk8 works, but fails with a jdk6. So we probably need to change the Cloudbee default jdk ?

jd0-sag:
APIs `1.0.9.beta` changed some other server-side modules to Java8, including `standard-cluster-services`.

This module normally shouldn't be seen by client-side code (where the Java6 requirement exists).  Proxy seems to mix client and server code in the same module, which is probably the issue.

anthonydahanne:
test this please

mathieucarbou:
Replaced by #176

PR #175, opened 2016-10-28
split voltron proxy into modules

mathieucarbou:
Fix required for #169


voltron-proxy/voltron-proxy-client/pom.xml
```diff
@@ -0,0 +1,50 @@
...
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <parent>
+    <artifactId>voltron-proxy</artifactId>
+    <groupId>org.terracotta.voltron.proxy</groupId>
+    <version>5.0-SNAPSHOT</version>
+  </parent>
+  <modelVersion>4.0.0</modelVersion>
+
+  <artifactId>voltron-proxy-client</artifactId>
+
```
anthonydahanne: isn't it required to set the maven compiler plugin and define the java.version to 1.6 ?
same thing for common

mathieucarbou: its defined as 1.6 by default at upper level ;-)

anthonydahanne: ok, thanks !

mathieucarbou: at least... it is... now :P

anthonydahanne:
LGTM

PR #182, opened 2016-11-12
Close #178 and close #180: management calls on server entities

### LINKED ISSUE #178: [mnm] Support management calls from voltron registry service
`ConsumerManagementRegistryProvider` can provide a `SharedManagemenetRegistry` service a little like Ehcache standalone, that would support management calls and redirect them to the right `ConsumerManagementRegistry` provided to a specific server entity.

`SharedManagemenetRegistry` would know about all the provided `ConsumerManagementRegistryProvider`.

From the tms entity (or management entity), it would then be possible to request the `SharedManagemenetRegistry` to execute a management call on any exposed object indirectly from this service.

### LINKED ISSUE #180: When client disconnect, clear ongoing management call list

mathieucarbou:
See #178 and #180.

I merged the monitoring service and the registry service in the same module to avoid the complexity of passing a service registry to another service at entity creation time. This also allows a direct access to the consumerId, which we need to create contexts for exposed objects in the management providers.

Also extracted ehcache collector service that is targeted by the management call to put it at a common place, since it is used in both clustered and standalone in Ehcache.

The collector service also support collecting stats over a shared management registry. It means that we can configure one collector, and it is able to look into all the statistic providers that will be exposed (pools and stores at the moment)


management/monitoring-service-api/README.adoc
```diff
@@ -0,0 +1 @@
+API of the monitoring service. Entities can depend on this module at compile time. And this module can be put on serevr classpath
```
anthonydahanne: typo : "server"

management/management-entity/README.adoc
```diff
@@ -0,0 +1,3 @@
+The management entity is used client-side to bridge a client's management registry (with some management metadata about some client-side objects) to the server monitoring service.
+
+This entity enables a client to send its client-side statistics and notification to the server, and also receive and respond to management calls.
```
anthonydahanne: typo "statistics and notificationS"

mathieucarbou:
@anthonydahanne : rebased

management/pom.xml
```diff
@@ -42,7 +42,6 @@
     <module>management-registry-service</module>
     <module>management-entity</module>
     <module>tms-entity</module>
-    <module>samples</module>
```
anthonydahanne: well, yes, we're not supposed to display such entities in MnM
But still, it's a nice example you provided, not sure why we _do_ need to get rid of it (I know it'll still be in the history, but this is not really convenient and won't adapt to future api changes)

mathieucarbou: I disagree. 

The goal if this sample was to show hot to bridge the management entity with a custom user entity on client-side so that a management registry on a client can expose management metadata and receive management calls.

This was entity was developed when we needed a way in TMS to show some entities other than Ehcache for testing purposes.

This sample is not meant to show how to integrate the M&M stack because it is incomplete: it does not depend on the right libraries, it does not show how to package, it does not show all the server-entity part for management registry and management calls.

Also, if one day management is made "official", the API that the users will use is not this API but the API that is currently closed-source inside TMS. This is this API that will need to be documented and put OSS.

We need to keep (and improve) this sample ONLY if the goal is to make the management stack OSS AND to provide users a way to develop their own entities and manage them. 
As far as I know, this is not the case: only Ehcache clustered is supported and the voltron server has no standalone kit or documentation to officially show to develop custom entities.

So to my opinion, this sample, __that we do not need anymore in M&M__ and that is uncomplete anyway, should be removed until we know what to do with management OSS.

mathieucarbou:
@anthonydahanne : updated PR, removed commit that removed the hello world test entity

anthonydahanne:
LGTM

PR #194, opened 2016-11-23
lazily initialize metadata to be able to create struct 'aliases'

lorban:
This feature is needed to be able to configure structs of structs with loops, e.g.:
```
struct a:
 struct b 10

struct b:
 struct c 10

struct c:
 struct a 10
```

### DIFF
--- a/runnel/src/main/java/org/terracotta/runnel/metadata/Metadata.java
+++ b/runnel/src/main/java/org/terracotta/runnel/metadata/Metadata.java
@@ -27,12 +27,19 @@ import java.util.Map;
  */
 public class Metadata {
 
-  private final Map<String, Field> fieldsByName;
+  private final List<? extends Field> metadata;
+  private Map<String, Field> fieldsByName;
 
   public Metadata(List<? extends Field> metadata) {
-    fieldsByName = new HashMap<String, Field>();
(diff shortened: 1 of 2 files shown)


runnel/src/main/java/org/terracotta/runnel/metadata/Metadata.java
```diff
@@ -27,12 +27,18 @@
...
 
-  private final Map<String, Field> fieldsByName;
+  private final Map<String, Field> fieldsByName = new HashMap<String, Field>();
+  private final List<? extends Field> metadata;
 
   public Metadata(List<? extends Field> metadata) {
-    fieldsByName = new HashMap<String, Field>();
-    for (Field field : metadata) {
-      fieldsByName.put(field.name(), field);
+    this.metadata = metadata;
```
henri-tremblay: Here I would initialize the hashmap size based on the fields in metadata

henri-tremblay:
Apart from my comment, LGTM

runnel/src/main/java/org/terracotta/runnel/metadata/Metadata.java
lorban: Sure, I can do that.

Issue #196, opened 2016-11-23
Update to the latest terracotta Config / API / Core versions


ljacomet:
This includes updating the offheap-resource to the new configuration system in place of being a `ServiceProvider`

PR #198, opened 2016-11-23
Issue #196 Transform off heap resource in ExtendedConfiguration

ljacomet:
* It is now an ExtendedConfigParser instead of a ServiceProvider
* Bump jaxb plugin version to fix build lifecycle issues

This should not be merged right now as passthrough does not have yet support for `ExtendedConfig` and thus a release of tc-platform with the update contained here will not be usable by Ehcache 3.

However, still interested in comments (from @myronkscott amongst others) on the usage of that new feature.


mathieucarbou:
@ljacomet : I was wondering... If we do some modif inside this module, could it be also splitted into 2 modules: offheap-resource-api and -impl ?

The reason is that server entities (like tms) cannot depend on just an api right now (which would allow to have the impl being optional on the  server classpath, like we do for m&m). 

Currently it forces us to always have this offheap jar in the classpath for our tests.

Can this split be done in this PR (since this pr will lead to another release and integration within ehcache) or I should open another issue for that ?

offheap-resource/src/main/java/org/terracotta/offheapresource/OffHeapResourcesProvider.java
```diff
@@ -44,81 +40,56 @@
...
+        LOGGER.warn("More offheap configured than there is physical memory [{} > {}]", totalSize, physicalMemory);
       }
     } else {
-      return false;
+      throw new IllegalStateException("Resources already initialized");
     }
   }
 
   @Override
-  public <T> T getService(long consumerID, ServiceConfiguration<T> unknownConfiguration) {
```
mathieucarbou: If it is no more a service provider, how the `TmsAgentImpl` line 83 will work now to access `OffHeapResources ` to expose the settings and stats of server resources ?

myronkscott: Can you link it in via the initialize method of consumerManagerRegistryService?

https://github.com/Terracotta-OSS/terracotta-apis/blob/master/entity-server-api/src/main/java/org/terracotta/entity/PlatformConfiguration.java#L25

myronkscott:
this looks inline with ExtendedConfigParser change.

offheap-resource/src/main/java/org/terracotta/offheapresource/OffHeapResourcesProvider.java
mathieucarbou: I'll see. I think so.. It means that now, the monitoring service can directly access all "extended services" ?

The problem is that if we  do so, it will create a hard dependency between the monitoring service and the off heap service.

@ljacomet : what do you think ?

ljacomet: I do not see the different between the link that existed with the `ServiceProvider` based impl. Dependency relation did not change IMO. Am I missing something?

ljacomet:
@mathieucarbou I am puzzled by the need to decouple api and impl. What will it bring us? When you look at what is being built on top of the Terracotta platform, this offheap-resource is mandatory always.

mathieucarbou:
@ljacomet yes it is mandatory, but is it normal that M&M depends on all the implementations of things it wants to monitor ?
I.e. if we introduce otehr services on day, and some are built-ins, then having m&m depends on them will bring these deps on the classpaths and they will be loaded, whereas this is not what we would want perhaps.

Having an api separated than the impl brings the choice to use or not a service, even if m&m code depends on it (service provider will just return null).

Its like you want, but i am raising a problem that we will have.

offheap-resource/src/main/java/org/terracotta/offheapresource/OffHeapResourcesProvider.java
mathieucarbou: @ljacomet  : yes you need to make sure `TmsAgentImpl ` can access the offheap api otherwise clustered integration tests in ehcache will fail when you will test this change because we ensure the offheap descriptors and stats are there. They are required for M&M.

ljacomet:
@mathieucarbou Let's say that now this is extended config and no longer service provider, it is by design a much smaller component and thus I do not find the need to split so important.

mathieucarbou:
@ljacomet : do I have the rights to complete this PR to add the commit for the management part ? Or I'll make another PR after ?

ljacomet:
@mathieucarbou you have the rights

ljacomet:
This is now complete and good to merge

mathieucarbou:
@anthonydahanne : PR is good for me and @ljacomet and tested in ehcache so you can review / merge __and release__ :-)

management/monitoring-service/src/main/java/org/terracotta/management/service/monitoring/PlatformConsumerManagementRegistry.java
```diff
@@ -0,0 +1,69 @@
...
+    super(consumerId, monitoringService, statisticsService);
+    this.platformConfiguration = platformConfiguration;
+    this.statisticConfiguration = statisticConfiguration;
+  }
+
+  @Override
+  public void init() {
+    // manage offheap service if it is there
+    Collection<OffHeapResources> offHeapResources = platformConfiguration.getExtendedConfiguration(OffHeapResources.class);
+    if (!offHeapResources.isEmpty()) {
```
anthonydahanne: isn't offheap supposed to be mandatory now ?

ljacomet: Well, Ehcache requires it, but it is not _mandatory_

mathieucarbou: only mandatory to be in the server classpath, but the list of plugins will be empty if the user does not configure it in tc-config file

PR #203, opened 2016-11-25
[closes #202] add basic documentation, including expectable thread-safety characteristics

### LINKED ISSUE #202: [runnel] Document thread-safety characteristics
`Encoder`s and `Decoder`s are not thread-safe and need not be, but `Struct`s have to be as it is desirable for a `Struct` to be shared by many threads requesting encoders and decoders in parallel.

Such characteristics should be explicitly documented.

lorban:
Just some javadoc changes.

### DIFF
--- a/runnel/src/main/java/org/terracotta/runnel/decoding/StructDecoder.java
+++ b/runnel/src/main/java/org/terracotta/runnel/decoding/StructDecoder.java
@@ -30,7 +30,8 @@ import org.terracotta.runnel.utils.ReadBuffer;
 import java.nio.ByteBuffer;
 
 /**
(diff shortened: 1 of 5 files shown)


runnel/src/main/java/org/terracotta/runnel/decoding/StructDecoder.java
```diff
@@ -30,7 +30,8 @@
 import java.nio.ByteBuffer;
 
 /**
- * @author Ludovic Orban
+ * An decoder allows decoding structured data described by a {@link org.terracotta.runnel.Struct}.
```
mathieucarbou: "A decoder..."

PR #210, opened 2016-11-29
Close #208 (updateCollectedStatistics) and Close #209 (redundant ClientIdentifier)

mathieucarbou:
This PR is rebased on #205, which must be merged before this one.

The only relevant commit of this PR is fc0bfa65ef90e0fda1c7fffa6d22fa8804920806. The other ones are part of #205.

PR #205, opened 2016-11-25
Close #192: improving testing of management module, fixed some issues

mathieucarbou:
@anthonydahanne : since this PR is big (but splitted into several atomic ones), I opened it up before it is finished so that you can start looking at it.

The only thing remaining is to complete the tests methods that currently fail on purpose.

This PR is to have a solid testing module and solid tests in tc-platform, based on passthrough first.

There's another issue logged to add integration tests by using the new sample entity inside this PR.

- I fixed 2 issues (see :bug: commits)
- I had to refactored a little some generic methods and the list containing the capabilities and descriptors so that we can have reproducible tests more easily.

PR #211, opened 2016-11-29
Close #193 Fixing Management Entity, TMS Entity and Monitoring Service implementation and responsability, especially around management call usage

### LINKED ISSUE #193: [mnm] Fixing Management Entity, TMS Entity and Monitoring Service implementation and responsability, especially around management call usage
`ManagementAgent` that __only__ allows exposing registry and client tags (subset of the current management agent)

- `TmsAgent` that will allows to trigger management calls, read cluster messages and topology and receives management call results

- A ManagementCall object (like in TMS) that represents the call that we are waiting for (sort of completable future). We have that in TMS, needs to bring it in the client API.

This works is to fix:

- security issues: any client can currently do a mangement call to any other client
- have the same method call for both server and client management calls
- be async all the way (client, active server, passive servers)

Also, split the big classes in monitoring service in smaller services:

- ManagemetService: used for tms entity
- ConsumerMangementRegistry
- PlatformMangementRegistry
- SharedMangementRegistry
- ClientMonitoringService
- ActiveEntityMonitoringService
- PassiveEntityMonitoringService

This split will allow to implement management call correctly, because management call implementation depends whether the target is a client, an active entity or a passive entity.

mathieucarbou:
This PR is rebased on #210, which must be merged before this one.

The only relevant commit of this PR is the last one. The other ones are part of #205 and #210.


mathieucarbou:
test this please

mathieucarbou:
test this please

PR #212, opened 2016-11-30
Update to latest terracotta versions

ljacomet:


ljacomet:
I know some of this is already in #205 but it would be good to have a version of tc-platform aligned on latest which contains just the off heap resource changes.

mathieucarbou:
yes agree! our PRs must wait...

PR #216, opened 2016-12-06
[closes #201] rework aliasing feature to fix many initialization bugs

### LINKED ISSUE #201: [runnel] Make it possible to have loops in struct definitions
Such loops in `Struct` definitions should be possible:
```
Struct A:
 struct B 10

Struct B:
 struct C 10

Struct C:
 struct A 10
```

lorban:


lorban:
The aliasing code is now much more robust and more lean.

I've also checked that the Ehcache 3 test suite passes to make sure I didn't break anything.

runnel/src/main/java/org/terracotta/runnel/metadata/Metadata.java
```diff
@@ -16,48 +16,70 @@
...
+  public void checkFullyInitialized() throws IllegalStateException {
+    if (fullyInitialized) {
+      return; // already checked, no need to redo the work
+    }
+
+    if (!initialized) {
+      throw new IllegalStateException("Metadata not yet initialized");
+    }
+
+    fullyInitialized = true; // cache the fact that the check has been performed, has to be done early
```
chrisdennis: Scares me that this is both volatile and used in this manner.  This isn't threadsafe at the moment since a second checking thread could erroneously validate as fully initialized if it sees the trueness set by the other thread.

lorban: You're right, but a synchronized block with double checked locking should solve that, right?

chrisdennis: Thread 1: sets `fullyInitialized` to true
Thread 2: skips checking
Thread 1: throws Exception due to problem
Thread 1: sets `fullyInitialized` to false
Thread 2: claims success
Thread 1: claims failure

chrisdennis:
Test this please.

Issue #217, opened 2016-12-06
[mnm] Change Statistic Architecture


mathieucarbou:
__Currently, statistic architecture works like this:__

1. Ehcache has some __counters__ it increments
2. Ehcache clients needs to start a __scheduling thread__ to compute each 1 sec. __sampling__ statistics based on these counters. It fills a history buffer of about 100 entries for each computed stat. This computes automatically also rates, ratio, latencies based on the window we defined.
3. Stats query
  a) __For unclustered__, the stat query api queries against this __history buffer__ and returns all computed stats since a given time. So the only concept we have as an output are `HistoryStatistics`
  b) __For clustered__, a __collector service__ runs at periodic intervals (each 22 sec), based on the TTD, to avoid stats being disabled and collect each new computed samples, and push them into the voltron monitoring service. TMS collects them each 5-10 seconds.
4. TMS display stats: that are currently badly calculated

__Targeted new architecture:__ _The goal is to get rid of the scheduling layer and stats history computed on client-side, to only output counters._

1. Ehcache has some counters it increments
2. No scheduling: Ehcache management API only allows to query raw counters. No more buffer history, latencies and rates and ratio. The `StatisticRegistry` class will probably be deprecated or changed a lot. Also `HistoryStatistic` would be deprecated or left unused in favor of primitive statistics and raw counters
3. Stats query
  a) For unclustered, the stat query api queries would be changed: it will only output counter values.
  b) For clustered, a collector service runs at periodic intervals __that is configured and no more based on the TTD__ (i.e. each 5 or 10 seconds). This will be the minimal window usable by TMS. But by default, window is 1 min so it's OK. Collector will collect all stats, timestamp them and put the counters into voltron monitoring service. TMS collects them each 5-10 seconds.
4. TMS will then only store the counters in DB. When stats are requested, TMS will load counters from the DB and compute rates, ratio based on the window configured by the user (or hard value) default to 1min.

This work needs to follow an Ehcache PR first that will setup the right counters to use: https://github.com/ehcache/ehcache3/issues/1684

mathieucarbou:
@ljacomet @gibsong @chrisdennis @GaryWKeim  : can you please review the description based on what has been discussed yesterday ? This is not urgent work but I logged it to remember it. Thanks!

@anthonydahanne @lorban : FYI.

mathieucarbou:
__UPDATE__

See also https://github.com/ehcache/ehcache3/issues/1684 which is more updated.

PR #219, opened 2016-12-06
Do not expose latencies anymore

mathieucarbou:
Note: this does not prevent latencies to be computed. 
See https://github.com/ehcache/ehcache3/issues/1684 and #217 for that.


gibsong:
Is there any reason why you commented out the latency code rather than delete it?

mathieucarbou:
Yes! because for compounds, each type (average, min, max, etc) is related to a type in the stat library. So this is to remember to get rid of that also in the stat library later. 
I'll add a note for that

PR #223, opened 2016-12-08
[closes #222] add API to encode/decode struct inside structs

### LINKED ISSUE #222: [runnel] no API to encode/decode struct inside structs
Consider the following structs:

```
  private final Struct valueStruct = StructBuilder.newStructBuilder()
      .int32("aaa", 1)
      .build();

  private final Struct mapEntryStruct = StructBuilder.newStructBuilder()
      .string("key", 1)
      .struct("value", 2, valueStruct)
      .build();

  private final Struct myStruct = StructBuilder.newStructBuilder()
      .structs("mapEntry", 1, mapEntryStruct)
      .build();
```

`myStruct` contains an array of `mapEntryStruct` struct that itself contains a `valueStruct` struct.

It should be possible to do the following:

```
      struct.encoder()
        .structs("mapEntry")
          .string("key", "1")
          .struct("value") // not in the API
            .int32("aaa", 1)
          .end()
        .next()
          .string("key", "2")
          .struct("value") // not in the API
            .int32("aaa", 1)
          .end()
        .end()
        .encode();
```

Unfortunately, the API lacks the `struct()` methods that would allow encoding the sub-structure. The same is true for decoding as well: there is no API for decoding a struct from an array of structs.

lorban:
Beware:  this contains changes that might break the API of struct-decoding lambda helpers.

### DIFF
--- a/runnel/src/main/java/org/terracotta/runnel/Struct.java
+++ b/runnel/src/main/java/org/terracotta/runnel/Struct.java
@@ -45,8 +45,8 @@ public class Struct {
    * Note: this method is thread-safe.
    * @return the encoder.
    */
-  public StructEncoder encoder() {
-    return new StructEncoder(root);
+  public StructEncoder<Void> encoder() {
+    return new StructEncoder<Void>(root);
   }
 
   /**
@@ -55,8 +55,8 @@ public class Struct {
    * @param byteBuffer the byte buffer containing the data to be decoded.
    * @return the decoder.
    */
-  public StructDecoder decoder(ByteBuffer byteBuffer) {
-    return new StructDecoder(root, new ReadBuffer(byteBuffer));
+  public StructDecoder<Void> decoder(ByteBuffer byteBuffer) {
+    return new StructDecoder<Void>(root, new ReadBuffer(byteBuffer));
   }
 
   /**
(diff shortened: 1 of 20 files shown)


AbfrmBlr:
`StructArrayEncoder.string` throws a null pointer for null values

AbfrmBlr:
java.lang.NullPointerException
	at org.terracotta.runnel.encoding.dataholders.StringDataHolder.encodeString(StringDataHolder.java:48)
	at org.terracotta.runnel.encoding.dataholders.StringDataHolder.<init>(StringDataHolder.java:33)
	at org.terracotta.runnel.encoding.StructArrayEncoder.string(StructArrayEncoder.java:106)
	at org.ehcache.clustered.common.internal.messages.ExceptionCodec.encode(ExceptionCodec.java:65)

lorban:
@AbfrmBlr : this works as designed: runnel does not encode null values so there are some checks preventing them. What you're supposed to do is add a null check yourself and only encode the string when you have a non-null value.

On the decoding side, runnel doesn't care if there was an encoded value or not: if you decode a field that wasn't encoded, you'll get null back.

But now that makes me wonder if runnel should allow encoding null values and simply ignore them, as well as using wrapper types instead of primitives, that might make the API nicer. Could you open a bug for this? I'll track it and do it if we deem it's worth it. Thanks!

AbfrmBlr:
I have filed an issue #226 and marked it as a question.

PR #225, opened 2016-12-08
[closes #224] cleanup/fix offheap-resource build

### LINKED ISSUE #224: [offheap-resource] cleanup the way sources are generated by xjc
The `xjc` task isn't bound to the `generate-sources`, and the `jaxb2-maven-plugin` does not add the xjc-generated sources to the source path.

This makes the build non-reproducible and doesn't help IDE's understanding that some sources are generated.

lorban:


mathieucarbou:
agree! I reviewed and I'm ok with that. 
@lorban did you test from scratch with these modifications ? Because I already tried to bind to `generate-sources` to the jaxb plugin already, but when loading within IntelliJ, it was not triggered... I didn't add the build helper though, because nothing was generated.

lorban:
@mathieucarbou I've tried building platform with this change and everything seems to be fine, and intellij can now see the generated sources and include them in its classpath.

Nevertheless, I'd like @ljacomet and @chrisdennis to have a look at this change as I'd like to make sure the lifecycle hooks I added don't break anything I may have overlooked.

mathieucarbou:
@lorban : even without the build helper plugin, I already got this behavior. What I was saying is that if you first do a `mvn clean install`, and then load the project in Intellij. Like from the beginning. IntelliJ should detect that some plugins needs source generation and he should trigger them.

This is what was not working for me. But seeing the generated source folder mounted in IDEA, after a `mvn clean install`, this was already working, even without the build helper.

lorban:
@mathieucarbou No, that doesn't work, you still have to manually do `mvn process-resources` despite the fact that intellij is supposed to do it itself when the project is being refreshed.

And indeed, the build-helper-maven-plugin doesn't seem to be required anymore so I'll remove it.

lorban:
Apparently, common/data-root-config also uses jaxb but with a different plugin. We probably want to align that.

ljacomet:
Good for me - needs a rebase

PR #229, opened 2016-12-09
Close #162: Passive entities for management

### LINKED ISSUE #162: [mnm] A/P support
- [ ] create passive entities
- [ ] add passthrough tests for passive servers
- [ ] implement `EntityServerService.getExecutionStrategy()` in M&M entity services
- [ ] fix concurrency strategy that was set to non concurrent
- [ ] verify methods calls on entities in the case they are replicated on both active & passive to be sure they apply accordingly (i.e. `TmsAgentImpl.call()`)

mathieucarbou:
- Add passive entities for M&M
- Add concurrency strategy support in voltron-proxy
- Add execution strategy support in voltron-proxy
- Add passive support in voltron-proxy
- Add sync support in voltron-proxy
- Add passive entities for test sample
- Added more tests


mathieucarbou:
__Note: this PR builds fine with this fix: https://github.com/Terracotta-OSS/tc-passthrough-testing/pull/67__

anthonydahanne:
and of course, waiting on the  new pass through release , so that tests pass on cloudbees

voltron-proxy/voltron-proxy-server/src/main/java/org/terracotta/voltron/proxy/server/ActiveProxiedServerEntity.java
```diff
@@ -77,7 +84,27 @@ public void destroy() {
...
   }
 
-  protected void fireAndForgetMessage(Object message, ClientDescriptor ... clients) {
-    target.fireAndForgetMessage(message, clients);
+  protected void synchronizeKeyToPassive(int concurrencyKey) {
+  }
+
+  protected <M> void fireMessage(Class<M> type, M message, boolean echo) {invoker.fireMessage(type, message, echo);}
+
+  protected <M> void fireMessage(Class<M> type, M message, ClientDescriptor[] clients) {invoker.fireMessage(type, message, clients);}
```
mathieucarbou: we have to leave them. it's public api. it allows an entity to fire message to specific clients or all clients. we are using one of the methods in the sample. feature was already there before.

PR #233, opened 2016-12-12
Close #227: preparing sample use cases for a demo and to show how to use the clustered management layer

### LINKED ISSUE #227: Add a little M&M sample
that could be used as a startup doc

mathieucarbou:
FYI @henri-tremblay : you were asking how to use the clustered management layer... Here are some samples.

### DIFF
--- a/management/testing/doc/README.adoc
+++ b/management/testing/doc/README.adoc
@@ -1,0 +1,25 @@
+1. Download Ehcache Clustered Kit:
+
+ http://repo1.maven.org/maven2/org/ehcache/ehcache-clustered/3.2.0/ehcache-clustered-3.2.0-kit.zip
+
+2. Unzip
+
+3. Put management-server.jar inside `server/plugins/lib`
+
+4. Put a file named `tc-config.xml` inside `server` folder with this content:
+
+  <?xml version="1.0" encoding="UTF-8"?>
+  <tc-config xmlns="http://www.terracotta.org/config"
+             xmlns:ohr="http://www.terracotta.org/config/offheap-resource">
+    <plugins>
+      <config>
+        <ohr:offheap-resources>
+          <ohr:resource name="primary" unit="MB">64</ohr:resource>
+        </ohr:offheap-resources>
+      </config>
+    </plugins>
+  </tc-config>
+
+5. Go to `server` and run: `./bin/start-tc-server.sh -f tc-config.xml`
+
+6. Run any of the examples. For some, you might need to have a cache manager initialized first.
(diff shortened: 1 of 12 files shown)


mathieucarbou:
UPDATE: @anthonydahanne @GaryWKeim : this PR is done, complete and tested with a real server. So you can merge.

Also, it depends on PR #235 : so either you directly merge this (it will close #235 also), or you do it separately. It's the same.

management/testing/doc/src/main/java/org/terracotta/management/doc/ReadStatistics.java
```diff
@@ -0,0 +1,139 @@
...
+        .filter(c -> c.getName().startsWith("Ehcache:"))
+        .forEach(ehcache -> {
+
+          // 2. create a routing context
+          Context ctx = ehcache.getContext()
+              .with("cacheManagerName", "my-super-cache-manager");
+
+          try {
+            // 3. collect stats on client-side
+            tmsAgentService.updateCollectedStatistics(ctx, "StatisticsCapability", asList("Cache:HitCount", "Clustered:HitCount", "Cache:MissCount", "Clustered:MissCount")).waitForReturn();
```
anthonydahanne: hum, you assume here that an ehcache client is connected and running.
that's fine, but that's not part of the readme though.

management/testing/doc/README.adoc
```diff
@@ -0,0 +1,25 @@
...
+    <plugins>
+      <config>
+        <ohr:offheap-resources>
+          <ohr:resource name="primary" unit="MB">64</ohr:resource>
+        </ohr:offheap-resources>
+      </config>
+    </plugins>
+  </tc-config>
+
+5. Go to `server/bin` and run: `./start-tc-server.sh -f tc-config.xml`
```
mathieucarbou: I do not want to have server files or ehcache files in the tc-platform project... It is juts for demo so if we start putting some files, I fear we will continue and create a dependency over other projects... :-S

management/testing/doc/src/main/java/org/terracotta/management/doc/ReadStatistics.java
mathieucarbou: no I do not assume that and I tested them without an ehcache client, they work. BUT if some ehcache clients are there, I will also collect their stats.

README: `For some, you might need to have a cache manager initialized first.`

mathieucarbou:
@anthonydahanne : as I already said (emails and scrums) this is just a standalone example to show to use the management api to query remotely management information of a cluster because:
1. we had some people internally asking for how the m&m stack can be used
2. we had now something to demo to show how it works.
It is not intended for blog post or other external use since M&M is not official.

Issue #191, opened 2016-11-22
[mnm] Add integration tests with Galvan


mathieucarbou:
- [ ] Fix: https://github.com/Terracotta-OSS/galvan/issues/99
- [ ] Split the sample into 2
- [ ] integrate Galvan (see current other maven project with integration test and failsafe)
- [ ] do some stats test, test contexts
- [ ] do some notif tests
- [ ] do some A/P tests
- [ ] do some failover tests
- [ ] add also unit tests marked as Ignored that cannot be done with Galvan

PR #236, opened 2016-12-14
Close #171: Failover support on server side

### LINKED ISSUE #171: [mnm] Failover support on server-side
When a failover happens, on the passive server that became active, the platform sends to the monitoring service all the "old" monitoring structure before constructing the new one. 

So we need to differentiate between calls that are related to replaying the old monitoring tree, and calls that relates to the new one.

I.e. platforms is calling `addNode` to add entities with a sender that is active and entities that are passive (`PlatformServer.isActive == false`), and eventually other data after that. When we receive such events, we know that the platform is replaying some old data.

Platform then sends the SAME entities objects with `isActive == true`. This is when we know that the platform stopped replaying data and will start receiving new data.

- [x] Also, sends an event to the TMS saying that monitoring data is replayed, stopped replaying so that it can knows there has been a failover on this server.
- [x] add passthrough tests for failover

mathieucarbou:
Makes the monitoring service supports failover and detect when entities are transitioning from passive state to active state.
The transition is recorded, and when the new monitoring tree is built and new active entities are created, if this entity was passive before, then  a `SERVER_ENTITY_FAILOVER_COMPLETED` event is sent with the entity context.

It is possible that the TMS does not get these notifications if the tms entity does not reconnect BEFORE the other ones.

__WARNING__

With passthrough, platform does not send the right events to us after a failover, when client reconnects and when new active entities are created. So there's a lot of notification that I do not have and I cannot assert them because the monitoring tree is not reconstructed as it should be.

I placed some `TODO` in the project for the tests that needs to be also done ASAP with Galvan (https://github.com/Terracotta-OSS/terracotta-platform/issues/191)


mathieucarbou:
__DEPENDS ON https://github.com/Terracotta-OSS/tc-passthrough-testing/pull/71__

PR #239, opened 2016-12-16
Initial Galvan testing set up for tc platform

anthonydahanne:
@mathieucarbou here is the initial setup; let's discuss tomorrow about it !


management/testing/integration-tests/pom.xml
```diff
@@ -42,6 +42,125 @@
...
+    </dependency>
+
+    <!-- provided by voltron (client) -->
+    <dependency>
+      <groupId>org.terracotta</groupId>
+      <artifactId>entity-client-api</artifactId>
+      <scope>provided</scope>
+    </dependency>
+
+    <dependency>
```
mathieucarbou: We should only keep the client classpath (sample entity, client-api, management-client, etc) in the pom, because these are the only one dependencies we use from junit tests

Server dependencies that are just here so that maven can download them, and they have to be put in the kit plugins/lib folder, should be in a dependency plugin, but not in the dependency section (i.e. management-server, offheap, sample entity).

sample entity is the only exception: it is at both places.

management/testing/integration-tests/pom.xml
```diff
@@ -42,6 +42,125 @@
...
+              <goal>unpack</goal>
+            </goals>
+            <configuration>
+              <artifactItems>
+                <artifactItem>
+                  <groupId>org.terracotta.internal</groupId>
+                  <artifactId>terracotta-kit</artifactId>
+                  <version>${terracotta-core.version}</version>
+                  <type>tar.gz</type>
+                </artifactItem>
```
mathieucarbou: add other artifact items for server classpath (sample entity, offheap, management-common, management-server)

management/testing/sample-entity/src/main/resources/META-INF/services/org.terracotta.entity.EntityServerService
```diff
@@ -1 +1 @@
-sample.server.HelloWorldEntityServerService
\ No newline at end of file
+org.terracotta.management.entity.sample.server.CacheEntityServerService
```
mathieucarbou: good catch! 👍

management/testing/integration-tests/pom.xml
anthonydahanne: but I did ! 
have you looked below ?

management/testing/integration-tests/pom.xml
anthonydahanne: ok

management/testing/integration-tests/pom.xml
anthonydahanne: fixed

PR #246, opened 2016-12-20
Add support for entities to log to server files

myronkscott:
Entities can log to the server files by including this module then adding 

```
  static {
    LoggingSupplement.logNamespaceToServer("org.ehcache");
  }
```

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -49,6 +49,7 @@
     <module>communicator-support</module>
     <module>healthchecker-entity</module>
     <module>runnel</module>
+    <module>server-logging</module>
     <module>management</module>
   </modules>
 
--- a/server-logging/pom.xml
+++ b/server-logging/pom.xml
@@ -1,0 +1,49 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+
+    Copyright Terracotta, Inc.
+
+    Licensed under the Apache License, Version 2.0 (the "License");
+    you may not use this file except in compliance with the License.
+    You may obtain a copy of the License at
+
+        http://www.apache.org/licenses/LICENSE-2.0
+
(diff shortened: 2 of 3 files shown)


mathieucarbou:
@myronkscott : what is the difference between that and providing a server logging config ? Also, what is the default logging level ?Thanks!

CC @anthonydahanne - do we need this ?

pom.xml
```diff
@@ -50,6 +50,7 @@
     <module>healthchecker-entity</module>
     <module>runnel</module>
     <module>management</module>
+    <module>server-logging</module>
```
mathieucarbou: It would be better to put this module before `<module>management</module>` because management does not depend on it, so build order is undefined. And since management takes a lot of time to build and test, it would be better for you and other people if "faster" modules are built before management ;-)

ljacomet:
@myronkscott If I understand correctly, this is the _hack_ to get entity logging in the server file for the x.1 line?
We need to add this component to the kit and make sure to invoke the code you listed, for example in an `EntityServerService`.
Correct?

ljacomet:
@mathieucarbou This comes from an Ehcache request and the difference with adding a config is that it all goes to the `terracotta.log` file and not another one. Makes sense?

mathieucarbou:
Ok @ljacomet. So this is not something M&M should use right ? Or should we ?

myronkscott:
@ljacomet you are correct on all points.

ljacomet:
@myronkscott Ok, great!
@mathieucarbou You probably want to use that unless your code already logs in the right categories and thus does not need it.

ljacomet:
@mathieucarbou @anthonydahanne I would like to have this in a .1.1-pre5 release, can we chat and make it happen?

PR #247, opened 2016-12-20
#191: Support of new upcoming voltron changes for failover

### LINKED ISSUE #191: [mnm] Add integration tests with Galvan

mathieucarbou:
These changes support re-using some monitoring services, or cleaning before reusing. This is required to be compat' with the new voltron changes in master.

See: https://github.com/Terracotta-OSS/terracotta-core/issues/409

With these changes, failover can work, BUT, we still do see the replay phase (isActive=false) and thus we cannot detect that there has been a failover.

Also, with current tc-core and galvan master, Failover tests will block when terminating the active. O opened a ticket here: https://github.com/Terracotta-OSS/terracotta-core/issues/410.
Thus, tests times out, and we see no exception.

### DIFF
--- a/management/monitoring-service/src/main/java/org/terracotta/management/service/monitoring/DefaultClientMonitoringService.java
+++ b/management/monitoring-service/src/main/java/org/terracotta/management/service/monitoring/DefaultClientMonitoringService.java
@@ -101,6 +101,8 @@ class DefaultClientMonitoringService implements ClientMonitoringService, Topolog
 
   @Override
   public void onBecomeActive() {
+    LOGGER.trace("[{}] onBecomeActive()", this.consumerId);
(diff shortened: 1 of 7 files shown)


mathieucarbou:
Note: currently blocked by https://github.com/Terracotta-OSS/terracotta-core/issues/409

PR #255, opened 2017-01-03
Add support for all struct features within elements of struct arrays

chrisdennis:


mathieucarbou:
@chrisdennis : I think you need to rebase ;-)

ljacomet:
test this please

@mathieucarbou Could you look into the test failure? seems unrelated but may be an indication that there is an unstable test.

PR #259, opened 2017-01-05
Close #217 - Removal of statistic registry, sampling layer, using collector instead

### LINKED ISSUE #217: [mnm] Change Statistic Architecture

mathieucarbou:
This is the work required to get rid of the sampling/scheduling layer that generate stats history and samples.

See issues: #217  and ehcache/ehcache3#1684

- statistic registry scheduling layer and history are removed
- statistic registry layer that finds observers into the context tree is kept
- collector now collect by default all observers based on the interval set by the TMS
- no more statistic configuration, sampling, windowing or time to disable
- monitoring service is also not providing anymore any scheduling for stats

In general this is a BIG cleanup / simplification of a layer that was not required.

@henri-tremblay : `StandardEhcacheStatistics` should contain 2 methods not:

- `Statistic<?, ?> queryStatistic(String fullStatisticName)` : used by the local query api if a user uses the management registry to query some specific stats
- `Map<String, Statistic<?, ?>> queryStatistics()` : used by the local query api, or the collector, when we want to collect all stats. The stat name is returned as a key.

__NEW STUFF__

I have created a class `StatisticRegistry` in `org.terracotta.management.registry.collect` which responsibility is to hold a map of stat names - observers and contains the methods:

- `public Statistic<?, ?> queryStatistic(String fullStatisticName)`
- `public Map<String, Statistic<?, ?>> queryStatistics()`
- `public Collection<StatisticDescriptor> getDescriptors()`
- Plus all the method of the old stats registry used to register stats
- `StatisticDescriptor` now does not have its own StatType: it just returns as a string the `StatisticType` used directly from the stat framework so that you have much more information about what is a stat:

```
public enum StatisticType {
  COUNTER,
  RATE,
  LATENCY_MIN,
  LATENCY_MAX,
  LATENCY_AVG,
  RATIO,
  SIZE
}
```

These methods used by the providers from clients and entities to query the stat observers behind, mainly passthrough and raw counters (operations).


mathieucarbou:
test this please

management/management-registry/src/main/java/org/terracotta/management/registry/collect/StatisticRegistry.java
```diff
@@ -0,0 +1,229 @@
...
+  private final Map<String, ValueStatistic<? extends Number>> statistics = new HashMap<String, ValueStatistic<? extends Number>>();
+  private final Map<String, StatisticType> statisticTypes = new HashMap<String, StatisticType>();
+
+  public StatisticRegistry(Object contextObject) {
+    this.contextObject = contextObject; // allows null, in this case it will be a no-op registry
+  }
+
+  public Collection<StatisticDescriptor> getDescriptors() {
+    Set<StatisticDescriptor> descriptors = new HashSet<StatisticDescriptor>();
+    for (Map.Entry<String, ValueStatistic<? extends Number>> entry : statistics.entrySet()) {
```
henri-tremblay: Please add a set size here

management/management-registry/src/main/java/org/terracotta/management/registry/collect/StatisticRegistry.java
```diff
@@ -0,0 +1,229 @@
...
+      case SIZE: return new Size(statistic.value().longValue(), MemoryUnit.B);
+      case LATENCY_MIN: return new Duration(statistic.value().longValue(), TimeUnit.MILLISECONDS);
+      case LATENCY_MAX: return new Duration(statistic.value().longValue(), TimeUnit.MILLISECONDS);
+      case LATENCY_AVG: return new Average(statistic.value().doubleValue(), TimeUnit.MILLISECONDS);
+      default: throw new UnsupportedOperationException(type.name());
+    }
+  }
+
+  public Map<String, Statistic<?, ?>> queryStatistics() {
+    Map<String, Statistic<?, ?>> stats = new TreeMap<String, Statistic<?, ?>>();
```
henri-tremblay: Why a TreeMap?

management/management-registry/src/main/java/org/terracotta/management/registry/collect/StatisticRegistry.java
```diff
@@ -0,0 +1,229 @@
...
+  private <N extends Number> void registerStatistic(String statNameSuffix, StatisticType type, final ValueStatisticDescriptor descriptor) {
+    if (contextObject == null) {
+      return;
+    }
+
+    Set<TreeNode> result = queryBuilder()
+        .descendants()
+        .filter(context(attributes(Matchers.<Map<String, Object>>allOf(
+            hasAttribute("name", descriptor.getObserverName()),
+            hasAttribute("tags", new Matcher<Set<String>>() {
```
henri-tremblay: We should mutualize a `hasTag` (probably in stats)

management/management-registry/src/main/java/org/terracotta/management/registry/collect/StatisticRegistry.java
```diff
@@ -0,0 +1,229 @@
...
+    if (contextObject == null) {
+      return;
+    }
+
+    Set<TreeNode> result = queryBuilder()
+        .descendants()
+        .filter(context(attributes(Matchers.<Map<String, Object>>allOf(
+            hasAttribute("type", descriptor.getType()),
+            hasAttribute("name", descriptor.getObserverName()),
+            hasAttribute("tags", new Matcher<Set<String>>() {
```
henri-tremblay: Mutualize with the other matcher on tags

management/monitoring-service-api/src/main/java/org/terracotta/management/service/monitoring/registry/provider/AbstractExposedStatistics.java
```diff
@@ -16,51 +16,45 @@
...
-  private final StatisticsRegistryMetadata statisticsRegistryMetadata;
+  private final StatisticRegistry statisticRegistry;
 
-  protected AbstractExposedStatistics(Context context, T binding, StatisticsRegistry statisticsRegistry) {
+  protected AbstractExposedStatistics(Context context, T binding, Object contextObject) {
     super(context, binding);
-    // allows it to be null
-    this.statisticsRegistry = statisticsRegistry;
-    this.statisticsRegistryMetadata = new StatisticsRegistryMetadata(statisticsRegistry);
+    this.statisticRegistry = new StatisticRegistry(contextObject);
```
henri-tremblay: Inject the `StatisticRegistry`?

management/monitoring-service-api/src/main/java/org/terracotta/management/service/monitoring/registry/provider/AbstractStatisticsManagementProvider.java
```diff
@@ -53,40 +45,26 @@ protected void dispose(ExposedObject<T> exposedObject) {
...
-          statistics.put(statisticName, exposedObject.queryStatistic(statisticName, since));
-        } catch (IllegalArgumentException ignored) {
-          // ignore when statisticName does not exist and throws an exception
+      if (statisticNames == null || statisticNames.isEmpty()) {
+        statistics.putAll(exposedObject.queryStatistics());
+      } else {
+        for (String statisticName : statisticNames) {
+          try {
+            statistics.put(statisticName, exposedObject.queryStatistic(statisticName));
+          } catch (IllegalArgumentException ignored) {
```
henri-tremblay: Can we modify to prevent having a business flow handling exceptions?

management/monitoring-service/src/test/java/org/terracotta/management/service/monitoring/ManagementRegistryServiceTest.java
```diff
@@ -116,8 +116,7 @@ public void test_management_info_pushed() {
     registry.register(new MyObject("myCacheManagerName2", "myCacheName2"));
     registry.refresh();
 
-    assertThat(buffer.size(), equalTo(1));
-    assertThat(buffer.read().unwrap(ContextualNotification.class).get(0).getType(), equalTo("ENTITY_REGISTRY_UPDATED"));
+    assertThat(buffer.size(), equalTo(0));
```
henri-tremblay: Can use `assertThat(buffer, empty());`

management/monitoring-service/src/test/java/org/terracotta/management/service/monitoring/VoltronMonitoringServiceTest.java
```diff
@@ -338,8 +338,7 @@ public void pushBestEffortsData(String name, Serializable data) {
         new DefaultCapability("capabilityName", new CapabilityContext(), new CallDescriptor("myMethod", "java.lang.String")));
 
     messages = messages();
-    assertThat(messages.size(), equalTo(1));
-    assertThat(notificationTypes(messages), equalTo(Arrays.asList("ENTITY_REGISTRY_UPDATED")));
+    assertThat(messages.size(), equalTo(0));
```
henri-tremblay: Can use assertThat(buffer, empty());

management/management-registry/src/main/java/org/terracotta/management/registry/collect/StatisticRegistry.java
mathieucarbou: oups no need there!

management/management-registry/src/main/java/org/terracotta/management/registry/collect/StatisticRegistry.java
mathieucarbou: And move it here: https://github.com/Terracotta-OSS/statistics/issues/26

management/management-registry/src/main/java/org/terracotta/management/registry/collect/StatisticRegistry.java
mathieucarbou: also: https://github.com/Terracotta-OSS/statistics/issues/26

management/monitoring-service/src/test/java/org/terracotta/management/service/monitoring/VoltronMonitoringServiceTest.java
mathieucarbou: i do not have hamcrest-all in pom :-(

PR #264, opened 2017-01-13
Close #261: Removal of statistic wrapper classes

### LINKED ISSUE #261: Discard all stats wrappers because they are already known through descriptors
Will allow to get rid of all stat objects and primitives and just send back a map of stat name and values.
Will also allow to be less static and accept different types of stats in the future without the need of updating management libs

mathieucarbou:


anthonydahanne:
I agree with that simplification : since the caching clients will only report counts, sizes and eventually in the future, durations, I guess this is fine.
Why Number and not Long ?

mathieucarbou:
@anthonydahanne : because if we force to numbers, it will prevent the use of stat that could be decimals. Number is opened, even if for now we only use long (count and size)

PR #268, opened 2017-01-24
🎨 Close #269 : Open TmsAgentService

### LINKED ISSUE #269: Open TmsAgentService
Turn it into an interface so that we can implement different flavours of it

anthonydahanne:


mathieucarbou:
I'll check in Ehcache if this PR needs some change in Ehcache and if yes I'll add a commit to https://github.com/ehcache/ehcache3/pull/1795

PR #272, opened 2017-01-28
Support for Entity.reconfigure() and Added config option in TmsAgent Entity to set a stripe name

mathieucarbou:
Several additions:

__1. Set a stripe name to Tms Agent__

You can now do:

```
TmsAgentEntity tmsAgentEntity = tmsAgentEntityFactory.retrieveOrCreate(new TmsAgentConfig()
        .setMaximumUnreadMessages(1024 * 1024)
        .setStripeName("MY SUPER STRIPE"));
```

This will return a cluster object with the right stripe name :-) And simplify your code... You know where ;-)

__2. I also upgraded to -pre10__

__3. Support for reconfigure()__

@jd0-sag and @myronkscott did a change in tc-core to remove the `removeNode(entity)` call. When they will have done the fix, we will be able to remove the following code in `IStripeMonitoringPlatformListenerAdapter`:

```
        // ignore removeNode() if fetches are in progress on active server
        if (sender.getServerName().equals(currentActive)) {
          for (PlatformClientFetchedEntity fetch : fetches.values()) {
            if (fetch.entityIdentifier.equals(name)) {
              return false;
            }
          }
        }
```

This code is compatible for now and in the future so it is safe to commit for now so that you are unlocked and we will be able to upgrade to new platform versions transparently.

UPDATE: @anthonydahanne : done in commit 78f6976c2dd3f47f869a1e4cfeff1cc40e375dda

### DIFF
--- a/management/cluster-topology/src/main/java/org/terracotta/management/model/cluster/Stripe.java
+++ b/management/cluster-topology/src/main/java/org/terracotta/management/model/cluster/Stripe.java
@@ -144,7 +144,7 @@ public final class Stripe extends AbstractNode<Cluster> {
   public void remove() {
     Cluster parent = getParent();
     if (parent != null) {
-      parent.removeClient(getId());
+      parent.removeStripe(getId());
     }
   }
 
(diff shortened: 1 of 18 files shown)


anthonydahanne:
LGTM

PR #274, opened 2017-01-30
Close #273: Filtering out statistics that are unsupported and returned -1

### LINKED ISSUE #273: Filter our stats if the value is -1
-1 is when a stat is not supported or could not be computed (i.e. cache closing).

mathieucarbou:
It's done globally in the stat query system.

This PR takes as hypothesis that normal stats / counters that are sent will never be negative values of course. Because -1 is filtered out.

### DIFF
--- a/management/management-registry/src/main/java/org/terracotta/management/registry/DefaultStatisticQuery.java
+++ b/management/management-registry/src/main/java/org/terracotta/management/registry/DefaultStatisticQuery.java
@@ -71,7 +71,12 @@ public class DefaultStatisticQuery implements StatisticQuery {
       Map<String, Number> statistics = new HashMap<String, Number>();
       for (ManagementProvider<?> managementProvider : managementProviders) {
         if (managementProvider.supports(context)) {
-          statistics.putAll(managementProvider.collectStatistics(context, statisticNames));
+          for (Map.Entry<String, Number> entry : managementProvider.collectStatistics(context, statisticNames).entrySet()) {
+            if (entry.getValue().doubleValue() >= 0) {
+              statistics.put(entry.getKey(), entry.getValue());
+            }
+          }
+
         }
       }
       contextualStatistics.put(context, new ContextualStatistics(capabilityName, context, statistics));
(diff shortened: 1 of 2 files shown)


management/testing/integration-tests/src/test/java/org/terracotta/management/integration/tests/ReconfigureEntityIT.java
```diff
@@ -104,11 +104,16 @@ public void notification_after_reconfigure() throws Exception {
...
+    List<Message> messages;
+    List<ContextualNotification> notifs;
+    do {
+      messages = tmsAgentService.readMessages();
+      notifs = messages.stream()
+          .filter(message -> message.getType().equals("NOTIFICATION"))
+          .flatMap(message -> message.unwrap(ContextualNotification.class).stream())
+          .collect(Collectors.toList());
+    } while (!Thread.currentThread().isInterrupted()
+        && notifs.stream().noneMatch(contextualNotification -> contextualNotification.getType().equals("SERVER_ENTITY_RECONFIGURED")));
```
mathieucarbou: this test was unstable, We need to wait for the notification to be generated.

management/management-registry/src/main/java/org/terracotta/management/registry/DefaultStatisticQuery.java
```diff
@@ -71,7 +71,12 @@ public String getCapabilityName() {
       Map<String, Number> statistics = new HashMap<String, Number>();
       for (ManagementProvider<?> managementProvider : managementProviders) {
         if (managementProvider.supports(context)) {
-          statistics.putAll(managementProvider.collectStatistics(context, statisticNames));
+          for (Map.Entry<String, Number> entry : managementProvider.collectStatistics(context, statisticNames).entrySet()) {
+            if (entry.getValue().doubleValue() != -1.0d) {
```
ljacomet: This seems tricky. If we are talking counters only here, why not check `< 0`?

mathieucarbou:
@ljacomet @GaryWKeim @anthonydahanne fixed PR.

management/testing/integration-tests/src/test/java/org/terracotta/management/integration/tests/ReconfigureEntityIT.java
anthonydahanne: not sure what's the relationship with the -1 stats, but.. okay :-)

PR #275, opened 2017-01-31
Bump to latest Terracotta releases - pre12

ljacomet:

PR #276, opened 2017-01-31
Fix ordering of notifications and fix registry.refresh()

mathieucarbou:
... that needs to be forced because there could be some async registrations

Replaces PR #275


mathieucarbou:
@ljacomet : I found my mistake. It comes from that:

```
CompletableFuture c = new CompletableFuture()
c.thenRun({ println "call 1" })
c.thenRun({ println "call 2" })
c.complete(null)
```

output is not:

```
call 1
call 2
```

but:

```
call 2
call 1
```

So what I saw in the logs, is that the right registry was put in the entity, and then another one, empty, that corresponds to the first delegated call.

I am fixing that now :-)

mathieucarbou:
@anthonydahanne @ljacomet : tests pass locally. should pass also here. PR ready to be reviewed and merged.

PR #291, opened 2017-02-21
Close #260: Add support for managing voltron plugins and services

### LINKED ISSUE #260: [mnm] MonitoringServiceProvider should not depend on any plugin
Currently, monitoring service provider depends on offheap plugin to add its management provider.

The monitoring service should not depend on them. Instead it should be able to discover them dynamically.

This issue is urgent to fix because other teams that needs to expose settings from other voltron services will have the same need and we cannot make the monitoring service depend on all services.

mathieucarbou:
(in a generic way so that the monitoring service do not have to depend on those)

### DIFF
--- a/management/monitoring-service-api/src/main/java/org/terracotta/management/service/monitoring/StatisticService.java
+++ b/management/monitoring-service-api/src/main/java/org/terracotta/management/service/monitoring/StatisticService.java
@@ -15,12 +15,14 @@
  */
 package org.terracotta.management.service.monitoring;
 
+import com.tc.classloader.CommonComponent;
 import org.terracotta.management.registry.collect.StatisticCollector;
 
 /**
  * This statistic service is injected into {@link org.terracotta.management.registry.ManagementProvider} objects implementing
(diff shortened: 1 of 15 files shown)


jd0-sag:
Note:  We merged another change to update the versions and fix some other bugs found in downstream testing so you can drop the version-bump commit when you rebase this.

management/monitoring-service-api/src/main/java/org/terracotta/management/service/monitoring/StatisticService.java
```diff
@@ -15,12 +15,14 @@
...
+import com.tc.classloader.CommonComponent;
 import org.terracotta.management.registry.collect.StatisticCollector;
 
 /**
  * This statistic service is injected into {@link org.terracotta.management.registry.ManagementProvider} objects implementing
  * {@link org.terracotta.management.service.monitoring.registry.provider.MonitoringServiceAware} when they are added into a {@link ConsumerManagementRegistry}
  */
-interface StatisticsService {
+@CommonComponent
+public interface StatisticService {
```
ljacomet: I believe that's on interface method declarations. But you can certainly make an interface package private ... and thus need to make it public explicitly at the class level.

PR #298, opened 2017-03-15
TDB-1742: Being able to change collector interval in a generic fashion from a manageable topology element + tc-core upgrade

mathieucarbou:


management/management-registry/src/main/java/org/terracotta/management/registry/collect/DefaultStatisticCollector.java
```diff
@@ -41,76 +41,94 @@
...
+    }
+
+    final long itv = TimeUnit.MILLISECONDS.convert(interval, unit);
+
+    // cancel the current task if it is scheduled with a different time
+    if (scheduled && intervalMs != itv) {
+      stopStatisticCollector();
+    }
+
+    if (!scheduled) {
```
anthonydahanne: what about when it's scheduled ?
Anyway, I tried to run tests locally and I have Timeout exceptions everywhere, I'll check with you !

management/testing/integration-tests/src/test/java/org/terracotta/management/integration/tests/CollectorReschedulingIT.java
```diff
@@ -0,0 +1,102 @@
...
+  public void can_reschedule_client_collectors() throws Exception {
+    triggerClientStatComputation(1, TimeUnit.SECONDS);
+
+    long start, end;
+
+    do {
+      start = System.currentTimeMillis();
+      readClientStatistics();
+      end = System.currentTimeMillis();
+    } while (end - start < 1000 && !Thread.currentThread().isInterrupted());
```
mathieucarbou: it is tested @anthonydahanne by the fact that the test will timeout if we do not receive anything. 
I don't think we can test that it does not take more than 10 seconds because the way it is scheduled, we only have the guarantee that the task won't be executed before a specific delay. But if the system is too busy or slow, the delay when the task is executed can increase.

So we can test that the data does not come before a certain time (at least N seconds), but we cannot assert that the data arrive within a certain delay.

anthonydahanne: ok

management/management-registry/src/main/java/org/terracotta/management/registry/collect/DefaultStatisticCollector.java
mathieucarbou: if(scheduled && intervalMs == itv), then we do nothing because it is already scheduled with the right amount of time

mathieucarbou: Timeouts must be something else, like your `/etc/hosts`

anthonydahanne: thanks for the timeout issue, that solved it !

Now, the thing that I missed is that stopCollectorService re switches scheduled to false, this is why that works fine.

PR #302, opened 2017-04-06
Upgrade to latest terracotta-apis

ChrisGreenaway:


jd0-sag:
Ideally, the switch of the snapshot version would be a different commit from the update of the dependencies and API change but I am not sure if there is much concern for such things in this repo.

Otherwise, the change looks much like I would have expected.

mathieucarbou:
@ChrisGreenaway i am reviewing... Especially concerning the voltron proxy update ;-)

PR #317, opened 2017-05-11
Supporting all different status comming from voltron

mathieucarbou:
This PR:

1. Upgrade tc-core to get the fix for https://github.com/Terracotta-OSS/terracotta-core/issues/550
2. Improves the server state enums to correctly handle different state values coming from voltron from different places (monitoring service and diagnostic port).


mathieucarbou:
@ljacomet @myronkscott : could you please verify if the mapping is correct and if the state transition tha tI put in the javadoc is right ? This is how I would understand it but I am not sure it is right.
Thanks!

mathieucarbou:
test this please

mathieucarbou:
I am relaunching the tests because teh yall pass locally and the failure is not related to any tests assertion.

PR #325, opened 2017-05-25
Client message tracker service

albinsuresh:
Client message tracker service's interfaces for initial review. The relevant ones are:
1. ClientMessageTracker
2. MessageTracker
3. TrackerPolicy

The test package has some dummy entities generated to showcase the interaction between entities and this service.

### DIFF
--- a/client-message-tracker/pom.xml
+++ b/client-message-tracker/pom.xml
@@ -1,0 +1,59 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+
+    Copyright Terracotta, Inc.
+
+    Licensed under the Apache License, Version 2.0 (the "License");
+    you may not use this file except in compliance with the License.
+    You may obtain a copy of the License at
+
+        http://www.apache.org/licenses/LICENSE-2.0
+
+    Unless required by applicable law or agreed to in writing, software
+    distributed under the License is distributed on an "AS IS" BASIS,
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0"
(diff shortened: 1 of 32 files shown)


albinsuresh:
test this please

albinsuresh:
test this please

ljacomet:
Given the pattern of usage in the sample entity, and after discussing with @albinsuresh, why do we not have the following as the `MessageTracker` API:

```java
public interface MessageTracker<M extends EntityMessage, R extends EntityResponse> extends StateDumpable {
  R processMessage(InvokeContext context, M message, Function<M, R> processor);
}
```

albinsuresh:
The new abstraction built over client message tracker is called "Once and only once message handler". If anyone has a better name to suggest please do so. Else I'll go ahead and modify the poms, packages etc as well to reflect the same name.

client-message-tracker/src/main/java/org/terracotta/client/message/tracker/OOOPolicy.java
```diff
@@ -0,0 +1,32 @@
...
+ * limitations under the License.
+ */
+package org.terracotta.client.message.tracker;
+
+import org.terracotta.entity.EntityMessage;
+
+/**
+ * A {@link OOOPolicy} for an entity defines which one of its message types need to be tracked by a {@link ClientMessageTracker}
+ */
+public interface OOOPolicy {
```
ljacomet: Should be name with a more general name, such as `TrackingPolicy`

management/testing/integration-tests/pom.xml
```diff
@@ -67,13 +67,6 @@
       <artifactId>galvan-support</artifactId>
       <version>${terracotta-core.version}</version>
       <scope>test</scope>
-      <!-- TODO Remove once galvan version is updated - simply check mockito-all is not longer pulled -->
```
ljacomet: This cannot yet go, needs a later `terracotta-core` version

ljacomet:
Just need one more rebase

PR #328, opened 2017-05-31
Connection leasing

ChrisGreenaway:

PR #332, opened 2017-06-08
NmsAgent entity should be a @PermanentEntity

mathieucarbou:
@anthonydahanne @GaryWKeim : Nms Agent entity has no state and should always be there, whether the management implementation jar is there or not (management api is always there).

So this PR will prevent the bug Gary had to happen because platform guarantees the entity construction. Also, it will prevent a bug from happening, in the case where the client activates management, and the entity is fetched, but the management implementation code is not on teh server classpath. This is a use case not tested that was not working because the Nms Agent Entity was always expecting the monitoring services to be not null, whereas it could be if management jar is removed from server classpath (M&M is optional).

FYI @ChrisGreenaway : this is the PR I was talking to you. I'll release after it is merged.


management/nms-agent-entity/nms-agent-entity-server/src/main/java/org/terracotta/management/entity/nms/agent/server/NmsAgentEntityServerService.java
```diff
@@ -31,45 +30,45 @@
...
   @Override
   public long getVersion() {
-    return NmsAgentVersion.LATEST.version();
+    return 1;
   }
 
   @Override
   public boolean handlesEntityType(String typeName) {
-    return NmsAgentConfig.ENTITY_TYPE.equals(typeName);
+    return "org.terracotta.management.entity.nms.agent.client.NmsAgentEntity".equals(typeName);
```
anthonydahanne: that was nice having NmsAgentConfig.ENTITY_TYPE instead of a string, that probably will make refactoring harder...

mathieucarbou: yes but this thing is isolated in 1 class now and the constant string cannot be a constant of this class and also be accessible in the annotation. It's done like the other persistent entities... There's always a duplication I saw.

mathieucarbou:
test this please

mathieucarbou:
@anthonydahanne : I launched the tests again. Build was failing because of a Leasing-Thing test (https://terracotta-oss.ci.cloudbees.com/job/terracotta-platform-pull-requests/725/) probably introduced in PR https://github.com/Terracotta-OSS/terracotta-platform/pull/328

PR #334, opened 2017-06-08
TDB-1892 adds getClusterTierName() method to Cluster class

gibsong:
I added a helper function to the Cluster class.  The function is getClusterTierName(String clientId, String cacheAlias), which is used to extract a the clusterTierName from the topology by using the clientId and the cacheAlias.  This method is needed in the StatisticsService and the TmsClusterMessageListener.

### DIFF
--- a/management/cluster-topology/src/main/java/org/terracotta/management/model/cluster/Cluster.java
+++ b/management/cluster-topology/src/main/java/org/terracotta/management/model/cluster/Cluster.java
@@ -15,6 +15,7 @@
  */
 package org.terracotta.management.model.cluster;
 
+import org.terracotta.management.model.capabilities.descriptors.Descriptor;
 import org.terracotta.management.model.context.Context;
 import org.terracotta.management.model.context.Contextual;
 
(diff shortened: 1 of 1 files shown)


management/cluster-topology/src/main/java/org/terracotta/management/model/cluster/Cluster.java
```diff
@@ -186,6 +187,33 @@ public String getPath(Context context) {
     return stripeStream().flatMap(Stripe::serverStream);
   }
 
+  public String getClusterTierName(String clientId, String cacheAlias) {
```
mathieucarbou: @gibsong : the concept of cluster tier is a special concept of Ehcache. It is not related at all to this model, which only support a raw topology view of a cluster. This model doesn't have any interpretation of the data it holds.

Such computation must be done in the project where you are using the topology, not within the topology itself.

gibsong: Shoot ok I will put this method in another class in the terracotta-enterprise project.  I thought this was the best place to add it but I guess I was wrong.

mathieucarbou:
I will close this PR

gibsong:
ok thanks for closing

PR #337, opened 2017-06-13
Tweaks to lease pom.xml files

ChrisGreenaway:


ljacomet:
Needs a rebase then good to go

PR #338, opened 2017-06-13
Add time unit to lease length configuration

ChrisGreenaway:


lease/entity-server/src/main/resources/lease-service.xsd
```diff
@@ -17,28 +17,53 @@
...
+          </xs:documentation>
+        </xs:annotation>
+      </xs:element>
+    </xs:sequence>
+  </xs:complexType>
+
+  <xs:complexType name="lease-length-type">
+    <xs:simpleContent>
+      <xs:extension base="xs:positiveInteger">
+        <xs:attribute name="unit" type="lease:time-unit" use="required">
```
ljacomet: I would go with optional having a default, let's say `milliseconds`

lease/entity-server/src/main/resources/lease-service.xsd
```diff
@@ -17,28 +17,53 @@
...
       </xs:documentation>
     </xs:annotation>
-    <xs:complexType>
-      <xs:sequence>
-        <xs:element type="xs:positiveInteger" name="lease-length"  minOccurs="1" maxOccurs="1">
+  </xs:element>
+
+  <xs:complexType name="connection-leasing-type">
+    <xs:sequence>
+      <xs:element name="lease-length" type="lease:lease-length-type" minOccurs="1" maxOccurs="1">
```
ljacomet: Still find `lease-length` less than ideal. What about `lease-duration`?

lease/entity-server/src/main/resources/lease-service.xsd
lorban: I'd rather use `seconds` as a default.

ChrisGreenaway: I don't think there should be a default.

lease/entity-server/src/main/resources/lease-service.xsd
ChrisGreenaway: The word "length" really goes with "lease". If you were renting a house, the conversation would be how long a lease do you want, not what duration do you want on your lease.

ljacomet: Ok, respect the native speaker there 😉

lease/entity-server/src/main/resources/lease-service.xsd
ljacomet: Let's go with no default, which enables adding one later if needs be 😉

chrisdennis:
Am I reading this right that parsing a lease configuration logs two things at info, neither of which are the actual lease length?  Whats the reasoning here?

PR #339, opened 2017-06-15
Make LeaseServiceProvider a BuiltinService so that default leasing works

ChrisGreenaway:


lease/system-test/src/test/java/org/terracotta/lease/NoLeaseConfigurationTest.java
```diff
@@ -0,0 +1,61 @@
...
+import org.junit.Test;
+import org.terracotta.connection.Connection;
+import org.terracotta.connection.ConnectionFactory;
+import org.terracotta.passthrough.PassthroughClusterControl;
+import org.terracotta.passthrough.PassthroughTestHelpers;
+
+import java.net.URI;
+import java.util.Properties;
+
+public class NoLeaseConfigurationTest {
```
anthonydahanne: add a tiemout rule, in case the while loop never exits ?

lease/system-test/src/test/java/org/terracotta/lease/NoLeaseConfigurationTest.java
```diff
@@ -0,0 +1,61 @@
...
+  private static PassthroughClusterControl cluster;
+
+  @BeforeClass
+  public static void beforeClass() {
+    cluster = PassthroughTestHelpers.createActiveOnly("stripe",
+            server -> {
+              server.registerClientEntityService(new LeaseAcquirerClientService());
```
anthonydahanne: is it required to register 2 client services ?

lease/system-test/src/test/java/org/terracotta/lease/NoLeaseConfigurationTest.java
cschanck: Are you gunshy after the hanging cluster test?

ChrisGreenaway: That's fair.

PR #341, opened 2017-06-28
Lease expiry now kills the client side of the connection too

ChrisGreenaway:


lease/api/src/main/java/org/terracotta/lease/LeaseExpiryConnectionKillingThread.java
```diff
@@ -0,0 +1,60 @@
...
+    setName("LeaseExpiryConnectionKillingThread");
+    setDaemon(true);
+  }
+
+  public void run() {
+    while (true) {
+      try {
+        Lease lease = leaseMaintainer.getCurrentLease();
+
+        if (lease instanceof LeaseImpl) {
```
ljacomet: After discussing with @ChrisGreenaway, this condition should be inverted to use the other `Lease` type in the check.

ChrisGreenaway:
I rebased and modified that instanceof check as requested. I also fixed a race in LeaseExpiryConnectionKillingThreadTest: I'd factored some common code out to the before() method; each test has its own setup to do before the thread starts; but I'd moved thread.start() to the before() - doh! Fixed now.

PR #346, opened 2017-07-06
[DO NOT MERGE] Fix #345 - Add a rich object reporter for state collection

### LINKED ISSUE #345: Create a StateCollector add-on for rich reporting
There is a need for a higher level API for Object and primitive reporting in state collection.  Provide a convenient access point for pulling the jackson ObjectMapper reporting facilities.

myronkscott:
The value of this approach is we are using a well known object mapping API to provide JSON output.  The JSON output will be included in the string output of a full state collection of the server.  JSON is human readable for consumption by an administrator.  The output is also machine parsable in case the platform would like to reformat the data into some other output.

### DIFF
--- a/state-format/src/main/java/com/tc/state/ReportingServiceProvider.java
+++ b/state-format/src/main/java/com/tc/state/ReportingServiceProvider.java
@@ -1,0 +1,84 @@
+/*
+ * Copyright Terracotta, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.tc.state;
+
+import com.fasterxml.jackson.core.JsonProcessingException;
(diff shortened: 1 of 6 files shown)


myronkscott:
This should be installed in the apis subdirectory of the plugins directory.  This way it will be common to all entities and services that need it.

mathieucarbou:
@ljacomet @lorban : what do you think ? 
It means every service provider or entities or voltron plugin (i.e. offheap) will have to depend on this library (provided by server) if this is accepted, because of the many limitations of the state dump api.

myronkscott:
requires yet to be released core changes

state-format/src/main/java/com/tc/state/ReportingServiceProvider.java
```diff
@@ -0,0 +1,84 @@
...
+import org.terracotta.entity.ServiceProvider;
+import org.terracotta.entity.ServiceProviderCleanupException;
+import org.terracotta.entity.ServiceProviderConfiguration;
+
+
+/**
+ *
+ */
+@BuiltinService
+public class ReportingServiceProvider implements ServiceProvider {
```
mathieucarbou: what is the benefit / use case of this service provider, especially since it can only be used from entities ?

myronkscott: This service provider turns cluster state output into ison rather than the internal pretty print format currently used.

mathieucarbou: but how it work ? 
Each service provider, instead of implementing StateDumpable, could provider instead a PrettyPrinter service to print their state ?

myronkscott: ObjectStateReporter is for entity or service reporting sugar.

myronkscott:
not relevant any longer

PR #359, opened 2017-07-19
Fix deadlock: lease connection killing tried to wait on itself

ChrisGreenaway:


lease/api/src/main/java/org/terracotta/lease/LeaseExpiryConnectionKillingThread.java
```diff
@@ -49,6 +49,7 @@ public void run() {
             } catch (IllegalStateException e) {
               // Already closed.
             }
+            break;
```
ljacomet: a `return` might be clearer

lease/api/src/main/java/org/terracotta/lease/ThreadCleaningLeaseMaintainer.java
```diff
@@ -48,13 +43,9 @@ public void close() throws IOException {
...
     }
 
-    for (Thread thread : threads) {
-      try {
-        thread.join();
-      } catch (InterruptedException e) {
-        LOGGER.error("Interrupted while shutting down Thread: " + thread, e);
-      }
-    }
+    // Joining on these threads is not a great idea because
```
ljacomet: Not sure the comment will mean anything once the above code effectively disappears

PR #375, opened 2017-08-09
Prevent word tiering

henri-tremblay:
Using a volatile was allowing possible word tiering. Also, using `lastReconciledMessageId` throughout the `reconcile` method instead of a local variable wasn't providing a stable behaviour.

This is now fixed.

However, a data race is still present. I can't evaluate if it's harmful, not knowing the exact threading usage of this class.

Right now I think it is harmless.

### DIFF
--- a/client-message-tracker/src/main/java/org/terracotta/client/message/tracker/MessageTrackerImpl.java
+++ b/client-message-tracker/src/main/java/org/terracotta/client/message/tracker/MessageTrackerImpl.java
@@ -22,6 +22,7 @@ import org.terracotta.entity.StateDumpCollector;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
+import java.util.concurrent.atomic.AtomicLong;
 import java.util.stream.LongStream;
 
 public class MessageTrackerImpl<M extends EntityMessage, R extends EntityResponse> implements MessageTracker<M, R> {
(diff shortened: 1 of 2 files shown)


client-message-tracker/src/main/java/org/terracotta/client/message/tracker/MessageTrackerImpl.java
```diff
@@ -22,13 +22,14 @@
...
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
+import java.util.concurrent.atomic.AtomicLong;
 import java.util.stream.LongStream;
 
 public class MessageTrackerImpl<M extends EntityMessage, R extends EntityResponse> implements MessageTracker<M, R> {
 
   private final TrackerPolicy trackerPolicy;
   private final ConcurrentMap<Long, R> trackedResponses;
-  private volatile long lastReconciledMessageId = 0;
```
chrisdennis: I'm confused, volatile longs cannot be torn.

henri-tremblay: Yes... You are right of course... Brain got strangely confused. I've kept the local variable. And I am still unsure if "race" is harmful

chrisdennis: I think the race is "safe" assuming that duplicate calls to `trackedReponses.remove(...)` are harmless.

henri-tremblay: I wasn't thinking about that. More about one thread setting the `lastReconciledMessageId` to a number lower than the current value set meanwhile by another thread. But I guess in that cave indeed, the next time the removal will just stop from farther then needed.

albinsuresh:
`reconcile` will be called on every invocation on an entity(for tracked messages). And it could happen concurrently from multiple concurrency keys and hence multiple threads. If there are concurrent reconcile attempts on id 99 and 87, the end result should have everything under 99 reconciled/cleared.

henri-tremblay:
Ok. This will happen but in some cases, `lastReconciledMessageId` will be 87 even if there was reconciliation until 99. To be cleaner, I have added a CAS loop.

Issue #376, opened 2017-08-09
Stopping lease threads requires closing lease entities (which blocks connection close)


chrisdennis:
Currently there is no way to cleanly shutdown the lease system when the cluster if offline.  This is only because in order to shutdown the lease threads we have to close the lease entity which needs to talk to the cluster. Instead we should trigger lease thread termination on `didDisconnectUnexpectedly()`.

PR #377, opened 2017-08-09
Issue #376 : Shutdown lease maintenance threads when connections are disconnected

chrisdennis:


chrisdennis:
I've moved to the simpler approach (with less code changes).  So far I've gone with exposing the destroy method only on the `CleaningLicenseMaintainer` it could be moved to the interface but would have to throw UOE on the main implementation (ugly).  I also wanted to write a system test for the destroy behavior, but pass through cannot handle `connection.close()` when the cluster is down, which kind of defeats the object as I would like to know that the changes allow destroy to go ahead when the cluster is dead.  In order to system test properly we would have to add a system test against a real cluster.

lease/api/src/main/java/org/terracotta/lease/LeaseMaintainerFactory.java
```diff
@@ -33,16 +33,17 @@
...
    * Creates a LeaseMaintainer to maintain a lease on the connection. This should only be called once for any one
    * Connection.
    *
    * @param connection the connection on which leases should be maintained
-   * @return the LeaseMaintainer that will maintain leases on the connection
+   * @return the CleaningLeaseMaintainer that will maintain leases on the connection
    * @throws LeaseException if the LeaseMaintainer could not be created - probably because the LeaseAcquirer entity is
    * not correctly installed.
    */
-  public static LeaseMaintainer createLeaseMaintainer(Connection connection) {
```
ChrisGreenaway: Don't change the return type to CleaningLeaseMaintainer (in the method declaration and javadoc) just leave it as LeaseMaintainer.

chrisdennis:
Downstream testing seems to indicate things are okay. Merge away.

PR #378, opened 2017-08-10
Make the tracker generic instead of being a message only tracker

henri-tremblay:
There are two commits to this PR. I think both should be revised separately.

The first one is renaming and removing restrictions on the tracker so it can track pretty much anything instead of only messages.

The second one is about removing the typing. My understanding is that a tracker can track multiple types of unrelated messages. This means, it should not have an actual type specified. The same thing applies to the policy. It might access a given type but should be able to tell if any kind of object is trackable. I only kept return type inference to prevent the caller to cast.


henri-tremblay:
test this please

client-message-tracker/src/main/java/org/terracotta/client/message/tracker/OOOMessageHandler.java
```diff
@@ -32,7 +32,7 @@
  * Entities that want once and only once message invocation guarantees can delegate their invokes to this service.
  */
 @CommonComponent
-public interface OOOMessageHandler<M extends EntityMessage, R extends EntityResponse> extends StateDumpable {
+public interface OOOMessageHandler extends StateDumpable {
```
albinsuresh: Am not sure about the gains by this change. Especially since we ended up having method level typing on individual methods. Isn't it better to leave this tying here itself?

henri-tremblay:
Yes. I know. I haven't found a better naming yet.

client-message-tracker/src/main/java/org/terracotta/client/message/tracker/OOOMessageHandler.java
henri-tremblay: I've been wondering that too to be honest. Let's talk about it.

chrisdennis:
Is there an issue associated with this PR?  Whats the underlying motivation behind these changes?

henri-tremblay:
@chrisdennis No there's not. It comes from two things. 
1. Requests that something else than messages needs to be tracked
2. That trackers can receive any kind of things so they shouldn't be typed that much (how much they should still need to be decided)

albinsuresh:
Shouldn't there be a way to retrieve the `ClientTracker` from this service? Currently the provider only provides `OOOMessageHandler`s. I think it should be expanded to provide the trackers directly as well for such use-cases.

henri-tremblay:
@albinsuresh I don't want to guess needs right now. I think in the end we will have another service using the `ClientTracker`. So I don't think adding a service for it is useful right now.

client-message-tracker/src/main/java/org/terracotta/client/message/tracker/Tracker.java
```diff
@@ -0,0 +1,68 @@
...
+   */
+  void track(long id, Object object, Object response);
+
+  /**
+   * Returns the tracked response associated with the given ID, null otherwise.
+   *
+   * @param id Tracked entity ID
+   * @return Tracked entity response
+   */
+  <R> R getTrackedValues(long id);
```
albinsuresh: Shouldn't this be `getTrackedValue` (not with the 's' at the end)?

client-message-tracker/src/main/java/org/terracotta/client/message/tracker/Tracker.java
```diff
@@ -0,0 +1,68 @@
...
+   * @param id Incoming entity ID
+   */
+  void reconcile(long id);
+
+  /**
+   * Get all id - response mappings.
+
+   * @return all tracked responses
+   */
+  Map<Long, Object> getTrackedValues();
```
albinsuresh: The lack of a type for the map value here and the method level typing of `<R> R getTrackedValues(long id);` is making me wonder if the typing at the interface level would be better. Then we would have had the same consistent type across all the methods, no?

client-message-tracker/src/main/java/org/terracotta/client/message/tracker/ClientTracker.java
```diff
@@ -15,39 +15,36 @@
...
-public interface ClientMessageTracker<M extends EntityMessage, R extends EntityResponse> extends StateDumpable {
+public interface ClientTracker<K> extends StateDumpable {
 
   /**
-   * Retrieve the message tracker of the client with the specified {@code clientSourceId}
-   * If a tracker does not exist already create one and return.
+   * Retrieve the tracker of the client with the specified {@code clientSourceId}
+   * If a tracker does not exist already create one and return it.
    *
    * @param clientSourceId a client source id
```
albinsuresh: Argument names still referring to `clientSourceId`

client-message-tracker/src/main/java/org/terracotta/client/message/tracker/Tracker.java
```diff
@@ -0,0 +1,68 @@
...
+ * limitations under the License.
+ */
+package org.terracotta.client.message.tracker;
+
+import org.terracotta.entity.StateDumpable;
+
+import java.util.Map;
+
+/**
+ * Keeps track of an entity's objects with their ids and their corresponding responses.
```
albinsuresh: Still referring to the generic values as responses in arguments and javadoc.

henri-tremblay: Should be alright now I think

client-message-tracker/src/main/java/org/terracotta/client/message/tracker/ClientTracker.java
henri-tremblay: Renamed

client-message-tracker/src/main/java/org/terracotta/client/message/tracker/Tracker.java
henri-tremblay: It was on purpose to be helpful to the caller without forcing homogeneous content in the tracker. But I think you are right , most of the time it will be heterogenous. So I have put typing back for the response.

client-message-tracker/src/main/java/org/terracotta/client/message/tracker/Tracker.java
```diff
@@ -0,0 +1,68 @@
...
+   */
+  void track(long id, Object object, R response);
+
+  /**
+   * Returns the tracked response associated with the given ID, null otherwise.
+   *
+   * @param id Tracked entity ID
+   * @return Tracked entity response
+   */
+  R getTrackedValues(long id);
```
albinsuresh: This is still not addressed. Shouldn't this be get tracked "value" and not "values"?

henri-tremblay: Absolutely right. I missed it. Fixed.

PR #385, opened 2017-08-21
Issue #376 - Lease shutdown cleanup.

chrisdennis:


lease/api/src/main/java/org/terracotta/lease/CleaningLeaseMaintainer.java
```diff
@@ -56,7 +56,13 @@ public boolean waitForLease(long timeout, TimeUnit timeUnit) throws InterruptedE
   @Override
   public void close() throws IOException {
     for (Thread thread : threads) {
-      thread.interrupt();
+      /*
+       * Don't interrupt the calling thread - it's already dying, and
+       * the TC client cannot currently handle rogue interrupts.
+       */
+      if (thread != Thread.currentThread()) {
```
mathieucarbou: Qu: is it worth creating a variable for this since `currentThread()` is native ?

chrisdennis: There are 2 threads in the list and this is only executed on client shutdown - so no I don't think so.

ChrisGreenaway:
Is it true that the current thread is always "on its way out"?

Also, you've got two identical pieces of code - I would factor out a method to interrupt the threads.

lease/api/src/main/java/org/terracotta/lease/LeaseMaintenanceThread.java
```diff
@@ -47,11 +52,21 @@ public void run() {
...
         return;
       } catch (InterruptedException e) {
-        Thread.currentThread().interrupt();
-        return;
+        //reloop and check
       } catch (LeaseException e) {
         LOGGER.error("Error obtaining lease", e);
+      } finally {
+        //force a clean shutdown
+        if (shutdown) {
```
ChrisGreenaway: I don't think you need this as the loop tests the same condition.

lease/api/src/main/java/org/terracotta/lease/LeaseExpiryConnectionKillingThread.java
```diff
@@ -54,12 +59,26 @@ public void run() {
...
         timeSource.sleep(200L);
-      } catch (IOException e) {
-        LOGGER.error("Closing connection, due to lease expiry, caused an error", e);
       } catch (InterruptedException e) {
-        Thread.interrupted();
-        return;
+        //reloop and check
+      } finally {
+        //force a clean shutdown
+        if (shutdown) {
```
ChrisGreenaway: Same test as the loop again.

lease/api/src/main/java/org/terracotta/lease/CleaningLeaseMaintainer.java
```diff
@@ -68,8 +78,12 @@ public void close() throws IOException {
 
   @Override
   public void destroy() throws IOException {
-    for (Thread thread : threads) {
-      thread.interrupt();
+    for (Closeable resource : resources) {
+      try {
+        resource.close();
+      } catch (Throwable t) {
+        LOGGER.info("Exception closing resource {} : {}", resource, t);
```
ChrisGreenaway: Only "info"?

ChrisGreenaway: If you pass the throwable to the variant of the LOGGER method that takes an exception, do you get better exception logging than this version?

lease/api/src/main/java/org/terracotta/lease/CleaningLeaseMaintainer.java
```diff
@@ -68,8 +78,12 @@ public void close() throws IOException {
 
   @Override
   public void destroy() throws IOException {
-    for (Thread thread : threads) {
-      thread.interrupt();
+    for (Closeable resource : resources) {
```
ChrisGreenaway: This code is duplicated between destroy() and close(), consider factoring out a closeResources() method.

lease/api/src/main/java/org/terracotta/lease/LeaseMaintenanceThread.java
chrisdennis: This is intentional - the return clause ensures any exceptions get squelched if `shutdown` is true.

lease/api/src/main/java/org/terracotta/lease/LeaseExpiryConnectionKillingThread.java
chrisdennis: See previous comment.

lease/api/src/main/java/org/terracotta/lease/CleaningLeaseMaintainer.java
chrisdennis: Okay.

lease/api/src/main/java/org/terracotta/lease/CleaningLeaseMaintainer.java
chrisdennis: Probably - changed.

chrisdennis: Yeah - TC is overly noisy, and we have downstream users who are allergic to log messages higher than info.

chrisdennis:
@ljacomet @ChrisGreenaway Re-review is probably advised.  In particular note that I'm *not* interrupting the expiry thread on shutdown - just letting it die more naturally - the extra interrupt noise is not currently good for the core code.

lease/api/src/main/java/org/terracotta/lease/CleaningLeaseMaintainer.java
```diff
@@ -55,23 +61,26 @@ public boolean waitForLease(long timeout, TimeUnit timeUnit) throws InterruptedE
...
-    for (Thread thread : threads) {
-      thread.interrupt();
-    }
-
+    closeResources();
     connection.close();
   }
+
+  private void closeResources() {
+    // We could join on these threads, but that's not a great idea because
```
ljacomet: That join comment no longer applies 😉

chrisdennis: Well - it sorta does - the resources are still threads.  You want me to just delete it then?

ljacomet: From what this class knows, the resources are `Closeable` nothing else. So yes, I would remove the comment.

ljacomet:
The latest batch of changes seem to make sense. The one thing I would like @ChrisGreenaway to confirm is that moving away from interrupts does not risk having these threads remain blocked in some of the calls they are making that interrupts were ... interrupting.

lease/api/src/main/java/org/terracotta/lease/LeaseMaintenanceThread.java
ChrisGreenaway: That's just hard to maintain because it isn't obvious. If that's the behaviour you want then how about explicitly catching Throwable?

ChrisGreenaway:
LeaseExpiryConnectionKillingThread doesn't need to be interrupted. It has three calls in the loop:
getCurrentLease() - which is non-blocking;
connection.close() - and we are trying to avoid an interrupt in this;
Thread.sleep() - which will terminate naturally anyway, as Chris D says.

LeaseMaintenanceThread calls leaseMaintainer.refreshLease() so needs the interrupt otherwise it could block. Chris D has got the interrupt in the close() implementation for that Thread, so we are good.

It would be worth commenting in the LeaseExpiryConnectionKillingThread.close() to explain why it does not interrupt, especially as future maintainers could wonder why these two close() implementations are different.

lease/api/src/main/java/org/terracotta/lease/LeaseExpiryConnectionKillingThread.java
```diff
@@ -54,12 +59,21 @@ public void run() {
...
+          return;
+        }
       }
     }
   }
+
+  @Override
+  public void close() throws IOException {
+    // Only need to shutdown as there are no blocking calls
```
ChrisGreenaway: But more importantly, don't do the interrupt because we don't want to interrupt connection.close()

PR #389, opened 2017-08-31
Add segmentation support for OOOMessageHandler

albinsuresh:


client-message-tracker/src/main/java/org/terracotta/client/message/tracker/OOOMessageHandlerConfiguration.java
```diff
@@ -15,6 +15,7 @@
  */
 package org.terracotta.client.message.tracker;
 
+import org.terracotta.entity.ConcurrencyStrategy;
```
henri-tremblay: Unused import?

client-message-tracker/src/main/java/org/terracotta/client/message/tracker/OOOMessageHandlerImpl.java
```diff
@@ -54,26 +68,26 @@ public R invoke(InvokeContext context, M message, BiFunction<InvokeContext, M, R
 
   @Override
   public void untrackClient(ClientSourceId clientSourceId) {
-    clientMessageTracker.untrackClient(clientSourceId);
+    clientMessageTrackers.stream().forEach(tracker -> tracker.untrackClient(clientSourceId));
```
henri-tremblay: Shouldn't it be removed from `trackedClients` as well?

client-message-tracker/src/main/java/org/terracotta/client/message/tracker/TrackerPolicy.java
```diff
@@ -33,4 +33,8 @@
    * @return true for trackable messages and false for non-trackable messages
    */
   boolean trackable(M message);
+
```
henri-tremblay: I'm not sure to like the `TrackerPolicy` to know the number of segments. Why not just passing the information to the `OOOMessageHandlerImpl` and that's it?

client-message-tracker/src/test/java/org/terracotta/client/message/tracker/OOOMessageHandlerProviderTest.java
```diff
@@ -16,6 +16,7 @@
 package org.terracotta.client.message.tracker;
 
 import org.junit.Test;
+import org.terracotta.entity.ConcurrencyStrategy;
```
henri-tremblay: Again, unused import I think

client-message-tracker/src/test/java/org/terracotta/client/message/tracker/demo/DemoActiveEntity.java
```diff
@@ -30,6 +31,8 @@
 import org.terracotta.entity.ServiceException;
 import org.terracotta.entity.ServiceRegistry;
 
+import static org.mockito.Mockito.mock;
```
henri-tremblay: Unused

client-message-tracker/src/test/java/org/terracotta/client/message/tracker/demo/DemoPassiveEntity.java
```diff
@@ -17,8 +17,8 @@
 
 import org.terracotta.client.message.tracker.OOOMessageHandler;
 import org.terracotta.client.message.tracker.OOOMessageHandlerConfiguration;
-import org.terracotta.entity.ClientDescriptor;
 import org.terracotta.entity.ClientSourceId;
+import org.terracotta.entity.ConcurrencyStrategy;
```
henri-tremblay: Unused

client-message-tracker/src/main/java/org/terracotta/client/message/tracker/TrackerPolicy.java
albinsuresh: Are you suggesting that we move these new segmentation related methods to a different interface like `SegmentationStrategy` or something? And keep `TrackerPolicy` separate as it is today? I'm not against that idea. Was just trying to avoid yet another interface when we already had the minimal `TrackerPolicy`.

client-message-tracker/src/main/java/org/terracotta/client/message/tracker/OOOMessageHandlerImpl.java
albinsuresh: Yes

client-message-tracker/src/main/java/org/terracotta/client/message/tracker/TrackerPolicy.java
henri-tremblay: Yes, but I don't feel they belong there (and it was converted to a Predicate in my other PR). Let me have a look.

henri-tremblay: Ok. So I  would directly pass the number of segments and a `ConcurrencyStrategy`.  The constructor becomes `OOOMessageHandlerImpl(int numberOfSegments, ConcurrencyStrategy<M> concurrencyStrategy, TrackerPolicy<M> trackerPolicy)`. Another solution is to pass the segment index directly to the `invoke` and let the caller provide it. `invoke(InvokeContext context, int segmentIndex, M message, BiFunction<InvokeContext, M, R> invokeFunction)`

henri-tremblay: In fact I don't even need the number of segments. Because it is `concurrencyStrategy.getKeysForSynchronization().size()`. In fact, if `ConcurrencyStrategy` can accept default methods, we could add 

```
default int numberOfSegments() {
    return getKeysForSynchronization().size();
}
```

client-message-tracker/src/main/java/org/terracotta/client/message/tracker/OOOMessageHandlerImpl.java
```diff
@@ -54,26 +68,26 @@ public R invoke(InvokeContext context, M message, BiFunction<InvokeContext, M, R
...
   }
 
   @Override
   public Set<ClientSourceId> getTrackedClients() {
-    return clientMessageTracker.getTrackedClients();
+    return Collections.unmodifiableSet(trackedClients);
```
henri-tremblay: Here I would do this:

```
  @Override
  public Set<ClientSourceId> getTrackedClients() {
    return clientMessageTrackers.stream()
        .flatMap(tracker -> tracker.getTrackedClients().stream())
        .distinct()
        .collect(Collectors.toSet());
  }
```

(of just return the Stream instead of a set)

client-message-tracker/src/main/java/org/terracotta/client/message/tracker/TrackerPolicy.java
albinsuresh: `ConcurrencyStrategy` can not be used precisely for this reason. `concurrencyStrategy.getKeysForSynchronization()` need not be the entire super-set of all possible concurrency keys for all entities. We have seen some "special" entities that allows operations on concurrency keys that are not synchronization keys. In fact I initially had used `ConcurrencyStrategy` for this, but ended up removing those because of such special entities(hence all those unused imports)). That's is when I decided to introduce the segmentation strategy which forces the entity to provide the exact number of segments and the mapping logic.

henri-tremblay: Humkay... But, do we agree that `concurrencyKey()` always return the key for a message? Which means that is we also pass the total number of segments to the constructor, we should be fine.

ljacomet: It does not work. For example, Ehcache uses 16 segments for the backing store, but has 17 concurrency keys. The added one is the first and is reserved for lifecycle operations. So if you pass in the message handler 16 and use the concurrency key, you will end up with messages tied to keys 1 to 16 (instead of 0 to 15) thus most likely ending up in `ArrayIndexOutOfBoundsException` and we could even imagine more complicated concurrency key / passive sync keys schemes.
However, I agree that this does not belong in the same type as message tracking applicability, so there needs to be an added interface there. Something like `TrackerSegmenting` which carries the two methods below.

albinsuresh:
test this please

client-message-tracker/src/main/java/org/terracotta/client/message/tracker/OOOMessageHandlerImpl.java
```diff
@@ -22,25 +22,52 @@
...
+      ClientSourceId clientId = context.getClientSource();
+      int index = segmentationStrategy.applyAsInt(message);
+      Tracker<R> messageTracker = clientMessageTrackers.get(index).getTracker(clientId);
       messageTracker.reconcile(context.getOldestTransactionId());
       R response = messageTracker.getTrackedValue(context.getCurrentTransactionId());
+
+      if (sharedMessageTracker.getTrackedClients().contains(clientId)) {
+        Tracker<R> sharedTracker = sharedMessageTracker.getTracker(clientId);
+        sharedTracker.reconcile(context.getOldestTransactionId());
+        if (response == null) {
```
ljacomet: I would move that if around the `sharedMessageTracker` check as it is cheaper.

client-message-tracker/src/main/java/org/terracotta/client/message/tracker/OOOMessageHandlerImpl.java
```diff
@@ -55,28 +82,38 @@ public R invoke(InvokeContext context, M message, BiFunction<InvokeContext, M, R
...
+  }
+
+  @Override
+  public void loadTrackedResponsesForSegment(int index, ClientSourceId clientSourceId, Map<Long, R> trackedResponses) {
+    this.clientMessageTrackers.get(index).getTracker(clientSourceId).loadOnSync(trackedResponses);
   }
 
-  @SuppressWarnings("unchecked")
   @Override
   public void loadOnSync(ClientSourceId clientSourceId, Map<Long, R> trackedResponses) {
```
ljacomet: I would make this `@Deprecated` to be very clear it should not be used in normal scenario.

albinsuresh:
test this please

PR #391, opened 2017-09-11
[Fix2] TDB-2310: Close #358: M&M ignores all potential inconsistencies coming from voltron events

### LINKED ISSUE #358: M&M should handle inconsistent platform monitoring events
if a user or a code error causes M&M to throw an exception, it should be sent or reported but should not crash the server.

=> review all throws from the monitoring service when voltron is calling us

CC @myronkscott @anthonydahanne @ljacomet @lorban : I'll let you add other details if needed, thanks 😃

mathieucarbou:
This PR makes M&M supports bad clients with the same identifiers, and also bad sequence order sent from voltron.

By supporting, I mean that if something inconsistent is found, then the call is ignored.

Side effects:

1. the topology will only see 1 client
2. if a second client connect with the same client identifier as a first connected client, then both client will be able to send their notification and stats and they will have the same exact origin.
3. management calls will be received by these 2 clients BUT the NMS will only receive the first answer
4. if one of the client disconnects, it will be removed from the topology and if the other clients are sending stats or notifs, they will be ignored. Management calls won't be possible also
5. NMS only sees 1 client

### DIFF
--- a/management/monitoring-service-api/src/main/java/org/terracotta/management/service/monitoring/registry/provider/ClientBinding.java
+++ b/management/monitoring-service-api/src/main/java/org/terracotta/management/service/monitoring/registry/provider/ClientBinding.java
@@ -29,9 +29,13 @@ public class ClientBinding {
 
   volatile ClientIdentifier resolvedClientIdentifier;
 
+  public ClientBinding(ClientDescriptor clientDescriptor) {
+    this(clientDescriptor, null);
+  }
+  
   public ClientBinding(ClientDescriptor clientDescriptor, Object value) {
     this.clientDescriptor = Objects.requireNonNull(clientDescriptor);
-    this.value = Objects.requireNonNull(value);
+    this.value = value;
   }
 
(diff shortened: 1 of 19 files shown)


management/monitoring-service-api/src/main/java/org/terracotta/management/service/monitoring/registry/provider/ClientBinding.java
```diff
@@ -29,9 +29,13 @@
 
   volatile ClientIdentifier resolvedClientIdentifier;
 
+  public ClientBinding(ClientDescriptor clientDescriptor) {
+    this(clientDescriptor, null);
+  }
+  
   public ClientBinding(ClientDescriptor clientDescriptor, Object value) {
     this.clientDescriptor = Objects.requireNonNull(clientDescriptor);
-    this.value = Objects.requireNonNull(value);
```
mathieucarbou: needed by @henri-tremblay

mathieucarbou:
relaced by #394

PR #393, opened 2017-09-14
[Fix2] TDB-2403: Makes M&M more reliable against voltron inconsistencies

mathieucarbou:
- better support of passive entities placeholders that are not cleaned up by platform when server becomes active
- filter out any event coming from active in DefaultDataListener (DefaultDataListener should only receive passive events but was receiving events from the passive placeholders created on a server before it becomes active)


mathieucarbou:
replaced by #394

PR #396, opened 2017-10-04
Closes #395: Collected statistics are not send to every NMS entities and active management registries are not destroyed after entity promotion (failover)

### LINKED ISSUE #395: bug: statistic are sent from NMS entity even if its stat collector is not started
In the case where several nms entities are created, each nms entity will receive all stats generated from every stat colletors, even if they are not started.

I.e. if you connect 2 NMS clients, and start collecting each 1 sec in one and each 5 sec in the other one, then both will receive all the collected stats at both 1 sec and 5 sec interval.

Also, if 2 NMS clients are connected and only 1 start collecting stats, they are all receiving the collected stats from the started collector, even if one collector is not started.

@anthonydahanne : FYI

mathieucarbou:
This is an important bugfix of 2 issues discovered during the PoC for external management tool integration.

See #395.

Basically:

1. If several NMS entities were created, their statistic collector was sending all stats to everyone of them. So if a stat collector of NMS 1 was configured to send all server stats at 1 sec interval and the other one was configured to send all server stats at 5 sec interval, both of them were receiving stats at 1 sec and 5 sec interval

2. After a failover, the management registry of the active entity that was promoted was closed right away, its stats collector included. We didn't see this bug because during a promotion, passive entities are not destroyed: their management registry was kept alive and their stat collector also. So stats were generated by a stat collector inside a registry of a passive entity.

To overcome that, seen with @myronkscott, active entities needs to make the management layer aware that their promotion was successful and then discard the old passive management registry.

__This is done by adding this to the active entities using management:__

Important
=======

This __needs to be applied to Ehcache and other projects__. If not, after a failover, the statistic collector of the dead passive entity will still run and the NMS client will receive a duplicate set of stats from the passive AND active statistic collector.

```
  @Override
  public void loadExisting() {
    [...]
    entityManagementRegistry.cleanupPreviousPassiveStates();
    entityManagementRegistry.refresh();
    [...]
  }
```


management/monitoring-service/src/main/java/org/terracotta/management/service/monitoring/DefaultStatisticService.java
```diff
@@ -52,19 +52,28 @@
...
   @Override
-  public StatisticCollector createStatisticCollector(StatisticCollector.Collector collector) {
-    LOGGER.trace("[0] createStatisticCollector()");
+  public StatisticCollector createStatisticCollector(EntityManagementRegistry statCollectorRegistry, StatisticCollector.Collector collector) {
+    long consumerId = statCollectorRegistry.getMonitoringService().getConsumerId();
+    LOGGER.trace("[{}] createStatisticCollector()", consumerId);
     return new DefaultStatisticCollector(
-        capabilityManagementSupport,
+        // create a stat collector that can collect stats over all management registries
+        // and only the registry combined
```
GaryWKeim: Make these proper sentences... capitalization, punctuation.

management/monitoring-service/src/main/java/org/terracotta/management/service/monitoring/DefaultStatisticService.java
```diff
@@ -52,19 +52,28 @@
...
+        // this will avoid collecting stats on a registry from another management entity 
+        // that already has its own stat collector
+        new CombiningCapabilityManagementSupport(sharedEntityManagementRegistry, statCollectorRegistry),
         managementScheduler,
-        collector
+        list -> {
+          // add a marker on the statistics to know which collected has collected them
```
GaryWKeim: which collector has

management/monitoring-service/src/main/java/org/terracotta/management/service/monitoring/DefaultEntityManagementRegistry.java
```diff
@@ -53,24 +53,27 @@
...
-    sharedManagementRegistry.addManagementService(this);
-  }
 
-  @Override
-  public void onBecomeActive() {
-    if (monitoringService instanceof DefaultPassiveEntityMonitoringService) {
-      close();
-    }
+    // the sharedManagementRegistry only contains management registries of normal entities.
+    // management registries of NMS entities (which contain the stat collector), is not included
```
anthonydahanne: management registries of NMS entities (which contain the **stats** collector), **are** not included

PR #401, opened 2017-10-15
Bump Terracotta versions

ljacomet:
Includes aligning logback version with the one from terracotta-core

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -28,11 +28,11 @@
     <maven.version>3.2.5</maven.version>
     <maven-forge-plugin.version>1.2.16</maven-forge-plugin.version>
     <slf4j.version>1.7.7</slf4j.version>
-    <logback.version>1.1.3</logback.version>
-    <terracotta-apis.version>1.4.0-pre3</terracotta-apis.version>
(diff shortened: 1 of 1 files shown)


mathieucarbou:
@ljacomet @myronkscott  : there has been any changes in the reconfigure entity lifecycle recently ?

I am calling entityRef.reconfigure

old behavior:

1. receive a notification server entity destroyed (for old one)
2. receive 1 notification s for entity creation (new one)
3. receive 2 notifications for client attach (re-fetched entity)

new (wrong ?) behavior:

1. I do not get an entity destroyed event for the old entity
2. I am receiving TWICE the same notification for entity created for the same entity consumer id
3. I am receiving the 2 client attach events BUT one come BEFORE entity creation events (???)

So yes, there are several issues here… the events that platform send us should be in the correct order ans one event is missing now. I.e. not normal to receive:

1. entity 1 creation
2. client attached to 1
3. client attached to 1
4. entity 1 creation

and the destroy event has disappeared.

mathieucarbou:
Replaced by #403

PR #404, opened 2017-11-07
Config generator

anthonydahanne:


tc-config-generator/src/main/java/org/terracotta/config/generator/GenerateFromEnvironmentVariables.java
```diff
@@ -0,0 +1,158 @@
...
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+public class GenerateFromEnvironmentVariables {
+
+  private static final Logger LOGGER = LoggerFactory.getLogger(GenerateFromEnvironmentVariables.class);
+  static final String TCCONFIG_ = "TCCONFIG_";
+  private static final String GENERIC_ERROR_MESSAGE = "You should call the tc-config-generator with exactly four arguments.\n" +
+      "Examples : \n" +
+      "java -jar tc-config-generator -i /Users/anthony/tc-template.ftlh -o /Users/anthony/tc-config.xml";
```
mathieucarbou: `anthony ` is my friend 😄

anthonydahanne: héhé 😄

tc-config-generator/README.md
```diff
@@ -0,0 +1,90 @@
...
+export TCCONFIG_SERVERS0_NAME=server11
+export TCCONFIG_SERVERS0_HOST=server11
+export TCCONFIG_SERVERS0_PORT=9410
+export TCCONFIG_SERVERS0_GROUP_PORT=9430
+export TCCONFIG_SERVERS1_NAME=server12
+export TCCONFIG_SERVERS1_HOST=server12
+export TCCONFIG_SERVERS1_PORT=9410
+export TCCONFIG_SERVERS1_GROUP_PORT=9430
+export TCCONFIG_OFFHEAPS0_NAME=offheap1
+export TCCONFIG_OFFHEAPS0_UNIT=MB
```
mathieucarbou: we could force unit to MB to avoid having a lot of env vars ?

anthonydahanne: well, it's a matter of default template when you think about it; this is just an example.
I guess the kits builders will be the modules attaching the default template.
And yes, probably unit could be a variable like ```TCCONFIG_OFFHEAP_UNIT``` , and in the template : ```${offheapUnit}```

anthonydahanne:
@mathieucarbou well, this person will be able to either use his /her own tc-config.xml, or use his / her own template.
That will be part of the "advanced" configurations use cases, for which the users need to come up with their own config file or template.
The scope of the default template is for "first time experience" - using megatron is clearly an advanced use case

anthonydahanne:
some one please merge ! it's rebased ! thanks

mathieucarbou:
still 1 pending check ? build stuck ?

anthonydahanne:
nope, all green :-)

PR #405, opened 2017-11-08
Bump slf4j and logback versions

ljacomet:


ljacomet:
test this please

mathieucarbou:
@ljacomet : these 2 seems to pass sometimes - it seems that depending on the speed where tms entity can reconnect after a failover, we lose some notifications...

mathieucarbou:
See #406

mathieucarbou:
Replaced by #407

PR #406, opened 2017-11-08
Fixed unreliable M&M tests

mathieucarbou:
Passed locally. Hope it will pass the build here also... I am not able to reproduce test failures locally.

We cannot test notifications after a failover because if the NMS entity is the last one to reconnect, then we have lost everything sent previously from other clients.

### DIFF
--- a/management/testing/integration-tests/src/test/java/org/terracotta/management/integration/tests/FailoverIT.java
+++ b/management/testing/integration-tests/src/test/java/org/terracotta/management/integration/tests/FailoverIT.java
@@ -17,14 +17,11 @@ package org.terracotta.management.integration.tests;
 
 import org.junit.Before;
 import org.junit.Test;
+import org.terracotta.management.model.cluster.AbstractManageableNode;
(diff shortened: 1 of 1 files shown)


management/testing/integration-tests/src/test/java/org/terracotta/management/integration/tests/FailoverIT.java
```diff
@@ -84,18 +79,6 @@ public void all_registries_reexposed_after_failover() throws Exception {
   }
 
   @Test
-  public void notifications_after_failover() throws Exception {
```
anthonydahanne: ah... we don't make assertions on the new server joining after fail over anymore ?

mathieucarbou: it's implicitly done in `FailoverIT.setup()`

anthonydahanne: ok

PR #411, opened 2017-11-15
upgrade to statistics 1.5.0

mathieucarbou:
waiting to be able to release statistics v1.5.0...

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -33,7 +33,7 @@
     <terracotta-configuration.version>10.4.0-pre6</terracotta-configuration.version>
     <passthrough-testing.version>1.4.0-pre6</passthrough-testing.version>
     <terracotta-core.version>5.4.0-pre8</terracotta-core.version>
-    <statistics.version>1.4.3</statistics.version>
+    <statistics.version>1.5.0</statistics.version>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
     <java.version>1.8</java.version>


chrisdennis:
+1 to what Henri said.

mathieucarbou:
@chrisdennis @henri-tremblay : this is because the latency support requires some changes to be made in sync in statistics + tc-platform + Ehcache so that everything can work correctly and be tested also with real use cases.

Currently, statistics cannot be used as a snapshot in Ehache.

The stuff I'll add won't break api: it will be additions. It would have be nice that after some big changes a release is issued because otherwise who will have to take care of the big PR the twill include both the compatibility for the upgrade + the latency work ? M&M team of course...

test this please

henri-tremblay:
Why can't statistics be used as a snapshot? That seems to be the real problem no?

mathieucarbou:
@henri-tremblay : yes! And this is exactly the problem I want to solve by making Ehcache compatible with this version ;-)

henri-tremblay:
Still don't get it. Why no using the latest statistics snapshot in ehcache? Fix ehcache and then release everything.

mathieucarbou:
Because ehcache like any other project must not depend on snapshots...

anthonydahanne:
my 2 cents here ...
* ehcache3 just upgraded from 3.4 to 3.5 for java 8 as minimum version, so I'm ok with 1.4 to 1.5, at least for consistency
* relying on released versions i a good thing. If you want to test things out, well, do it locally and make your own local releases; that said if there are only minor changes to adapt to stats 1.5 and we are confident we can fix them quickly, let's do this release now

mathieucarbou:
@henri-tremblay : in conclusion, we can merge... Yes this is not ideal but since we do not have -pre releases for this library (like tc-platform) we don't have other choices... Seen with @anthonydahanne @ljacomet @chrisdennis this morning.

PR #413, opened 2017-11-21
passive entity's management registry (and statistic collector) was not closed after failover

mathieucarbou:

PR #416, opened 2017-12-01
Make entity management service common to all entities

myronkscott:


mathieucarbou:
I will put these ideas in https://github.com/Terracotta-OSS/terracotta-platform/pull/413/commits with the changes we talked about. Thanks!

PR #418, opened 2017-12-19
IPv6 changes

GaryWKeim:
More changes in terracotta-core and terracotta-enterprise.

### DIFF
--- a/management/cluster-topology/src/main/java/org/terracotta/management/model/cluster/Endpoint.java
+++ b/management/cluster-topology/src/main/java/org/terracotta/management/model/cluster/Endpoint.java
@@ -45,7 +45,11 @@ public final class Endpoint implements Serializable {
 
   @Override
   public String toString() {
-    return getAddress() + ":" + getPort();
+    String address = getAddress();
+    if (address.contains(":")) {
+      address = "[" + address + "]";
+    }
+    return address + ":" + getPort();
   }
 
   @Override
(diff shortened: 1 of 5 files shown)


management/cluster-topology/src/main/java/org/terracotta/management/model/cluster/Endpoint.java
```diff
@@ -75,8 +79,18 @@ public static Endpoint create(String address, int port) {
   }
 
   public static Endpoint valueOf(String str) {
-    String[] parts = str.split(":");
-    return create(parts[0], Integer.parseInt(parts[1]));
+    int lastColon = str.lastIndexOf(':');
+    String host = str.substring(0, lastColon);
```
mathieucarbou: it wasn't handled before, but I would also add something like:
`if(lastColon == -1) { throw new IllegalArgumentException('Invalid endpoint format. Required format is<ip>:<port>.') }`

management/management-model/src/main/java/org/terracotta/management/model/cluster/ClientIdentifier.java
```diff
@@ -130,12 +136,21 @@ public static ClientIdentifier create(String name, String logicalConnectionUid)
 
   public static ClientIdentifier valueOf(String identifier) {
     try {
-      int copy = identifier.indexOf('@');
-      int firstColon = identifier.indexOf(':', copy + 1);
+      int ampIdx = identifier.indexOf('@');
```
mathieucarbou: Could you add a test for different patterns in `ClientIdentifierTest` ?

PR #430, opened 2018-01-15
Leased connection api

AbfrmBlr:


lease/api/pom.xml
```diff
@@ -55,6 +55,11 @@
       <artifactId>entity-client-api</artifactId>
       <scope>provided</scope>
     </dependency>
+    <dependency>
+      <groupId>org.terracotta.internal</groupId>
```
ChrisGreenaway: Why the extra dependency?

lease/api/src/main/java/org/terracotta/lease/CleaningLeaseMaintainer.java
```diff
@@ -71,6 +72,15 @@ public void destroy() throws IOException {
     connection.close();
   }
 
+  @Override
+  public void setDisconnectListener(LeasedConnection.DisconnectedEventListener listener) {
```
ChrisGreenaway: Perhaps DisconnectedEventListener can be an interface rather than an inner interface?

lease/api/src/main/java/org/terracotta/lease/LeaseMaintainerImpl.java
```diff
@@ -151,4 +153,19 @@ private void logLease(boolean updated, Lease previousLease, Lease newLease) {
   public void destroy() throws IOException {
     throw new UnsupportedOperationException();
   }
+
+  @Override
+  public void setDisconnectListener(LeasedConnection.DisconnectedEventListener listener) {
+    if (this.listener == null) {
```
ChrisGreenaway: There's a race here.

lease/api/src/main/java/org/terracotta/lease/LeaseMaintainerImpl.java
```diff
@@ -151,4 +153,19 @@ private void logLease(boolean updated, Lease previousLease, Lease newLease) {
...
+      this.listener = listener;
+    } else {
+      throw new IllegalStateException("Disconnected Event Listener already registered");
+    }
+  }
+
+  public void leaseExpired() {
+    if (listener != null) {
```
ChrisGreenaway: There's also a race here if someone calls setDisconnectListener(null). Do you want to allow unsetting with null? If not, prevent it.

lease/api/src/main/java/org/terracotta/lease/connection/BasicLeasedConnection.java
```diff
@@ -83,7 +84,13 @@ public void close() throws IOException {
   }
 
   @Override
-  public LeaseMaintainer getLeaseMaintainer() {
-    return leaseMaintainer;
+  public boolean isConnectionValid() {
+    Lease currentLease = leaseMaintainer.getCurrentLease();
```
ChrisGreenaway: This implementation of isConnectionValid() is equivalent to "has valid lease".

What was needed before was to know that the lease was continuously held over the duration of an operation. In other words, it would get a lease before and after and then check isValidAndContiguous() from one to the other.

I thought we wanted isConnectionValid() to return true if and only iff the lease had never been lost. Which means that when you first get a lease you hold on to that and then isConnectionValid() gets the current lease and does isValidAndContiguous() against the first lease.

Not sure what is supposed to happen before the connection gets a lease issued to it, or how the code knows when the *first* lease is available though...

lease/api/src/main/java/org/terracotta/lease/connection/LeasedConnection.java
```diff
@@ -19,8 +19,24 @@
 import org.terracotta.lease.LeaseMaintainer;
 
 /**
- *
+ * Connection provided on lease by server
```
ChrisGreenaway: I don't understand what "Connection provided on lease by server" means.

lease/api/src/main/java/org/terracotta/lease/connection/LeasedConnection.java
```diff
@@ -19,8 +19,24 @@
...
  */
 public interface LeasedConnection extends Connection {
-  LeaseMaintainer getLeaseMaintainer();
+
+  /**
+   * returns true if connection is still valid, else false
```
ChrisGreenaway: This should explain what is meant by "is still valid".

It should also say what will be returned if the lease has not yet been issued. Perhaps you need a different return type to indicate it hasn't been issued, or make sure that a LeasedConnection can't be obtained until it has a lease.

lease/api/src/main/java/org/terracotta/lease/connection/LeasedConnectionFactory.java
```diff
@@ -0,0 +1,52 @@
...
+   * @param uri URI to connect to
+   * @param properties any configurations to be applied (implementation specific)
+   * @return an established connection
+   * @throws ConnectionException if there is an error while attempting to connect
+   */
+  public static LeasedConnection connect(URI uri, Properties properties) throws ConnectionException {
+
+    Collection<LeasedConnectionService> serviceLoaders = ManagedServiceLoader.loadServices(LeasedConnectionService.class,
+            LeasedConnectionService.class.getClassLoader());
+    for (LeasedConnectionService connectionService : serviceLoaders) {
```
ChrisGreenaway: This seems weird - now every ConnectionService needs converting to a LeaseConnectionService?

Isn't there some way to just use ConnectionService and then layer the leasing over that.

I think what I'm saying is that why have the extra infrastructure of service loading for LeasedConnectionService - why not just push LeasedConnectionServiceImpl into LeaseConnectionFactory?

lease/api/src/main/java/org/terracotta/lease/connection/LeasedConnectionServiceImpl.java
```diff
@@ -0,0 +1,53 @@
...
+  @Override
+  public boolean handlesURI(URI uri) {
+    return SCHEME.equals(uri.getScheme());
+  }
+
+  @Override
+  public LeasedConnection connect(URI uri, Properties properties) throws ConnectionException {
+    Connection connection = ConnectionFactory.connect(uri, properties);
+
+    LeasedConnection leasedConnection = BasicLeasedConnection.create(connection, createTimeBudget(properties));
```
ChrisGreenaway: The TimeBudget should have been created before the call to ConnectionFactory.connect(), otherwise you are just timing getting the lease, but the budget is supposed to cover the whole connection.

Ideally there would be a way of passing the TimeBudget into the ConnectionFactory.connect(), but as it is the first call, it can manage its own TimeBudget.

lease/api/src/main/java/org/terracotta/lease/connection/LeasedConnectionServiceImpl.java
```diff
@@ -0,0 +1,53 @@
...
+
+    return leasedConnection;
+  }
+
+  private static TimeBudget createTimeBudget(Properties properties) {
+    String timeoutString = properties.getProperty(ConnectionPropertyNames.CONNECTION_TIMEOUT,
```
ChrisGreenaway: Inelegant that you convert the default timeout to a string to be parsed back to a long.

lease/api/src/main/java/org/terracotta/lease/connection/TimeBudget.java
```diff
@@ -17,15 +17,15 @@
 
 import java.util.concurrent.TimeUnit;
 
-public class TimeBudget {
```
ChrisGreenaway: This class should almost definitely be public as it is generally useful. It probably should not sit inside leasing either.

lease/api/src/test/java/org/terracotta/lease/LeaseMaintainerImplTest.java
```diff
@@ -165,6 +167,15 @@ public void waitForLeaseWithLease() throws Exception {
     leaseMaintainer.waitForLease();
   }
 
+  @Test
+  public void leaseExpiryListenerDoubleRegistration() {
+    exception.expect(IllegalStateException.class);
+    leaseMaintainer.setDisconnectListener(() -> {});
+
+    leaseMaintainer.setDisconnectListener(() -> {});
```
ChrisGreenaway: Add some tests for setDisconnectListener(null).

lease/api/pom.xml
AbfrmBlr: this one is for `ManagedServiceLoader`

lease/api/src/main/java/org/terracotta/lease/connection/BasicLeasedConnection.java
AbfrmBlr: I am confused by this explanation here, as this is the exact logic we use to determine whether client kills the connection or not in connection killing thread. We need to talk then

lease/api/src/main/java/org/terracotta/lease/connection/LeasedConnectionFactory.java
AbfrmBlr: No ConnectionService will only be one always, and that will only provide Connnection.
LeasedConnnectionService is layering on top of that Connection, therefore all the entities will end up using LeasedConnectionFactory. I think we alongwith @ljacomet talked about this, let's talk again if you need further clarification

lease/api/src/main/java/org/terracotta/lease/connection/LeasedConnectionServiceImpl.java
AbfrmBlr: Right now no one is using it, Will be done when someone really uses it or needs it, till then I would leave it like this

lease/api/src/main/java/org/terracotta/lease/connection/TimeBudget.java
AbfrmBlr: Right now no one is using it, Will be done when someone really uses it or needs it, till then I would leave it like this

lease/api/src/main/java/org/terracotta/lease/CleaningLeaseMaintainer.java
AbfrmBlr: why, I think it kind of makes sense since it is related to LeasedConnection

lease/api/src/main/java/org/terracotta/lease/LeaseMaintainerImpl.java
AbfrmBlr: You mean lease expiry event even before the listener is set??

AbfrmBlr: Meanwhile I just realized that it will allow multiple listeners as same connection can be used by multiple entities and each of them will be interested in disconnect event

PR #431, opened 2018-01-16
Adding lease connection service

AbfrmBlr:


lease/api/pom.xml
```diff
@@ -55,6 +55,11 @@
       <artifactId>entity-client-api</artifactId>
       <scope>provided</scope>
     </dependency>
+    <dependency>
+      <groupId>org.terracotta.internal</groupId>
+      <artifactId>client-runtime</artifactId>
```
ljacomet: Ideally find the exact binary that has the dependency and add it as `provided`

AbfrmBlr: sure, that makes sense

AbfrmBlr: done

PR #432, opened 2018-01-24
Statistics upgrade and do not include it in mnm-common shaded jar

mathieucarbou:
Alsof fixes M&M uber jar to not include the stat lib inside but only m&n stuff.

### DIFF
--- a/management/distribution/mnm-common/pom.xml
+++ b/management/distribution/mnm-common/pom.xml
@@ -92,6 +92,7 @@
               <artifactSet>
                 <excludes>
                   <exclude>org.slf4j:slf4j-api</exclude>
+                  <exclude>org.terracotta:statistics</exclude>
                 </excludes>
               </artifactSet>
             </configuration>
--- a/pom.xml
+++ b/pom.xml
@@ -33,7 +33,7 @@
     <terracotta-configuration.version>10.4.0-pre9</terracotta-configuration.version>
     <passthrough-testing.version>1.4.0-pre9</passthrough-testing.version>
     <terracotta-core.version>5.4.0-pre18</terracotta-core.version>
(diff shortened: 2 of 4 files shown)


pom.xml
```diff
@@ -33,7 +33,7 @@
     <terracotta-configuration.version>10.4.0-pre9</terracotta-configuration.version>
     <passthrough-testing.version>1.4.0-pre9</passthrough-testing.version>
     <terracotta-core.version>5.4.0-pre18</terracotta-core.version>
-    <statistics.version>2.0.2</statistics.version>
+    <statistics.version>2.0-SNAPSHOT</statistics.version>
```
mathieucarbou: Need a new releaser after https://github.com/Terracotta-OSS/statistics/pull/49 is merged

management/distribution/mnm-common/pom.xml
```diff
@@ -92,6 +92,7 @@
               <artifactSet>
                 <excludes>
                   <exclude>org.slf4j:slf4j-api</exclude>
+                  <exclude>org.terracotta:statistics</exclude>
```
anthonydahanne: can you explain to me the stats exclude ? thanks

mathieucarbou: Something we discussed with @ljacomet  a long time ago. Our `mnm-commo`n uber jar is wrong because it should only contains M&M code and not external libs. The stats lib is used also by Ehcache and others as a dependency without the management api.
So now the effect is that we have in the classpath twice the stat code, as a normal transitive dependency and also included inside `mnm-common`

mathieucarbou:
test this please

mathieucarbou:
test this please

PR #437, opened 2018-02-06
Cancelling management calls can create a while loop in case a management call does not return from server and is never removed from the list

mathieucarbou:


management/nms-entity/nms-entity-client/src/main/java/org/terracotta/management/entity/nms/client/DefaultNmsService.java
```diff
@@ -165,7 +165,12 @@ public Message waitForMessage(long time, TimeUnit unit) throws InterruptedExcept
   public void cancelAllManagementCalls() {
     InterruptedException stopper = new InterruptedException();
     while (!managementCallAnswers.isEmpty()) {
-      managementCallAnswers.values().forEach(f -> f.completeExceptionally(stopper));
+      managementCallAnswers.forEach((k, v) -> {
```
anthonydahanne: could iterate on keyset, instead of entrySet : managementCallAnswers.keys().foreach

mathieucarbou: yes better, i was using v before

PR #445, opened 2018-02-21
Fix Ehcache#2289: Call addNode() on different paths

mathieucarbou:
See also https://github.com/ehcache/ehcache3/issues/2289.

After discussing with @myronkscott, actually this is a misunderstanding about how the voltron tree is working especially during entity promotion.

So we have to use separate path for the different message types. 

- There is at most 1 registry per entity so using 1 path for this is OK
- There can be several management call answers per entity, so the new path will allow only at most one to remain after a failover, so other calls will time out
- Same for notifications. They will be lost except the last one

When no failover (normal usage), nothing is lost and all consecutive messages, even on the same path, are sent to the monitoring service.

__CHANGE__ All entities using management must now call call `registry.entityCreated()` in the `createNew()` method for active and passive.

@myronkscott : please correct me if I am wrong.

### DIFF
--- a/management/monitoring-service-api/src/main/java/org/terracotta/management/service/monitoring/EntityManagementRegistry.java
+++ b/management/monitoring-service-api/src/main/java/org/terracotta/management/service/monitoring/EntityManagementRegistry.java
@@ -108,8 +108,13 @@ public interface EntityManagementRegistry extends CapabilityManagementSupport, C
   /**
    * Must be called in ServerEntity.loadExisting() method to clear the previous
    * management registry created by the passive entity once the promotion to active
-   * is completed
+   * is completed, and initialize the monitoring service for this entity
    */
   void entityPromotionCompleted();
 
+  /**
+   * Must be called in ServerEntity.createNew() method to initialize the monitoring service
+   * for this entity
+   */
+  void entityCreated();
 }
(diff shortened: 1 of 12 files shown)


anthonydahanne:
LGTM - I'll let @myronkscott confirm

management/monitoring-service-api/src/main/java/org/terracotta/management/service/monitoring/EntityManagementRegistry.java
```diff
@@ -112,4 +112,10 @@ default void unregisterAndRefresh(Object managedObject) {
    */
   void entityPromotionCompleted();
 
+  /**
+   * Must be called in ServerEntity.createNew() method to clear the previous
+   * management registry created by the passive entity once the promotion to active
+   * is completed
+   */
```
myronkscott: ?? createNew() is only called when the entity is created.  loadExisting() is called when an ActiveEntity is first loaded or promotion from passive occurs.

mathieucarbou: @myronkscott : this is because in all cases, I now need to setup the initial tree structure for the entity. See the changes in `DefaultEntityMonitoringService`:

```
+  public void init() {
+    LOGGER.trace("[{}] init()", getConsumerId());
+    monitoringProducer.addNode(new String[0], RELIABLE_CHANNEL_KEY, null);
+    Stream.of(REGISTRY, MANAGEMENT_ANSWER, NOTIFICATION)
+        .forEach(type -> monitoringProducer.addNode(new String[]{RELIABLE_CHANNEL_KEY}, type.name(), null));
+  }
```

PR #471, opened 2018-05-11
Close #470: User input tags ordering should be kept

### LINKED ISSUE #470: clustered client tags ordering should be kept
Users can tag a clustered client. Tags should be unique (put in a set), but the way we store them we should keep as much as possible user's ordering. 

I.e. if a user is ordering his tags for example for readiness like this: `"webapp-1", "server-2", "cluster-1"` I think it is nice if we are able to also show these tags in the same order., because he did this for a reason. 

Tags are user's inputs. And as such, I strongly think that for a good experience a system should output user's inputs as close as they entered it.

From a user experience, if I were to tag a client with some tags in a specific ordering for visibility such as: "webapp-1", "server-1", "cluster-1", then I would expect the system to output the tags I have entered in the same order

So it is juts cosmetic. There is no technical reason for doing that. It is just a way to be consistent and nice to the user and not reorder the inputs he gave to us

Also I think that:
- using TreeSet is worse because it forces an ordering that should not be decided by the backend, but much more in the UI if we want an ASCII ordering (or reverse one)
- using simple HashSet is the worst of them because it would output random ordering for the tags and thus the list displayed in the UI won't be consistent from time to time

mathieucarbou:
See description and reasons inside #470

### DIFF
--- a/management/cluster-topology/src/main/java/org/terracotta/management/model/cluster/Client.java
+++ b/management/cluster-topology/src/main/java/org/terracotta/management/model/cluster/Client.java
@@ -20,12 +20,11 @@ import org.terracotta.management.model.context.Context;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.Comparator;
+import java.util.LinkedHashSet;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
-import java.util.SortedSet;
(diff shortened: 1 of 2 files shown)


mathieucarbou:
VM crashed....

test this please

PR #475, opened 2018-06-28
Fix race condition

mathieucarbou:


mathieucarbou:
my modification is not working and I do not not why...

mathieucarbou:
ok now all local tests passed.

PR #504, opened 2018-12-11
bump platform versions

ramsai1729:


rkavanap:
hmm.. looks like we have to bump up all versions now..is that correct?

PR #510, opened 2019-01-30
Add setCapacity() to OffHeapResource

ChrisGreenaway:


offheap-resource/src/main/java/org/terracotta/offheapresource/OffHeapResourceImpl.java
```diff
@@ -104,23 +102,32 @@ public OffHeapResourceBinding getManagementBinding() {
...
-          remainingUpdated(current - size);
-          return true;
-        }
+    }
+
+    while (true) {
+      OffHeapResourceState currentState = state.get();
+      OffHeapResourceState newState = currentState.reserve(size);
+
+      if (newState.getUsed() > newState.getCapacity()) {
```
mathieucarbou: It would be nice to have something like: `if (newState.overflows()) {` to move the logic into the `OffHeapResourceState` object since the fields are there.

offheap-resource/src/main/java/org/terracotta/offheapresource/OffHeapResourceImpl.java
```diff
@@ -163,12 +177,33 @@ public void release(long size) throws IllegalArgumentException {
...
+
+    while (true) {
+      OffHeapResourceState currentState = state.get();
+      OffHeapResourceState newState = currentState.withCapacity(size);
+
+      if (newState.getUsed() > newState.getCapacity()) {
+        return false;
+      }
+
+      if (state.compareAndSet(currentState, newState)) {
```
mathieucarbou: I'd like to see M&M wired too here, like for the threshold call, something like:

`onCapacityChanged.accept(this, [oldState, newState]);`

And then in `OffHeapResourcesProvider` we would just need to add this piece of code that is triggering the event:

```
(res, states) -> {
       def oldState = states[0]
       def newState = states[1]
        for (EntityManagementRegistry registry : registries) {
          Map<String, String> attrs = new HashMap<>();
          attrs.put("oldCapacity", String.valueOf(oldState.getCapacity()));
          attrs.put("newCapacity", String.valueOf(newState.getCapacity()));
          registry.pushServerEntityNotification(res.getManagementBinding(), "OFFHEAP_RESOURCE_CAPACITY_CHANGED", attrs);
        }
      }
```

As simple as that, and this does not trigger any comm: just put the event in a queue. So this is fast.

anthonydahanne: thanks for thinking of m&m @mathieucarbou , we definitely want M&M to know that a specific offheap resource was changed, so that we share this new value with all m&m consumers.
That would even need its own test

offheap-resource/src/main/java/org/terracotta/offheapresource/OffHeapResourceImpl.java
anthonydahanne: +1

offheap-resource/src/main/java/org/terracotta/offheapresource/OffHeapResourceImpl.java
anthonydahanne: maybe in stateUpdated ?

mathieucarbou: no because state updated is only related to used space over the capacity to send threashold events. new used space is already updated when this method is executed. So it should be before. M&M should not receive the new threshold events before the capacity change event otherwise this would look very weird in the event panel...

offheap-resource/src/main/java/org/terracotta/offheapresource/OffHeapResourceImpl.java
ChrisGreenaway: Done

offheap-resource/src/main/java/org/terracotta/offheapresource/OffHeapResourceImpl.java
ChrisGreenaway: Done, but not in stateUpdated() - just in the method that changes the capacity - makes more sense there.

myronkscott:
@ChrisGreenaway please rebase and I will merge

PR #511, opened 2019-02-07
bump underlying terracotta-core version

myronkscott:


anthonydahanne:
wrong commit message ? should rather be "Adapt to tc core, tc api, config, tc passthrough upgrades"

myronkscott:
> wrong commit message ? should rather be "Adapt to tc core, tc api, config, tc passthrough upgrades"

updated

PR #531, opened 2019-05-24
Faster encode/decode of strings for runnel.

cschanck:
This steals some code from interal projects (StringTool) and uses it for
faster encode/decode of strings for runnel.

Note that the testing was only done locally, ands needs validation.

And StringTool should, ideally, be in some seperate repo that
both internal and external projects could depend on.

### DIFF
--- a/runnel/src/main/java/org/terracotta/runnel/utils/StringTool.java
+++ b/runnel/src/main/java/org/terracotta/runnel/utils/StringTool.java
@@ -1,0 +1,554 @@
+/*
+ * Copyright Terracotta, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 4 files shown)


runnel/src/main/java/org/terracotta/runnel/utils/StringTool.java
```diff
@@ -0,0 +1,554 @@
...
+    int start = binary.position();
+    boolean utfSeen = false;
+    int passes = Math.min(binary.remaining() / Long.BYTES, probeSize / Long.BYTES);
+    if (passes > 0) {
+      for (int i = start; !utfSeen && passes-- > 0; i+= Long.BYTES) {
+        long val = binary.getLong(i);
+        utfSeen = (val & 0x8080808080808080L) != 0L;
+      }
+    }
+    if (!utfSeen) {
```
cschanck: Dubious about this, but it does show a small perf increase when trapping UTF chars, I assume because of the fewer ByteBuffer calls?

myronkscott: I'd leave it out if it is only a small win.

myronkscott:
We need a way to flip between old and new so we have backwards compatibility.  How should we do this?

myronkscott:
Or is this way of encoding backwards compatible?

runnel/src/main/java/org/terracotta/runnel/utils/StringTool.java
cschanck: Removed it.

PR #532, opened 2019-05-29
Remove unnecessary shutdown hook in service provider

mathieucarbou:


management/monitoring-service/src/main/java/org/terracotta/management/service/monitoring/MonitoringServiceProvider.java
```diff
@@ -65,18 +64,13 @@
   private DefaultManagementDataListener managementDataListener;
   private Collection<ManageableServerComponent> manageablePlugins;
 
-  public MonitoringServiceProvider() {
-    // because only passthrough is calling close(), not tc-core, so this is to cleanly close services (thread pools) at shutdown
-    Runtime.getRuntime().addShutdownHook(new Thread(this::close));
```
anthonydahanne: so what will be calling close() now ?

mathieucarbou: see @myronkscott explanation in the chat

mathieucarbou:
> Rebase and I can merge. Thanks!

@myronkscott : I think it is already...

PR #539, opened 2019-06-25
 Fixing tests and dependencies to be compatible with java11.

mobasherul:


pom.xml
```diff
@@ -129,6 +129,11 @@
         <artifactId>jackson-databind</artifactId>
         <version>${jackson-databind.version}</version>
       </dependency>
+      <dependency>
+        <groupId>org.glassfish.jaxb</groupId>
```
cljohnso: We need to ensure we note this dependency addition for downstream impact.

cljohnso:
@chrisdennis Can you please assure that the use of the Glassfish JAXB dependency is consistent with Ehcache3?

pom.xml
mobasherul: same version 2.3.1of glassfish dependency is used in ehcache3 as well.

lease/entity-server/src/test/java/org/terracotta/lease/LeaseAcquirerServerServiceTest.java
```diff
@@ -72,6 +76,7 @@ public void codec() {
   }
 
   @Test
+  @Ignore
```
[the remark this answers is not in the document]
mobasherul: Need to fix this it is passing with java8 but failing with java11

management/sequence-generator/pom.xml
```diff
@@ -71,6 +71,10 @@
       <artifactId>junit</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+    <groupId>org.glassfish.jaxb</groupId>
+    <artifactId>jaxb-runtime</artifactId>
```
chrisdennis: This shouldn't be needed. The management code shouldn't be depending on a JAXB runtime. See my PR #545

PR #540, opened 2019-06-25
client versions in cluster state dumps.

GaryWKeim:


management/monitoring-service/src/main/java/org/terracotta/management/service/monitoring/MonitoringServiceProvider.java
```diff
@@ -86,8 +86,16 @@ public boolean initialize(ServiceProviderConfiguration configuration, PlatformCo
...
+
+    if (topologyService == null) {
+      dump.addState("status", "The TopologyService was not loaded.");
+      return;
+    }
+
+    if (topologyService.isCurrentServerActive()) {
       dump.addState("cluster", topologyService.getClusterCopy().toMap());
+    } else {
+      dump.addState("status", "This is a passive server, so it is unaware of connected clients and other stripe members.");
```
anthonydahanne: not version related, but nice to have, thank you

management/monitoring-service/src/main/java/org/terracotta/management/service/monitoring/TopologyService.java
```diff
@@ -250,6 +251,17 @@ public synchronized void clientConnected(PlatformServer currentActive, PlatformC
...
 
+          try {
+            Field f = platformConnectedClient.getClass().getDeclaredField("version");
+            f.setAccessible(true);
+            Object versionObj = f.get(platformConnectedClient);
+            if (versionObj != null) {
+              client.setVersion(versionObj.toString());
+            }
+          } catch (NoSuchFieldException | IllegalAccessException e) {
+            //
```
anthonydahanne: could be a trace log with "Client " + clientName + " does not have a version field set, probably because it's older

jhouserizer:
Is there a reason this was never merged?

GaryWKeim:
@myronkscott wanted to do it a different way. I don't know if he ever did it though.

PR #545, opened 2019-07-09
Issue 544

chrisdennis:


management/management-model/src/main/java/org/terracotta/management/model/cluster/ClientIdentifier.java
```diff
@@ -201,25 +199,6 @@ static String discoverHostName() {
     return hostname;
   }
 
-  static String generateNewUUID() {
```
mathieucarbou: why the removal of `generateNewUUID` ?

chrisdennis: It didn't look like anyone was using it.

mathieucarbou: oh yes you are right! 
Because now the UUID coming into the client identifier comes from the logical connection UUID.

myronkscott:
I will also note that something in this call chain asks for the MAC address of the host which in some environments can take a very long time (minutes)

PR #546, opened 2019-07-10
Clients are loosing management registry and tags after in case …

mathieucarbou:
…a timeout occurs causing entity to be flushed. Also improved management logging.


mathieucarbou:
> Are you sure changing this logging to info is not too spammy?

Yes, I have only change the logging of some important calls executed either when client starts, is closed, connects or reconnects. This will be less spammy than the log output from ClientChannelEventController, ClientMessageTransport, ClientHandshakeManagerImpl and DistributedObjectClient ;-)

PR #548, opened 2019-07-25
use JSONAssert to compare JSON data

lorban:


anthonydahanne:
why this change ? was there something wrong with the previous comparison ?

lorban:
Yes, comparing JSON as `String`s is prone to spurious failures when the expected JSON file's line endings or indentation changes.

anthonydahanne:
emergency merge : there's controversy about the choice of the library, but it's test only, so we can unblock @myronkscott

PR #549, opened 2019-07-25
Client versions

myronkscott:
Add Client version update to management registry
Add Client reported source address to management registry
Adapt tests


management/monitoring-service/src/main/java/org/terracotta/management/service/monitoring/Notification.java
```diff
@@ -38,5 +38,7 @@
   CLIENT_REGISTRY_AVAILABLE,
   CLIENT_TAGS_UPDATED,
 
-  ENTITY_REGISTRY_AVAILABLE
+  ENTITY_REGISTRY_AVAILABLE,
+  
+  CLIENT_META_UPDATED
```
mathieucarbou: could we have a notification tag for each use case (version and address) ?
Exactly like we did with `CLIENT_TAGS_UPDATED `.
It would be far easier for M&M to determine what has been the update

management/monitoring-service/src/main/java/org/terracotta/management/service/monitoring/TopologyService.java
```diff
@@ -258,6 +259,38 @@ public synchronized void clientConnected(PlatformServer currentActive, PlatformC
         });
   }
 
+  @Override
+  public void clientVersion(PlatformConnectedClient platformClient, String version) {
+    LOGGER.trace("{} client version({})", platformClient, version);
```
mathieucarbou: is it possible to keep a log like this:
```
LOGGER.trace("[0] clientVersion({}, {})", platformConnectedClient, version);
```
reason is that for debugging purposes to be able to trace the callers and consumer ids the first argument in bracket is the consumer ID doing the call.

management/monitoring-service/src/main/java/org/terracotta/management/service/monitoring/TopologyService.java
```diff
@@ -258,6 +259,38 @@ public synchronized void clientConnected(PlatformServer currentActive, PlatformC
...
+
+    stripe.getServerByName(currentActive.getServerName())
+        .ifPresent(server -> {
+
+          ClientIdentifier clientIdentifier = toClientIdentifier(platformClient);
+          cluster.getClient(clientIdentifier)
+              .ifPresent(client -> {
+                client.setVersion(version);
+                firingService.fireNotification(new ContextualNotification(client.getContext(), Notification.CLIENT_META_UPDATED.name()));
```
mathieucarbou: `CLIENT_VERSION_UPDATED`

management/monitoring-service/src/main/java/org/terracotta/management/service/monitoring/TopologyService.java
```diff
@@ -258,6 +259,38 @@ public synchronized void clientConnected(PlatformServer currentActive, PlatformC
...
+    LOGGER.trace("{} client address({})", platformClient, address);
+
+    stripe.getServerByName(currentActive.getServerName())
+        .ifPresent(server -> {
+
+          ClientIdentifier clientIdentifier = toClientIdentifier(platformClient);
+          cluster.getClient(clientIdentifier)
+              .ifPresent(client -> {
+                client.setClientAddress(address);
+                firingService.fireNotification(new ContextualNotification(client.getContext(), Notification.CLIENT_META_UPDATED.name()));
```
mathieucarbou: `CLIENT_ADDRESS_UPDATED` (like we have CLIENT_TAGS)

management/monitoring-service/src/main/java/org/terracotta/management/service/monitoring/TopologyService.java
```diff
@@ -258,6 +259,38 @@ public synchronized void clientConnected(PlatformServer currentActive, PlatformC
...
+              .ifPresent(client -> {
+                client.setVersion(version);
+                firingService.fireNotification(new ContextualNotification(client.getContext(), Notification.CLIENT_META_UPDATED.name()));
+              });
+        });
+  }
+
+  @Override
+  public void clientAddress(PlatformConnectedClient platformClient, String address) {
+    LOGGER.trace("{} client address({})", platformClient, address);
```
mathieucarbou: is it possible to keep a log like this:
```
LOGGER.trace("[0] clientAddress({}, {})", platformConnectedClient, address);
```

management/cluster-topology/src/main/java/org/terracotta/management/model/cluster/Client.java
```diff
@@ -43,6 +43,8 @@
   private final ClientIdentifier clientIdentifier;
   private final Collection<String> tags = new LinkedHashSet<>();
   private String hostName;
+  private String clientReportedAddress = "UNKNOWN";
+  private String version = "UNKNOWN";
```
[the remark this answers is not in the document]
myronkscott: I'm generally not a fan of Optional but I'm happy to accommodate this request

management/cluster-topology/src/main/java/org/terracotta/management/model/cluster/Client.java
```diff
@@ -80,6 +82,23 @@ public String getHostAddress() {
...
+    return version;
+  }
+  
+  public String getClientAddress() {
+    return clientReportedAddress;
+  }
+  
+  public void setClientAddress(String address) {
+    this.clientReportedAddress = address;
+  }
```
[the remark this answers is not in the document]
myronkscott: I will try try and come up with a more specific name.  Maybe getClientReportedAddress().  We should not use InetSocketAddress.  This is used for informational purposes only and is reported by the client to the server as a string.  Further, the client's network may have a very different setup than the server's (DNS, NAT, Cloud) so we should not try and impart too much information into this string.  It should be considered an informational tag for display only.

management/cluster-topology/src/main/java/org/terracotta/management/model/cluster/Client.java
```diff
@@ -243,7 +266,9 @@ public int hashCode() {
     map.put("vmId", getVmId());
     map.put("clientId", getClientId());
     map.put("hostName", getHostName());
-    map.put("tags", tags);
+    map.put("clientReportedAddress", this.clientReportedAddress != null ? this.clientReportedAddress : "UNKNOWN");
+    map.put("tags", getTags());
+    map.put("version", version != null ? version : "UNKNOWN");
```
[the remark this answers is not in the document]
myronkscott: seems reasonable.  Will do.

management/monitoring-service/src/main/java/org/terracotta/management/service/monitoring/Notification.java
myronkscott: I considered that but I think there is a point that we can get too specific.  From the backend standpoint, it would be nice to add more meta information and just have the TMC refresh.  If that is unreasonable, we can be more specific.

management/monitoring-service/src/main/java/org/terracotta/management/service/monitoring/TopologyService.java
myronkscott: ok.

management/monitoring-service/src/main/java/org/terracotta/management/service/monitoring/TopologyService.java
myronkscott: see above.  If you insist, we can do it this way but I prefer more generic refresh.

management/monitoring-service/src/main/java/org/terracotta/management/service/monitoring/TopologyService.java
myronkscott: ok.

management/monitoring-service/src/main/java/org/terracotta/management/service/monitoring/TopologyService.java
myronkscott: see above.  If you insist, we can do it this way but I prefer more generic refresh.

management/cluster-topology/src/main/java/org/terracotta/management/model/cluster/Client.java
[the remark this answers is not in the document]
mathieucarbou: it is not required but this is just to be consistent with what we use to have in management... And it forces the devs to think about the presence or absence of this field.

management/monitoring-service/src/main/java/org/terracotta/management/service/monitoring/Notification.java
mathieucarbou: My comment is based on my knowledge on how it is implemented and use in management.
We would at least need a way to know what has been updated.
So 2 choices:

1. either a notification type that is different for each kind of update => this is what we have right now and this is what is easier to support
2. either we create a more generic event but in this case:
    * we need to have the information (key,value) of what has been updated in the notification attributes
    * and of course we will have to threat that differently in management because we can have both specific events like tags and both generic events that requires to look inside the event object. So this is bad IMO because not consistent.

myronkscott: I think you make a good point.  Why don't we mimic what tags is doing and add generic "properties" with a string key and value.  Properties can only ever be added added and a CLIENT_PROPERTY_ADDED notification will be sent.  The properties map iteration will always be presented in the order added.

This most closely represents what we are trying to do with this concept.

management/monitoring-service/src/main/java/org/terracotta/management/service/monitoring/TopologyService.java
```diff
@@ -258,6 +259,38 @@ public synchronized void clientConnected(PlatformServer currentActive, PlatformC
         });
   }
 
+  @Override
+  public void clientVersion(PlatformConnectedClient platformClient, String version) {
```
mathieucarbou: do not forget `synchronized` ;-)

management/monitoring-service/src/main/java/org/terracotta/management/service/monitoring/TopologyService.java
```diff
@@ -258,6 +259,38 @@ public synchronized void clientConnected(PlatformServer currentActive, PlatformC
...
+          cluster.getClient(clientIdentifier)
+              .ifPresent(client -> {
+                client.setVersion(version);
+                firingService.fireNotification(new ContextualNotification(client.getContext(), Notification.CLIENT_META_UPDATED.name()));
+              });
+        });
+  }
+
+  @Override
+  public void clientAddress(PlatformConnectedClient platformClient, String address) {
```
mathieucarbou: do not forget `synchronized` ;-)

management/cluster-topology/src/test/resources/cluster.json
```diff
@@ -109,7 +109,9 @@
     "vmId" : "12345@127.0.0.1",
     "clientId" : "12345@127.0.0.1:ehcache:uid",
     "hostName" : null,
+    "clientReportedAddress" : "UNKNOWN",
```
anthonydahanne: will become null with @mathieucarbou comments no ?

management/cluster-topology/src/test/resources/cluster.json
```diff
@@ -109,7 +109,9 @@
...
     "tags" : [ ],
+    "version" : "UNKNOWN",
```
anthonydahanne: will become null with @mathieucarbou comments no ?

management/cluster-topology/src/main/java/org/terracotta/management/model/cluster/Client.java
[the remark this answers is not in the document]
myronkscott: moved to generic properties

management/cluster-topology/src/main/java/org/terracotta/management/model/cluster/Client.java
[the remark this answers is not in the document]
myronkscott: moved to generic properties

management/monitoring-service/src/main/java/org/terracotta/management/service/monitoring/Notification.java
myronkscott: modified.  please review generic model

myronkscott:
Moved version and clientReportedAddress to a generic properties map.  This is more inline with implementation and will accommodate additional properties in the future.  Client properties are only ever added, never removed or changed.

management/monitoring-service/src/main/java/org/terracotta/management/service/monitoring/TopologyService.java
```diff
@@ -258,6 +259,22 @@ public synchronized void clientConnected(PlatformServer currentActive, PlatformC
...
+    LOGGER.trace("[0] client property added ({}, key:{}, value:{})", platformClient, key, value);
+
+    stripe.getServerByName(currentActive.getServerName())
+        .ifPresent(server -> {
+
+          ClientIdentifier clientIdentifier = toClientIdentifier(platformClient);
+          cluster.getClient(clientIdentifier)
+              .ifPresent(client -> {
+                client.addProperty(key, value);
+                firingService.fireNotification(new ContextualNotification(client.getContext(), Notification.CLIENT_PROPERTY_ADDED.name()));
```
mathieucarbou: @myronkscott : could you use the third argument of the constructor to pass:
Collections.emptyMap(key, value) => to let M&M to know the context of this notification.

myronkscott: Do you mean singletonMap?   How about just the key?  Or do you really need the value as well?

PR #556, opened 2019-10-16
Added support for new offheap-resource addition in OffheapResourceProvider

saurabhagas:


offheap-resource/src/main/java/org/terracotta/offheapresource/OffHeapResourcesProvider.java
```diff
@@ -158,13 +136,66 @@ static BigInteger convert(BigInteger value, MemoryUnit unit) {
...
+          updateCurrentUsage(newCapacity - oldCapacity);
+          for (EntityManagementRegistry registry : registries) {
+            Map<String, String> attrs = new HashMap<>();
+            attrs.put("oldCapacity", Long.toString(oldCapacity));
+            attrs.put("newCapacity", Long.toString(newCapacity));
+            registry.pushServerEntityNotification(res.getManagementBinding(), "OFFHEAP_RESOURCE_CAPACITY_CHANGED", attrs);
+          }
+        });
+
+    resources.put(identifier, offHeapResource);
```
chaktatha: There may be subtle race here. We should use putIfAbsent. If putIfAbsent does not return null, return false.

offheap-resource/src/main/java/org/terracotta/offheapresource/OffHeapResourcesProvider.java
```diff
@@ -158,13 +136,66 @@ static BigInteger convert(BigInteger value, MemoryUnit unit) {
...
+        "allocatedMemory",
+        new HashSet<>(Arrays.asList("OffHeapResource", "tier")),
+        properties,
+        StatisticType.GAUGE,
+        () -> offHeapResource.capacity() - offHeapResource.available()
+    );
+    return true;
+  }
+
+  private void updateCurrentUsage(long delta) {
```
chaktatha: Considering we are only printing a warning message even when totalConfiguredOffheap > physicalMemory, this is acceptable.

offheap-resource/src/main/java/org/terracotta/offheapresource/OffHeapResourcesProvider.java
saurabhagas: Done

offheap-resource/src/test/java/org/terracotta/offheapresource/OffHeapResourcesProviderTest.java
```diff
@@ -58,64 +67,73 @@ public void testObserverExposed() {
...
+  }
+
+  @Test
+  public void testConcurrentOffheapAddition() throws Exception {
+    long perThreadIncrement = 100L;
+    int updatesPerThread = 50;
+    int numThreads = 20;
+    ExecutorService executorService = Executors.newFixedThreadPool(numThreads);
+    OffHeapResourcesProvider offHeapResourcesProvider = new OffHeapResourcesProvider(configuration);
+    executorService.execute(() -> {
```
ramsai1729: is this a concurrent test?

saurabhagas: Oops! Raised PR#557 to fix it. Thanks!

PR #572, opened 2020-03-03
Supports to compile with -Xlint:all -Werror in IDE

mathieucarbou:
@saurabhagas @chrisdennis @cschanck  : the reason I would like it merge is because the dynamic config code is clean relative to `-Xlint:all -Werror`. And the sad story is that the tc-platform code does not enforce and and the project has a lot of warnings....

I would like that we can at least continue working cleanly in dynamic config, so that implied being  able to easily spot the mistakes or warnings from our IDE at least.

Right now, there are too many, so if we introduce some more, it is hard to see.

I would like to enforce that with Maven if we all agree on that.


mathieucarbou:
@saurabhagas @chrisdennis @cschanck  : the reason I would like it merge is because the dynamic config code is clean relative to `-Xlint:all -Werror`. And the sad story is that the tc-platform code does not enforce and and the project has a lot of warnings....

I would like that we can at least continue working cleanly in dynamic config, so that implied being  able to easily spot the mistakes or warnings from our IDE at least.

Right now, there are too many, so if we introduce some more, it is hard to see.

I would like to enforce that with Maven if we all agree on that.

mathieucarbou:
@saurabhagas : rebased

PR #573, opened 2020-03-04
Fixing PhysicalMemory logging and access

mathieucarbou:
See the email thread sent to you guys.

Basically, the methods of this class are called a lot now, and it can happen that the methods are not accessible, or do a lot of things and lot of logging each time.

Example with:

```
openjdk version "1.8.0_242"
OpenJDK Runtime Environment (AdoptOpenJDK)(build 1.8.0_242-b08)
OpenJDK 64-Bit Server VM (AdoptOpenJDK)(build 25.242-b08, mixed mode)
```

We can have in TRACE mode more than 1300 lines of repeating exceptions like that:

```
2020-02-28 22:56:18,493 [main] TRACE org.terracotta.offheapresource.PhysicalMemory - Bean lookup failed on class java.lang.Object
java.lang.NoSuchMethodException: java.lang.Object.getTotalPhysicalMemorySize()
 at java.lang.Class.getMethod(Class.java:1786)
 at org.terracotta.offheapresource.PhysicalMemory.getAttribute(PhysicalMemory.java:57)
 at org.terracotta.offheapresource.PhysicalMemory.totalPhysicalMemory(PhysicalMemory.java:34)
 at org.terracotta.offheapresource.OffHeapResourcesProvider.warnIfOffheapExceedsPhysicalMemory(OffHeapResourcesProvider.java:206)
 at org.terracotta.offheapresource.OffHeapResourcesProvider.updateConfiguredOffheap(OffHeapResourcesProvider.java:197)
 at org.terracotta.offheapresource.OffHeapResourcesProvider.lambda$addToResources$3(OffHeapResourcesProvider.java:189)
 at java.util.concurrent.ConcurrentHashMap.computeIfAbsent(ConcurrentHashMap.java:1660)
 at org.terracotta.offheapresource.OffHeapResourcesProvider.addToResources(OffHeapResourcesProvider.java:152)
 at org.terracotta.offheapresource.OffHeapResourcesProvider.<init>(OffHeapResourcesProvider.java:63)
 at org.terracotta.offheapresource.OffHeapResourceConfigurationParser.parse(OffHeapResourceConfigurationParser.java:58)
 at org.terracotta.offheapresource.OffHeapResourceConfigurationParser.parse(OffHeapResourceConfigurationParser.java:41)
 at org.terracotta.config.TCConfigurationParser.parseStream(TCConfigurationParser.java:152)
 at org.terracotta.config.TCConfigurationParser.convert(TCConfigurationParser.java:245)
 at org.terracotta.config.TCConfigurationParser.parse(TCConfigurationParser.java:263)
 at org.terracotta.config.TCConfigurationParser.parse(TCConfigurationParser.java:259)
 at org.terracotta.dynamic_config.server.config_provider.DynamicConfigConfigurationProvider.createdTcConfiguration(DynamicConfigConfigurationProvider.java:182)
 at org.terracotta.dynamic_config.server.config_provider.DynamicConfigConfigurationProvider.initialize(DynamicConfigConfigurationProvider.java:98)
 at org.terracotta.dynamic_config.server.config_provider.SelectableConfigurationProvider.initialize(SelectableConfigurationProvider.java:26)
 at com.tc.server.TCServerMain.main(TCServerMain.java:65)
 at org.terracotta.dynamic_config.server.startup.StartupManager.startServer(StartupManager.java:181)
 at org.terracotta.dynamic_config.server.startup.StartupManager.startUsingConfigRepo(StartupManager.java:104)
 at org.terracotta.dynamic_config.server.startup.ConfigRepoStarter.startNode(ConfigRepoStarter.java:34)
 at org.terracotta.dynamic_config.server.startup.NodeProcessor.process(NodeProcessor.java:36)
 at org.terracotta.dynamic_config.server.startup.Options.process(Options.java:163)
 at org.terracotta.dynamic_config.server.TerracottaNode.main(TerracottaNode.java:20)
```

Also, with OpenJDK, the methods, even if being public and found, are not accessible if we do not use `method.setAccessible(true);`

So:
- I have reduced the logging
- I am finding the method another way and making it accessible (because on OpenJDK 8, it is not accessible)
- caching it so that the lookup is not done at each call
- and I do not log but throw in case of invocation error

### DIFF
--- a/offheap-resource/src/main/java/org/terracotta/offheapresource/PhysicalMemory.java
+++ b/offheap-resource/src/main/java/org/terracotta/offheapresource/PhysicalMemory.java
@@ -15,79 +15,68 @@
  */
 package org.terracotta.offheapresource;
 
-import java.lang.management.ManagementFactory;
-import java.lang.management.OperatingSystemMXBean;
-import java.lang.reflect.InvocationTargetException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.lang.management.ManagementFactory;
+import java.lang.management.OperatingSystemMXBean;
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
+
 /**
- *
  * @author cdennis
  */
 class PhysicalMemory {
 
   private static final Logger LOGGER = LoggerFactory.getLogger(PhysicalMemory.class);
-  private static final OperatingSystemMXBean OS_BEAN = ManagementFactory.getOperatingSystemMXBean();
+  private static final Method getTotalPhysicalMemorySize = find("getTotalPhysicalMemorySize");
+  private static final Method getFreePhysicalMemorySize = find("getFreePhysicalMemorySize");
+  private static final Method getTotalSwapSpaceSize = find("getTotalSwapSpaceSize");
+  private static final Method getFreeSwapSpaceSize = find("getFreeSwapSpaceSize");
+  private static final Method getCommittedVirtualMemorySize = find("getCommittedVirtualMemorySize");
 
   public static Long totalPhysicalMemory() {
-    return getAttribute("getTotalPhysicalMemorySize");
+    return invoke(getTotalPhysicalMemorySize);
   }
 
   public static Long freePhysicalMemory() {
-    return getAttribute("getFreePhysicalMemorySize");
+    return invoke(getFreePhysicalMemorySize);
   }
 
   public static Long totalSwapSpace() {
-    return getAttribute("getTotalSwapSpaceSize");
+    return invoke(getTotalSwapSpaceSize);
   }
 
   public static Long freeSwapSpace() {
-    return getAttribute("getFreeSwapSpaceSize");
+    return invoke(getFreeSwapSpaceSize);
   }
 
   public static Long ourCommittedVirtualMemory() {
-    return getAttribute("getCommittedVirtualMemorySize");
+    return invoke(getCommittedVirtualMemorySize);
   }
 
-  @SuppressWarnings("unchecked")
-  private static <T> T getAttribute(String name) {
-    LOGGER.trace("Bean lookup for {}", name);
-    for (Class<?> s = OS_BEAN.getClass(); s != null; s = s.getSuperclass()) {
-      try {
-        T result = (T) s.getMethod(name).invoke(OS_BEAN);
-        LOGGER.trace("Bean lookup successful using {}, got {}", s, result);
-        return result;
-      } catch (SecurityException e) {
-        LOGGER.trace("Bean lookup failed on {}", s, e);
-      } catch (NoSuchMethodException e) {
-        LOGGER.trace("Bean lookup failed on {}", s, e);
-      } catch (IllegalAccessException e) {
-        LOGGER.trace("Bean lookup failed on {}", s, e);
-      } catch (IllegalArgumentException e) {
-        LOGGER.trace("Bean lookup failed on {}", s, e);
-      } catch (InvocationTargetException e) {
-        LOGGER.trace("Bean lookup failed on {}", s, e);
+  private static Method find(String methodName) {
+    try {
+      Method method = ManagementFactory.getOperatingSystemMXBean().getClass().getMethod(methodName);
+      if (!method.isAccessible()) {
+        method.setAccessible(true);
       }
+      return method;
+    } catch (NoSuchMethodException | SecurityException e) {
+      LOGGER.trace("Unable to find or access method '{}' on the {}", methodName, OperatingSystemMXBean.class.getSimpleName());
+      return null;
     }
-    for (Class<?> i : OS_BEAN.getClass().getInterfaces()) {
-      try {
-        T result = (T) i.getMethod(name).invoke(OS_BEAN);
-        LOGGER.trace("Bean lookup successful using {}, got {}", i, result);
-        return result;
-      } catch (SecurityException e) {
-        LOGGER.trace("Bean lookup failed on {}", i, e);
-      } catch (NoSuchMethodException e) {
-        LOGGER.trace("Bean lookup failed on {}", i, e);
-      } catch (IllegalAccessException e) {
-        LOGGER.trace("Bean lookup failed on {}", i, e);
-      } catch (IllegalArgumentException e) {
-        LOGGER.trace("Bean lookup failed on {}", i, e);
-      } catch (InvocationTargetException e) {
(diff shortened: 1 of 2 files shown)


mathieucarbou:
@saurabhagas : rebased

saurabhagas:
@mathieucarbou small outstanding comment.

Would need another rebase now as the other PR got merged.

mathieucarbou:
@saurabhagas : rebased. I would like @chrisdennis to have a look quick because he wrote this class initially.

offheap-resource/src/main/java/org/terracotta/offheapresource/PhysicalMemory.java
```diff
@@ -15,79 +15,70 @@
...
-        LOGGER.trace("Bean lookup failed on {}", s, e);
-      } catch (IllegalAccessException e) {
-        LOGGER.trace("Bean lookup failed on {}", s, e);
-      } catch (IllegalArgumentException e) {
-        LOGGER.trace("Bean lookup failed on {}", s, e);
-      } catch (InvocationTargetException e) {
-        LOGGER.trace("Bean lookup failed on {}", s, e);
+  private static Method find(String methodName) {
+    try {
+      Method method = ManagementFactory.getOperatingSystemMXBean().getClass().getMethod(methodName);
```
chrisdennis: `getMethod(methodName)` can throw a SecurityException as well.

offheap-resource/src/main/java/org/terracotta/offheapresource/PhysicalMemory.java
```diff
@@ -15,79 +15,70 @@
...
+      if (!method.isAccessible()) {
+        method.setAccessible(true);
```
chrisdennis: This can throw a `SecurityException`. As as an aside are you sure this is necessary in Java 8? I can imagine it might be in Java 9 due to the increased encapsulation caused by modules... but I'd be surprised if it failed in Java 8.

offheap-resource/src/main/java/org/terracotta/offheapresource/PhysicalMemory.java
```diff
@@ -15,79 +15,70 @@
...
       }
+      return method;
+    } catch (NoSuchMethodException e) {
+      LOGGER.warn("Unable to find method '{}' on the {}", methodName, OperatingSystemMXBean.class.getSimpleName());
```
chrisdennis: Do we really want to elevate this to a warn? It was previously trace for good reason. There is nothing the customer can do to prevent the logging and it shouldn't affect the product functionality. Checking the allocations against the physical memory is a 'convenience' thing for the customer.

offheap-resource/src/main/java/org/terracotta/offheapresource/PhysicalMemory.java
```diff
@@ -15,79 +15,70 @@
...
+    try {
+      return method == null ? null : (Long) method.invoke(ManagementFactory.getOperatingSystemMXBean());
+    } catch (Throwable throwable) {
+      if (throwable instanceof Error) {
+        throw (Error) throwable;
+      }
+      if (throwable instanceof RuntimeException) {
+        throw (RuntimeException) throwable;
       }
+      throw new RuntimeException(throwable);
```
chrisdennis: Why are we throwing exceptions rather than just returning null here?

offheap-resource/src/test/java/org/terracotta/offheapresource/PhysicalMemoryTest.java
```diff
@@ -0,0 +1,35 @@
...
+package org.terracotta.offheapresource;
+
+import org.junit.Test;
+
+import static org.junit.Assert.assertNotNull;
+
+/**
+ * @author Mathieu Carbou
+ */
+public class PhysicalMemoryTest {
```
chrisdennis: The `PhysicalMemory` code is copied from `offheap-store`. If we're adding a test class here we might want to think about pulling over the test cases from the original too.

offheap-resource/src/test/java/org/terracotta/offheapresource/PhysicalMemoryTest.java
```diff
@@ -0,0 +1,35 @@
...
+  @Test
+  public void invocations_do_not_fail() {
+    assertNotNull(PhysicalMemory.totalPhysicalMemory());
```
chrisdennis: How confident are we that this will never fail on any test environment. The original class accepted that fetching these values was a best effort thing. Shouldn't we be tolerant of failures here?

offheap-resource/src/main/java/org/terracotta/offheapresource/PhysicalMemory.java
```diff
@@ -15,79 +15,70 @@
...
-      } catch (IllegalArgumentException e) {
-        LOGGER.trace("Bean lookup failed on {}", i, e);
-      } catch (InvocationTargetException e) {
-        LOGGER.trace("Bean lookup failed on {}", i, e);
+  }
+
+  private static Long invoke(Method method) {
+    try {
+      return method == null ? null : (Long) method.invoke(ManagementFactory.getOperatingSystemMXBean());
+    } catch (Throwable throwable) {
```
chrisdennis: What's wrong with multi-catch of the known expected types?

offheap-resource/src/test/java/org/terracotta/offheapresource/PhysicalMemoryTest.java
mathieucarbou: This test is only there to ensure that the method can be accessed from the JDK we are using at least. IF this test was there, we would have seen way before that it wasn't working for this OpenJDK 8 version. So you can see it as a way to catch this issue early now.

offheap-resource/src/test/java/org/terracotta/offheapresource/PhysicalMemoryTest.java
mathieucarbou: It is good to know if it fails... Perhaps we can catch some stupid issues like the one I caught on OpenJDK 8...

offheap-resource/src/main/java/org/terracotta/offheapresource/PhysicalMemory.java
mathieucarbou: oups!

mathieucarbou:
@chrisdennis : changed done. 

I do not catch IllegalArgumentException because like the other method calls that are not catching exceptions related to arguments, the code as it was before and as it is now rely on the fact that this is a programming error if such things occurs (IllegalArgumentException or NPE) because the wrong object is passed.

PR #581, opened 2020-03-06
Moving Lease ConfigChangeHandler to OSS

mathieucarbou:


lease/entity-server/src/main/java/org/terracotta/lease/LeaseAcquirerServerService.java
```diff
@@ -126,4 +133,16 @@ public int concurrencyKey(LeaseMessage leaseMessage) {
   public SyncMessageCodec<LeaseMessage> getSyncMessageCodec() {
     return null;
   }
+
+  private void wireChangeHandler(ServiceRegistry serviceRegistry) throws ConfigurationException {
+    try {
+      LeaseDuration leaseDuration = getService(serviceRegistry, new BasicServiceConfiguration<>(LeaseDuration.class));
+      ConfigChangeHandlerManager configChangeHandlerManager = serviceRegistry.getService(new BasicServiceConfiguration<>(ConfigChangeHandlerManager.class));
+      if (configChangeHandlerManager != null) {
```
GaryWKeim: Is it OK that this may return null and nothing be done about it?

lease/entity-server/src/main/java/org/terracotta/lease/LeaseAcquirerServerService.java
```diff
@@ -126,4 +133,16 @@ public int concurrencyKey(LeaseMessage leaseMessage) {
...
+        configChangeHandlerManager.compute(CLIENT_LEASE_DURATION, () -> new LeaseConfigChangeHandler(leaseDuration));
+      }
+    } catch (ServiceException e) {
+      throw new IllegalStateException("Failed to obtain Backup and Restore Service Instance. Ex: " + e);
```
GaryWKeim: This exception message doesn't seem relevant.

lease/entity-server/src/main/java/org/terracotta/lease/service/LeaseServiceImpl.java
```diff
@@ -20,18 +20,20 @@
 import org.terracotta.entity.ClientDescriptor;
 import org.terracotta.lease.service.monitor.LeaseState;
 
+import java.util.function.LongSupplier;
```
GaryWKeim: Is this needed?

lease/entity-server/src/main/java/org/terracotta/lease/service/LeaseServiceProviderImpl.java
```diff
@@ -56,7 +57,7 @@
   public boolean initialize(ServiceProviderConfiguration configuration, PlatformConfiguration platformConfiguration) {
     LOGGER.info("Initializing LeaseServiceProvider");
```
GaryWKeim: Seems like extraneous logging in this module.

lease/entity-server/src/main/java/org/terracotta/lease/LeaseAcquirerServerService.java
mathieucarbou: yes - if returns null, it means there is no dynamic config service on the server, the we do not register any config change handler to update the lease config

lease/entity-server/src/main/java/org/terracotta/lease/LeaseAcquirerServerService.java
mathieucarbou: Hum right... Copy pasted code from other repo. Will clean that. Need to throw a ConfigurationException instead.

lease/entity-server/src/main/java/org/terracotta/lease/service/LeaseServiceImpl.java
mathieucarbou: no - will clean thanks!

lease/entity-server/src/main/java/org/terracotta/lease/service/LeaseServiceProviderImpl.java
mathieucarbou: lol right! will remove. This lease service needs a good cleanup lol

lease/config/src/main/java/org/terracotta/lease/service/config/LeaseConfigurationImpl.java
```diff
@@ -15,30 +15,54 @@
...
+    this.leaseLength = leaseLength;
+  }
+
   @Override
   public Class<? extends ServiceProvider> getServiceProviderType() {
     return LeaseServiceProvider.class;
   }
+
+  private static long use(long leaseLength) {
+    LOGGER.info("Using lease length of " + leaseLength + "ms");
```
GaryWKeim: Could use a space prefixing the last token here.

lease/entity-server/src/main/java/org/terracotta/lease/service/LeaseServiceProviderImpl.java
GaryWKeim: If you think it's that important to note this in the log, then OK. It's just that we don't generally log things being init'd everywhere.

mathieucarbou: well, it's also that it's not my code and there's a lot of things I would change if we start changing how the logging is done / levels / etc / format / etc.
I would rather focus on  the changes I need to do for dynamic config.
but I agree that from what I see, the logging could be reviewed.

lease/config/src/main/java/org/terracotta/lease/service/config/LeaseConfigurationImpl.java
mathieucarbou: I have change that plus removed the second logger

PR #584, opened 2020-03-08
New kit with new core versions (plugins/api vs plugins/lib)

mathieucarbou:
__Progress:__

- [X] Upgraded core versions
- [X] Remove unnecessary `@CommonComponent`
- [X] Remove unnecessary `packaging-support` dependency
- [X] Split the dynamic config server in parts (config provider, config repository, conversion, services)
- [X] Provided a dist to package dynamic config server plugin
- [X] Fixed scopes of some modules in tc-platform
- [X] Packaging entities plus dependencies in their own jar, ;like its done for `default-configuration`
- [X] New platform kit
- [X] Refactoring to get rid of our main class and use only config provider only
- [X] Remove the XML file generation for the temporary startup when building a cluster
- [X] Reviewed creation:
    -  to only include terracotta classes and not external libs in shaded jars
    -  fixed scopes
    -  fixed configuration of jar plugin with OSGI bundle plugin
    - to generate Class-Path entries according to maven dependencies
- [X] Verified all generated jar content, maven inclusions + MANIFEST
- [X] Verifying KIT (exposing dependencies for default-configuration jar, etc)
- [X] Applying spotBugs over all project
- [X] Added `-Pslow` and `-Pfast` profiles, slow being the default (with spotBugs, license check, javadoc and sources)
- [X] config tool in the kit
- [x] Use the new Galvan version and new kit for the existing systems tests in project
- [x] PoC to remove XML config
- [x] Fix Galvan tests assertions (i.e. regarding entity count, et)
- [x] Resurrect dynamic config system tests and use the new KIT

```
# fast clean the project
./mvnw clean -T4C

# fast build of the project
./mvnw install -Pfast -DskipTests

# run all unit tests - no system tests
./mvnw test -DskipITs

# rebuild the kit and all modules depending on the kit
./mvnw clean install -DskipTests -Pfast -rf :platform-kit
```

Note: `-Pfast` means:
- No javadoc jar
- No source jar
- No spotBugs
- No license check

### DIFF
--- a/kit/pom.xml
+++ b/kit/pom.xml
@@ -1,0 +1,456 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+
+    Copyright Terracotta, Inc.
+
+    Licensed under the Apache License, Version 2.0 (the "License");
+    you may not use this file except in compliance with the License.
+    You may obtain a copy of the License at
+
+        http://www.apache.org/licenses/LICENSE-2.0
+
+    Unless required by applicable law or agreed to in writing, software
+    distributed under the License is distributed on an "AS IS" BASIS,
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+
+  <parent>
+    <groupId>org.terracotta</groupId>
+    <artifactId>platform-root</artifactId>
+    <version>5.7-SNAPSHOT</version>
+    <relativePath>../pom.xml</relativePath>
+  </parent>
+
+  <artifactId>platform-kit</artifactId>
+  <name>KIT :: Platform</name>
+  <packaging>jar</packaging>
+
+  <dependencies>
+
+    <!--
+      server/plugins/api
+    -->
+
+    <dependency>
+      <groupId>org.terracotta</groupId>
+      <artifactId>statistics</artifactId>
+      <version>${statistics.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>org.terracotta.management</groupId>
+      <artifactId>monitoring-service-api</artifactId>
+      <version>${project.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>org.terracotta.dynamic-config</groupId>
+      <artifactId>dynamic-config-api</artifactId>
+      <version>${project.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>org.terracotta.dynamic-config.server</groupId>
+      <artifactId>dynamic-config-server-api</artifactId>
+      <version>${project.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>org.terracotta.diagnostic</groupId>
+      <artifactId>diagnostic-service-api</artifactId>
+      <version>${project.version}</version>
+    </dependency>
+
+    <!--
+      server/plugins/lib
+    -->
+
+    <dependency>
+      <groupId>org.terracotta</groupId>
+      <artifactId>data-root-resource</artifactId>
+      <version>${project.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>org.terracotta</groupId>
+      <artifactId>offheap-resource</artifactId>
+      <version>${project.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>org.terracotta</groupId>
+      <artifactId>lease-entity-server</artifactId>
+      <version>${project.version}</version>
+    </dependency>
+    <!-- management -->
+    <dependency>
+      <groupId>org.terracotta.management</groupId>
+      <artifactId>monitoring-service</artifactId>
+      <version>${project.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>org.terracotta.management</groupId>
+      <artifactId>nms-agent-entity-server</artifactId>
+      <version>${project.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>org.terracotta.management</groupId>
+      <artifactId>nms-entity-server</artifactId>
+      <version>${project.version}</version>
+    </dependency>
+    <!-- dynamic config -->
+    <dependency>
+      <groupId>org.terracotta.dynamic-config.server</groupId>
+      <artifactId>dynamic-config-configuration-provider</artifactId>
+      <version>${project.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>org.terracotta.dynamic-config.server</groupId>
+      <artifactId>dynamic-config-services</artifactId>
+      <version>${project.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>org.terracotta.dynamic-config.entities</groupId>
+      <artifactId>dynamic-config-management-entity-server</artifactId>
+      <version>${project.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>org.terracotta.dynamic-config.entities</groupId>
+      <artifactId>dynamic-config-topology-entity-server</artifactId>
+      <version>${project.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>org.terracotta.dynamic-config.entities</groupId>
+      <artifactId>dynamic-config-nomad-entity-server</artifactId>
+      <version>${project.version}</version>
+    </dependency>
+
+    <!--
+      tools/config-tool/lib
+    -->
+
+    <dependency>
+      <groupId>org.terracotta.dynamic-config.cli</groupId>
+      <artifactId>dynamic-config-cli-config-tool</artifactId>
+      <version>${project.version}</version>
+    </dependency>
+
+    <!--
+      tools/config-convertor/lib
+    -->
+
+    <dependency>
+      <groupId>org.terracotta.dynamic-config.cli</groupId>
+      <artifactId>dynamic-config-cli-config-convertor</artifactId>
+      <version>${project.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>org.terracotta.dynamic-config.cli</groupId>
+      <artifactId>config-convertor-oss</artifactId>
+      <version>${project.version}</version>
+    </dependency>
+
+    <!--
+      KIT
+    -->
+
+    <dependency>
+      <groupId>org.terracotta.internal</groupId>
+      <artifactId>terracotta-kit</artifactId>
+      <version>${terracotta-core.version}</version>
+      <type>tar.gz</type>
+      <exclusions>
+        <exclusion>
+          <groupId>*</groupId>
+          <artifactId>*</artifactId>
+        </exclusion>
+      </exclusions>
+    </dependency>
+
+  </dependencies>
+
+  <build>
+    <plugins>
+      <plugin>
+        <artifactId>maven-jar-plugin</artifactId>
+        <executions>
+          <execution>
+            <id>default-jar</id>
+            <phase>none</phase>
+          </execution>
+        </executions>
+      </plugin>
+      <!-- open the KIT -->
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-dependency-plugin</artifactId>
+        <executions>
+          <execution>
+            <id>unpack-kit</id>
+            <phase>prepare-package</phase>
+            <goals>
+              <goal>unpack</goal>
+            </goals>
+            <configuration>
+              <outputDirectory>${project.build.directory}/terracotta-kit-${terracotta-core.version}</outputDirectory>
+              <artifactItems>
+                <artifactItem>
+                  <groupId>org.terracotta.internal</groupId>
+                  <artifactId>terracotta-kit</artifactId>
+                  <version>${terracotta-core.version}</version>
+                  <type>tar.gz</type>
+                </artifactItem>
+              </artifactItems>
+            </configuration>
+          </execution>
+        </executions>
+      </plugin>
+      <!-- generate plugins -->
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-shade-plugin</artifactId>
+        <configuration>
+          <shadedArtifactAttached>false</shadedArtifactAttached>
+          <createDependencyReducedPom>false</createDependencyReducedPom>
+          <createSourcesJar>false</createSourcesJar>
+          <transformers>
+            <transformer implementation="org.apache.maven.plugins.shade.resource.ServicesResourceTransformer"/>
+          </transformers>
+          <artifactSet>
+            <excludes>
+              <exclude>org.terracotta:platform-kit</exclude>
+            </excludes>
+          </artifactSet>
+        </configuration>
+        <executions>
+          <!-- api -->
+          <execution>
+            <id>terracotta-api-management</id>
+            <phase>prepare-package</phase>
+            <goals>
+              <goal>shade</goal>
+            </goals>
+            <configuration>
+              <outputDirectory>${project.build.directory}/packages/server/plugins/api</outputDirectory>
+              <finalName>terracotta-api-management-${project.version}</finalName>
+              <artifactSet>
+                <includes>
+                  <include>org.terracotta.management:monitoring-service-api</include>
+                  <include>org.terracotta.management:cluster-topology</include>
+                  <include>org.terracotta.management:management-model</include>
+                  <include>org.terracotta.management:management-registry</include>
+                  <include>org.terracotta.management:sequence-generator</include>
+                </includes>
+              </artifactSet>
+            </configuration>
+          </execution>
+          <execution>
+            <id>terracotta-api-diagnostic</id>
+            <phase>prepare-package</phase>
+            <goals>
+              <goal>shade</goal>
+            </goals>
+            <configuration>
+              <outputDirectory>${project.build.directory}/packages/server/plugins/api</outputDirectory>
+              <finalName>terracotta-api-diagnostic-${project.version}</finalName>
+              <artifactSet>
+                <includes>
+                  <include>org.terracotta.diagnostic:diagnostic-service-api</include>
+                </includes>
+              </artifactSet>
+            </configuration>
+          </execution>
+          <execution>
+            <id>terracotta-api-dynamic-config</id>
+            <phase>prepare-package</phase>
+            <goals>
+              <goal>shade</goal>
+            </goals>
+            <configuration>
+              <outputDirectory>${project.build.directory}/packages/server/plugins/api</outputDirectory>
+              <finalName>terracotta-api-dynamic-config-${project.version}</finalName>
+              <artifactSet>
+                <includes>
+                  <include>org.terracotta.dynamic-config:dynamic-config-api</include>
+                  <include>org.terracotta.dynamic-config.server:dynamic-config-server-api</include>
+                  <include>org.terracotta.common:common-inet-support</include>
+                  <include>org.terracotta.common:common-nomad</include>
+                  <include>org.terracotta.common:common-structures</include>
+                </includes>
+              </artifactSet>
+            </configuration>
+          </execution>
+          <!-- lib -->
+          <execution>
+            <id>terracotta-plugin-offheap</id>
+            <phase>prepare-package</phase>
+            <goals>
+              <goal>shade</goal>
+            </goals>
+            <configuration>
+              <outputDirectory>${project.build.directory}/packages/server/plugins/lib</outputDirectory>
+              <finalName>terracotta-plugin-offheap-${project.version}</finalName>
+              <artifactSet>
+                <includes>
+                  <include>org.terracotta:offheap-resource</include>
+                </includes>
+              </artifactSet>
+            </configuration>
+          </execution>
+          <execution>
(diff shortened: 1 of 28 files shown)


kit/src/main/assembly/vanilla-kit.xml
```diff
@@ -0,0 +1,162 @@
...
+    <fileSet>
+      <outputDirectory>.</outputDirectory>
+      <directory>src/main/kit</directory>
+      <fileMode>0755</fileMode>
+    </fileSet>
+  </fileSets>
+
+  <dependencySets>
+    <dependencySet>
+      <outputDirectory>server/plugins/api</outputDirectory>
```
saurabhagas: Will let others with knowledge of kit construction review this file.

kit/pom.xml
```diff
@@ -0,0 +1,213 @@
...
+  <modelVersion>4.0.0</modelVersion>
+
+  <parent>
+    <groupId>org.terracotta</groupId>
+    <artifactId>platform-root</artifactId>
+    <version>5.7-SNAPSHOT</version>
+    <relativePath>../pom.xml</relativePath>
+  </parent>
+
+  <artifactId>vanilla-kit</artifactId>
```
saurabhagas: The term "Vanilla kit" is fine as a cute description of what this kit is. Can we just call this module "kit" or "distribution"?

Also, please make sure that the name of the generated kit is "terracotta-5.x" (or something similar, as long as it's not vanilla.zip :-D).

kit/src/main/kit/tools/config-convertor/bin/config-convertor.bat
```diff
@@ -0,0 +1,43 @@
...
+@REM
+@REM Unless required by applicable law or agreed to in writing, software
+@REM distributed under the License is distributed on an "AS IS" BASIS,
+@REM WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+@REM See the License for the specific language governing permissions and
+@REM limitations under the License.
+@REM
+
+@echo off
+@REM See common/docs/src/main/asciidocs/windowsBatchCoding.adoc for coding guidance
```
saurabhagas: This comments needs to be removed. Applies to all `.bat` files.

kit/pom.xml
mathieucarbou: Not sure I like the ideea because this name is reserved for core kit I think. I'll ask @myronkscott @chrisdennis who knows better.

This kit is not meant __only to be used internally__ so we should name it __whatever we want and whatever name makes more sens to describe what it is__

kit/src/main/kit/tools/config-convertor/bin/config-convertor.bat
mathieucarbou: done at all places

kit/src/main/assembly/vanilla-kit.xml
mathieucarbou: @chrisdennis @myronkscott : please review this ASAP this is blocking us as long as it is not merged.

kit/pom.xml
```diff
@@ -0,0 +1,214 @@
...
+        <artifactId>maven-dependency-plugin</artifactId>
+        <executions>
+          <execution>
+            <id>unpack-kit</id>
+            <phase>prepare-package</phase>
+            <goals>
+              <goal>unpack</goal>
+            </goals>
+            <configuration>
+              <outputDirectory>${project.build.directory}/terracotta-kit-${terracotta-core.version}</outputDirectory>
```
chrisdennis: The tar.gz will contain a top level directory of this form already... why do this twice? At a minimum you could call this just `core-kit` or something compact like that.

kit/src/main/assembly/vanilla-kit.xml
```diff
@@ -0,0 +1,166 @@
...
+      </excludes>
+    </fileSet>
+    <fileSet>
+      <outputDirectory>.</outputDirectory>
+      <directory>src/main/kit</directory>
+      <fileMode>0755</fileMode>
+    </fileSet>
+  </fileSets>
+
+  <dependencySets>
```
chrisdennis: Could we not make this both shorter and more maintainable by automatically fetching the transtive dependencies?

kit/src/main/assembly/vanilla-kit.xml
```diff
@@ -0,0 +1,166 @@
...
+        <include>com.fasterxml.jackson.core:jackson-core</include>
+        <include>com.fasterxml.jackson.core:jackson-databind</include>
+        <include>com.fasterxml.jackson.datatype:jackson-datatype-jdk8</include>
+        <include>com.fasterxml.jackson.datatype:jackson-datatype-jsr310</include>
+        <include>com.beust:jcommander</include>
+        <include>org.terracotta:lease-config</include>
+        <include>ch.qos.logback:logback-classic</include>
+        <include>ch.qos.logback:logback-core</include>
+        <include>org.terracotta:offheap-resource</include>
+        <include>org.slf4j:slf4j-api</include>
```
chrisdennis: In the enterprise kit we have a `tools/common/lib` directory that contains all the logging related things and what used to be `:common:utilities` but I guess now gets translated in to something else.

kit/src/main/kit/tools/config-convertor/bin/config-convertor.sh
```diff
@@ -0,0 +1,35 @@
...
+
+# this will only happen if using sag installer
+if [ -r "${CONFIG_CONVERTOR_TOOL_DIR}/bin/setenv.sh" ] ; then
+  . "${CONFIG_CONVERTOR_TOOL_DIR}/bin/setenv.sh"
+fi
+
+java_opts="$JAVA_OPTS"
+
+if [ ! -d "$JAVA_HOME" ]; then
+   echo "ERROR: JAVA_HOME must point to Java installation."
```
chrisdennis: `point to a Java installation`

kit/src/main/kit/tools/config-tool/bin/config-tool.bat
```diff
@@ -0,0 +1,43 @@
...
+
+if exist "!CONFIG_TOOL_DIR!\bin\setenv.bat" (
+  pushd "!CONFIG_TOOL_DIR!\bin" && (
+    call .\setenv.bat
+    popd
+  )
+)
+
+if not defined JAVA_HOME (
+  echo Environment variable JAVA_HOME needs to be set
```
chrisdennis: Would be nice for the bat and bash scripts to be consistent about this error message.

lease/config/src/main/java/org/terracotta/lease/service/config/LeaseConfigurationImpl.java
```diff
@@ -56,9 +56,18 @@ public void setLeaseLength(long leaseLength) {
     this.leaseLength = leaseLength;
   }
 
+  @SuppressWarnings("unchecked")
   @Override
   public Class<? extends ServiceProvider> getServiceProviderType() {
-    return LeaseServiceProvider.class;
+    // this is a hack that allows us to split the config module from server entity code
```
chrisdennis: Could you not have the 'server-side' config object be a subclass of the real thing to avoid this hack?

lease/system-test/src/test/java/org/terracotta/lease/LeaseSystemTest.java
```diff
@@ -54,7 +54,7 @@ public static void beforeClass() {
             server -> {
               server.registerClientEntityService(new LeaseAcquirerClientService());
               server.registerServerEntityService(new LeaseAcquirerServerService());
-              server.registerOverrideServiceProvider(new LeaseServiceProviderImpl(), new LeaseConfigurationImpl(500L));
+              server.registerOverrideServiceProvider(new LeaseServiceProvider(), new LeaseConfigurationImpl(500L));
```
chrisdennis: Was this rename necessary - it seems to be causing quite a bit of noise?

lease/config/src/main/java/org/terracotta/lease/service/config/LeaseServiceProvider.java
```diff
@@ -1,22 +0,0 @@
...
- * distributed under the License is distributed on an "AS IS" BASIS,
- * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
- * See the License for the specific language governing permissions and
- * limitations under the License.
- */
-package org.terracotta.lease.service.config;
-
-import org.terracotta.entity.ServiceProvider;
-
-public interface LeaseServiceProvider extends ServiceProvider {
```
chrisdennis: What this always unecessary or has something changed?

kit/src/main/assembly/vanilla-kit.xml
```diff
@@ -0,0 +1,166 @@
...
+      </includes>
+      <excludes>
+        <exclude>server/bin/**</exclude>
+        <exclude>server/conf/**</exclude>
+      </excludes>
+    </fileSet>
+    <fileSet>
+      <outputDirectory>.</outputDirectory>
+      <directory>src/main/kit</directory>
+      <fileMode>0755</fileMode>
```
chrisdennis: Not great practice to set executable bits on everything... could we confine that to just the scripts?

kit/pom.xml
```diff
@@ -0,0 +1,214 @@
...
+
+    <!--
+      server/plugins/lib
+    -->
+
+    <dependency>
+      <groupId>org.terracotta</groupId>
+      <artifactId>runnel</artifactId>
+      <version>${project.version}</version>
+    </dependency>
```
chrisdennis: Shouldn't runnel be pulled transitively?

kit/pom.xml
chrisdennis: To be honest for now I'd just link its name to its origin and call it the platform kit.

dynamic-config/server/src/main/java/org/terracotta/dynamic_config/server/service/DynamicConfigServiceProvider.java
```diff
@@ -182,27 +182,25 @@ public void prepareForSynchronization() {
     // no-op
   }
 
-  // note: if any of the get() calls below fail, it means the servers have not been started with dynamic config code
-
   private IParameterSubstitutor getSubstitutor() {
-    return DiagnosticServices.findService(IParameterSubstitutor.class).get();
+    return DiagnosticServices.findService(IParameterSubstitutor.class).orElse(null);
```
chrisdennis: Assuming the `note:` above still applies wouldn't it make sense to just have these be:
```
findService(...).orElseThrow(() -> new IllegalStateException("The servers have not been started with dynamic config code")`
```

kit/pom.xml
mathieucarbou: Not true... Name is different:
image

kit/src/main/assembly/vanilla-kit.xml
mathieucarbou: no - I did try... And this is not working well. The transitive feature of assembly plugin is not working well when using filtering and some deps are not fetched.

This is extensive, but also a god way to control exactly what ends up there.

In the past we had issues where we allowed transitive deps and we ended up with a lot of junk in the lib/api folders.

kit/src/main/assembly/vanilla-kit.xml
mathieucarbou: - yes the common utilities are now elsewhere in separate libs
- I can create this common folder this is a good thing because there are a lot of common stuff!

kit/pom.xml
chrisdennis: Right - but couldn't this just be `core-kit/terracotta-5.7.0-pre16`?

kit/src/main/kit/tools/config-convertor/bin/config-convertor.sh
mathieucarbou: copied from core scripts ;-)

lease/config/src/main/java/org/terracotta/lease/service/config/LeaseServiceProvider.java
mathieucarbou: This is the revert of a change I did. See my comment in `LeaseConfigurationImpl`.
Sadly, lease is not properly done.
Its configuration should have been an ExtendedConfig, totally separated from the service implementation.

I have talked with @myronkscott  about that 2 times already. I don't really like the way the lease stuff is done because dynamic config needs to access the config classes to update them. Currently, the lease config is a service config. And I think it would have been better to have the lease config be an ExtendedConfig instead (plugin), and then, used in the lease service.

Because right now, we have:

1 - dynamic config xml parsing uses lease config (provided - to parse xml elements)
2 - lease server uses and depends dynamic config API to register a config handler when starting the lease entity, because there is no other way to access the lease config from platform configuration from our code
3-  lease server uses lease config

So I had to split the lease config and server to make these dependencies cleaner.

Ideally, what I would have liked is that:

1 - dynamic config xml parsing uses lease config ExtendedConfig (provided - to parse xml elements)
2 -  lease server uses lease config ExtendedConfig
3 - dynamic config code is directly capable of updating the lease config and register the hander because it is an ExtendedConfig

That means lease server would not depend on any dynamic config code.

So for now, to be able to separate a service config and a config, we have to do an ugly hack because the `ServiceProviderConfiguration` is tightly linked to the service implementation. An interface does not work.

kit/src/main/assembly/vanilla-kit.xml
mathieucarbou: 0700 ?

kit/src/main/assembly/vanilla-kit.xml
chrisdennis: Doesn't that imply something broken in the transitive deps then? I don't see why we'd want to 'control' what goes in here at this stage. What goes in here should (all things being correct) be something that can be calculated from the transitive dependency set?

kit/src/main/assembly/vanilla-kit.xml
mathieucarbou: my mistake, wasn't what I meant to do. will put 644 for everything except 755 for scripts

kit/src/main/assembly/platform-kit.xml
```diff
@@ -50,7 +50,7 @@
 
   <dependencySets>
     <dependencySet>
-      <outputDirectory>server/plugins/lib</outputDirectory>
+      <outputDirectory>server/plugins/api</outputDirectory>
```
chrisdennis: Anything going in to `plugins/api` shouldn't need `@CommonComponent` any more.

chrisdennis:
You might want to talk to @myronkscott about the need to put `Class-Path` entries in the manifests of the server entity/plugin lib jars.

chrisdennis:
@cljohnso is probably a good candidate to review the batch and bash scripts here since he's done so much work on them recently.

dynamic-config/server/src/main/java/org/terracotta/dynamic_config/server/service/DynamicConfigServiceProvider.java
mathieucarbou: The note was removed actually the original code is what you see now. 
I first tries to make dynamic config required (coinfig provider also) but it does not work because there are some cases in dynamic config where we start servers temporarly from a generated tc-config,xml file in diagnostic mode.

Also, all the dynamic config code now supports the fact that dynamic config might be "optional". iI..e if someone is calling the normal TcServerMain, we have a `SelectableConfigurationProvider` which will redirect either to the `DefaultConfigurationProvider` or our own `DynamicConfigConfigurationProvider`

kit/src/main/assembly/platform-kit.xml
mathieucarbou: you really want me do go over all the 98 usages ?

mathieucarbou:
> @cljohnso is probably a good candidate to review the batch and bash scripts here since he's done so much work on them recently.

I will add him. But we were based on the work he did in EE so we should have the right stuff.

kit/src/main/assembly/vanilla-kit.xml
mathieucarbou: No this issues with the assembly plugin and the ordering of the dependencySet when using filtering. The next ones seem to get the filters from the previous ones.
I have reviewed all the deps in the project to put the proper scopes. I could have missed something but not to the point to have such issues when using the plugin.

mathieucarbou: Also, overall, controlling the dependencies like that prevents any mistake in case a developper puts the wrong scope. I sincerely don't see any issue. Just more reliability and more stability.

kit/src/main/assembly/platform-kit.xml
chrisdennis: It should just be two search and replaces in each module in here, right?

dynamic-config/distribution/api/pom.xml
```diff
@@ -0,0 +1,67 @@
...
+      <artifactId>slf4j-api</artifactId>
+      <scope>provided</scope>
+    </dependency>
+  </dependencies>
+
+  <build>
+    <plugins>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-shade-plugin</artifactId>
```
chrisdennis: What do we gain by shading here?

dynamic-config/distribution/server/pom.xml
```diff
@@ -0,0 +1,108 @@
...
+      <artifactId>jackson-annotations</artifactId>
+      <scope>provided</scope>
+    </dependency>
+  </dependencies>
+
+  <build>
+    <plugins>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-shade-plugin</artifactId>
```
chrisdennis: What do we gain by shading here?

dynamic-config/distribution/xml/pom.xml
```diff
@@ -0,0 +1,88 @@
...
+      <artifactId>jackson-annotations</artifactId>
+      <scope>provided</scope>
+    </dependency>
+  </dependencies>
+
+  <build>
+    <plugins>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-shade-plugin</artifactId>
```
chrisdennis: What do we gain by shading here?

dynamic-config/server/configuration-provider/src/main/java/org/terracotta/dynamic_config/server/configuration/DynamicConfigConfigurationProvider.java
```diff
@@ -0,0 +1,224 @@
...
+  @Override
+  public void close() {
+    // Do nothing
+  }
+
+  private void withMyClassLoader(Runnable runnable) {
+    ClassLoader classLoader = getClass().getClassLoader();
+    ClassLoader oldloader = Thread.currentThread().getContextClassLoader();
+    try {
+      Thread.currentThread().setContextClassLoader(classLoader);
```
chrisdennis: This should probably be outside the try-block right?

dynamic-config/server/configuration-provider/src/main/java/org/terracotta/dynamic_config/server/configuration/DynamicConfigConfigurationProvider.java
```diff
@@ -0,0 +1,224 @@
...
+        ConfigRepositoryMapper.class, LicenseParser.class);
+  }
+
+  private static Options parseCommandLineOrExit(List<String> args) {
+    Options options = new Options();
+    CustomJCommander jCommander = new CustomJCommander("start-tc-server", options);
+    jCommander.parse(args.toArray(new String[0]));
+    if (options.isHelp()) {
+      jCommander.usage();
+      System.exit(0);
```
chrisdennis: Should we be calling `System.exit(...)` at this level... or should this be handled by the caller somehow?

mathieucarbou: its ok. This mimics what the TcServerMain class is also doing in its CLI parsing class. As I said, we need to do some CLI parsing and show a useful help to the user, not an exception.
If I throw, we need a core modification and a contract saying that throwing a specific exception type translates to catching it, hiding it (not show in any system out) and then exit.... W
=> Why bother ?

dynamic-config/distribution/server/pom.xml
mathieucarbou: see my other comment

dynamic-config/distribution/xml/pom.xml
mathieucarbou: see my other comment

mathieucarbou:
> Is there a list somewhere of what ends up in `api` versus what ends up in `lib`? And do we have some indication of how hard the remaining `api` things would be to move across?

=> `platform-kit.xml`

kit/src/main/assembly/platform-kit.xml
mathieucarbou: done

dynamic-config/distribution/api/pom.xml
mathieucarbou: Update:

What I do not like with this is the intermediary need to have modules only there to create these packages and handle properly the manifest with class-path entry.

I will try to see if there is a way to get rid of all the shading, and only do that in the kit module

mathieucarbou: Found a nice way to remove them 👍 Now, only kit and normal dependencies. NO more intermediary shaded libs

mathieucarbou:
@saurabhagas : this PR is based on #588 which can be merged. 

This PR only contains the KIT creation and its usage over all the project. The kit works, but here are the things left to do (I would need your help):

- Angela tests are not working because I need the new release for the script change
- MnM system tests are launching but there are some test failures because some assertions would need to be changed.

Note: if you see in the tests some classpath / classlaoder issues, do not try to fix them: they are most likely due to kit issues so let me know...

PR #585, opened 2020-03-08
[WIP] New Vanilla KIT with Galvan tests

mathieucarbou:
### This PR needs https://github.com/Terracotta-OSS/galvan/pull/209

This are the changes to use the new vanilla kit with Galvan in tc-platform.

@saurabhagas : This PR is failing right now as it need yours, but hopefully you could complete it to verify your changes in your PR.

PR #586, opened 2020-03-09
[WIP] New Vanilla KIT with New core versions (plugins/api vs plugins/lib)

mathieucarbou:
### Rebased on #587

@myronkscott @chrisdennis : this is the vanilla Kit using the new core versions with the new split.

__Issue 1:__

 I do not like the way the configuration-provider and default impl. are split: since we are still using the xml classes, I needed to replicate the whole code that you did in core. Bugs / maintenance will be at 2 places now.

Ideally, what I would like:  some classes of the `default-configuration` module without the META-INF/services descriptor.

__Issue 2:__

Dynamic config depends on XML parsing and lot of classes inside _configuration jar with all dependencies_... So we cannot just remove it to get rid of the meta inf service file. We just need the service descriptor to be removed.


dynamic-config/server/src/main/java/org/terracotta/dynamic_config/server/config_provider/DynamicConfigConfiguration.java
```diff
@@ -15,44 +15,181 @@
...
+
+  private static Server findServer(Servers servers, String serverName) throws ConfigurationException {
+    for (Server server : servers.getServer()) {
+      if (server.getName().equals(serverName)) {
+        return server;
+      }
+    }
+    throw new ConfigurationException("You have specified server name '" + serverName
+        + "' which does not exist in the specified configuration. \n\n"
+        + "Please check your settings and try again.");
```
mathieucarbou: @myronkscott : Code copy from core because I need some code from the wrapper class, but not the service file

dynamic-config/server/src/main/java/org/terracotta/dynamic_config/server/config_provider/ServerConfigurationImpl.java
```diff
@@ -0,0 +1,100 @@
...
+  @Override
+  public void setClientReconnectWindow(int value) {
+    this.clientReconnectWindow = value;
+  }
+
+  @Override
+  public File getLogsLocation() {
+    return new File(this.logs);
+  }
+}
```
mathieucarbou: @myronkscott : Code copy from core because I need some code from the wrapper class, but not the service file

management/testing/integration-tests/pom.xml
```diff
@@ -156,6 +145,29 @@
...
+                  <groupId>org.terracotta.management.testing</groupId>
+                  <artifactId>sample-entity</artifactId>
+                  <version>${project.version}</version>
+                  <classifier>plugin</classifier>
+                </artifactItem>
+                <artifactItem>
+                  <groupId>org.terracotta.management.dist</groupId>
+                  <artifactId>mnm-server</artifactId>
+                  <version>${project.version}</version>
+                </artifactItem>
```
mathieucarbou: @myronkscott : not using vanilla kit yet for galvan tests: need galvan release and #585 merges first.

mathieucarbou:
@chrisdennis @myronkscott : closing since I need to integrate these changes in #584 since we need that ASAP for the galvan integration (#585 )

PR #587, opened 2020-03-09
[WIP] Resurrecting Dynamic-Config system tests with Angela OSS

mathieucarbou:


saurabhagas:
@mathieucarbou when you say "this PR is pending some angela changes", do you mean to say that some changes in angela are pending, or some changes to make angela work in this PR are pending?

dynamic-config/testing/system-tests/pom.xml
```diff
@@ -0,0 +1,176 @@
...
+          </execution>
+        </executions>
+      </plugin>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-failsafe-plugin</artifactId>
+        <version>2.19.1</version>
+        <configuration>
+          <systemPropertyVariables>
+            <angela.rootDir>${project.build.directory}/angela</angela.rootDir>
```
saurabhagas: Several of these properties need to change based on the note I left last week:
1. instead of using angela.rootDir system property to point to the root kits directory, we need to use angela.kitInstallationDir to point to the extracted directory
2. angela.java.vendor can be set to empty string to allow any vendor
3. angela.java.version can be set to empty string to allow any java version
4. angela.kit.version needs to be renamed to angela.distribution

dynamic-config/testing/system-tests/pom.xml
```diff
@@ -0,0 +1,176 @@
...
+            <angela.skipUninstall>false</angela.skipUninstall>
+            <angela.tsa.fullLogging>true</angela.tsa.fullLogging>
+            <angela.kitVersion>${project.version}</angela.kitVersion>
+            <angela.kitInstallationDir>${project.build.directory}/vanilla-kit-${project.version}</angela.kitInstallationDir>
+            <angela.java.vendor/>
```
saurabhagas: I believe this would translate to the property being null (which is the same as not specifying this property). This property needs to be set to empty string to work correctly - the intention being that the user knows what he's doing. Value is set to the default if the value is null.

mathieucarbou: not true. debugged it. try yourslef ;) empty tag == empty string.

dynamic-config/testing/system-tests/pom.xml
mathieucarbou: 1 - these are not same same thing.

2 and 3: no, there is a bug in angela, see my PR. The validation is not skipped and empty tag means empty string. You can debug yourself ;-) I did.

4. Tried, NPE, because `angela.kitVersion` is REQUIRED and `angela.distribution` is used nowhere in angela.

Remember, this is not EE, but OSS... Perhaps you have made some changes in the wrong repo ?

saurabhagas: I'll fetch your PR locally and try it out.

saurabhagas: Added a commit here to correct an angela property and bump-up angela version (the version isn't released yet. Needs to be released once https://github.com/Terracotta-OSS/angela/pull/22 is merged).

Verified that the tests run locally after the changes. Maven was complaining about the new kit not being present on maven, so I manually copy-pasted the extracted kit under system-tests to run tests. This may need to be fixed if it's not a transient issue.

mathieucarbou: you cannot right click and run a system tests as-is with maven. You firs need to build the project, then launch the test. I will verify.

mathieucarbou:
@saurabhagas @aurbroszniowski : thank you I can run the tests now. Some are failing, I will have a look today.

dynamic-config/testing/system-tests/pom.xml
saurabhagas: I couldn't run tests from CLI either as the `mvn package -DskipTests` command failed in the root with failing to fetch the vanilla kit from maven.

mathieucarbou:
Closing: will be included in #584

PR #588, opened 2020-03-14
Dynamic config code refactoring to support new kit and new format

mathieucarbou:
I have extracted some of the changes from #584 here so that we can merge earlier.

This PR is still working with the old KIT version and this is only dynamic config code refactoring which does not impact kitting.

This PR also contains the changes to switch from XML to properties

Only the dynamic config modules are depending on the new core versions for compilation purposes.

Main POM still points to the old versions.


mathieucarbou:
@saurabhagas : you can start looking at the changes, but please do not merge yet this is still WIP and need to add some more changes and bug fixes.

We will be able to merge this PR before #584

mathieucarbou:
@saurabhagas : can you please merge this PR: contains the changes from #584 extracted (without the kit and new libraries).
Changes for dynamic config code so that it is compliant to the new kit.

Note: there are still some bugs to fix (not blocking):
- the config file we had never kept user inputs correctly in case the user set a config matching the default value. So we would use optional in our model to reflect the fact the the fields were configured or not.
- we should move all the cluster-wide params in the cluster object. This will simplify the model and validation and not attach-detach.

I will work on that while you focus on Angela and Galvan (see my comments in PR #584)

PR #592, opened 2020-03-17
Temporary fix to not save generated default in a file to keep as much as possible user input until TDB-4898 is fixed

mathieucarbou:


mathieucarbou:
@saurabhagas : can you merge this fix please ? It was in #584  - I have extracted it and made it better.

Basically, this allows to only save non default settings (settings that the user has modified that are not the default one) but still have when loading a topology or converting it, a full complete topology (in model or json) containing everything.

PR #595, opened 2020-03-19
Invoke platform mbean operations for node addition and removal

mobasherul:


dynamic-config/server/configuration-repository/pom.xml
```diff
@@ -52,5 +52,9 @@
       <version>${project.version}</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.terracotta.internal</groupId>
+      <artifactId>management</artifactId>
```
mathieucarbou: dependency should be in scope provided

PR #596, opened 2020-03-19
DetailedServerState MBean + cleanup

saurabhagas:


diagnostic/server/src/main/java/org/terracotta/diagnostic/server/DetailedServerState.java
```diff
@@ -0,0 +1,22 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.terracotta.diagnostic.server;
+
+import org.terracotta.diagnostic.common.LogicalServerState;
+
+public interface DetailedServerState {
```
mathieucarbou: the interface should not be in `server` module because this module is isolated (classlaoder).
Please move it to `server-api` so that entities can get it from the service provider

diagnostic/server/src/main/java/org/terracotta/diagnostic/server/DetailedServerStateProvider.java
```diff
@@ -0,0 +1,51 @@
...
+  @Override
+  public boolean initialize(ServiceProviderConfiguration serviceProviderConfiguration, PlatformConfiguration platformConfiguration) {
+    detailedServerState = DetailedServerStateImpl.init();
+    return true;
+  }
+
+  @Override
+  public <T> T getService(long l, ServiceConfiguration<T> serviceConfiguration) {
+    Class<T> serviceType = serviceConfiguration.getServiceType();
+    return !serviceType.equals(DetailedServerStateImpl.class) ? null : serviceType.cast(detailedServerState);
```
mathieucarbou: code should be: `rerturn  serviceType != DetailedServerState.class ? mull : ...`
(implementation is hidden, you should use the interface which needs to be in -api)

diagnostic/server/src/main/java/org/terracotta/diagnostic/server/DetailedServerStateProvider.java
```diff
@@ -0,0 +1,51 @@
...
+  }
+
+  @Override
+  public Collection<Class<?>> getProvidedServiceTypes() {
+    return Collections.singletonList(DetailedServerStateImpl.class);
```
mathieucarbou: use the interface here

diagnostic/server/pom.xml
```diff
@@ -0,0 +1,43 @@
...
+  </parent>
+  <modelVersion>4.0.0</modelVersion>
+
+  <artifactId>diagnostic-server</artifactId>
+  <dependencies>
+    <dependency>
+      <groupId>org.terracotta.diagnostic</groupId>
+      <artifactId>diagnostic-common</artifactId>
+      <version>${project.version}</version>
+      <scope>compile</scope>
```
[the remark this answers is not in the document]
mathieucarbou: You should not depend on common. common is for the json serialization handling.
Please move `LogicalServerState` class that you need in `server-api`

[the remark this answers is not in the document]
mathieucarbou: oh wait... more issue with that.. not psosible...

mathieucarbou:
Replaced by #597

PR #603, opened 2020-03-23
Fixes

saurabhagas:


dynamic-config/testing/system-tests/pom.xml
```diff
@@ -152,12 +152,11 @@
         <configuration>
           <systemPropertyVariables>
             <angela.rootDir>${project.build.directory}/angela</angela.rootDir>
-            <angela.skipUninstall>false</angela.skipUninstall>
+            <angela.skipUninstall>true</angela.skipUninstall>
```
mathieucarbou: @saurabhagas : Is the folder size acceptable now to be able to keep everything ?

saurabhagas: The biggest chunk was the kit, which we don't have anymore. Apart from that, it's the server-side logs, which we absolutely need.

Angela uploads everything specified in the pom to the client - meaning that whatever dependency you specify in the test pom will be seen in the client work dir. This can be problematic. Let's wait for a green build and see.

saurabhagas: Just realized that we don't have any client array specification. So no storage overhead from client-related libs!

PR #604, opened 2020-03-23
Fixing Nomad entity client-server com

mathieucarbou:
- client should also cache any exception that happened per stripe during the commit phase (not only the method return)
- server entity should log failures
- passive entity should not restart in case of failure because if the passive restarts when the active failed to commit and is still in PREPARE state, the passive won't be able to restart because the sync process will shutdown the passive.
- passive entity should ignore messages that could be received twice and lead to a DEAD or BAD state in Nomad for the subsequent call


mobasherul:
@mathieucarbou Currently if node addition commit fails on the active but passes in the existing passives then the overall attach command fails. But suppose the active went down and one of the passive becomes active and then the old active comes up and sync the configuration from the current active so that it gets the commit part of earlier attach that failed but the dynamic node didn't get added to the stripe.
So now all the existing nodes in the cluster will think that the dynamic node is part of cluster but it isn't. So is it appropriate to commit something that fails on active ?

PR #605, opened 2020-03-23
Fixing Nomad entity client-server com. and passive sync process

mathieucarbou:
- Client should also cache any exception that happened per stripe during the commit phase (not only the method return)
- Nomad Server entities should log failures
- Passive Nomad entity should not restart in case of duplicate message sent (DEAD message)
- Passive Nomad entity should restart and try to sync again in case of Nomad prepare/commit/rollback failure
- Sync process should not fail if active ends with PREPARE: this change should be also synced to passive nodes
- Warning log should be displayed after sync in case the server starts with a prepared change


mobasherul:
@mathieucarbou Currently if node addition commit fails on the active but passes in the existing passives then the overall attach command fails. But suppose the active went down and one of the passive becomes active and then the old active comes up and sync the configuration from the current active so that it gets the commit part of earlier attach that failed but the dynamic node didn't get added to the stripe.
So now all the existing nodes in the cluster will think that the dynamic node is part of cluster but it isn't. So is it appropriate to commit something that fails on active ?

PR #609, opened 2020-03-24
Added test to attach a passive to a cluster with a node that is down

mathieucarbou:


mathieucarbou:
Moved into #610

mathieucarbou:
Moved into #610

mathieucarbou:
Moved into #610

PR #611, opened 2020-03-25
Several fixes and improvements

mathieucarbou:
-  Typos
- Moving common libs in common api in kit
- Nomad entity communication should not have any timeout. Command should wait until completion.
- Better logging to warn user that the operation can take time (i.e. i failover is happening)
- Added information about who and from in toString()
- Do not support takeover in Nomad entity
- Added UpgradableNomadServer#getNomadChangeInfo(UUID)
- Fixing Nomad Server Entity to be tolerant in case the same message is received more than once
- Improved attach/detach/activate commands:
    - Better logging
    - Refactored activation (same code for attach and activate commands)
    - activate command now support -r flag to allow the activation of a node only with a given topology, which can be used to activate and restart a node that wasn't added to a cluster with the attach command if the later has failed during commit phase
    - detach command is always restarting the node to detach regardless if the Nomad Change was successful or not
    - attach command is not activating and not restarting the node to attach in case of a Nomad Change failure, and displays a warning to the user

@mobasherul : this PR contains the changes we have discussed plus a fix for the kit packaging.

@mobasherul @saurabhagas : I added a flag on the activate command (`-R`) which allows to restrict the activation to the node only.

__Example:__

```java
  @Test
  public void testRestrictedActivation() throws Exception {
    String config = copyConfigProperty("/config-property-files/single-stripe_multi-node.properties").toString();

    // import the cluster config to node 1 and node 2
    assertThat(configToolInvocation("import", "-f", config), is(successful()));

    // restrict activation to only node 1
    assertThat(configToolInvocation("activate", "-R", "-n", "my-cluster", "-s", "localhost:" + getNodePort(1, 1)),
        allOf(is(successful()), containsOutput("No license installed"), containsOutput("came back up")));

    waitUntil(out.getLog(1, 1), containsLog("Moved to State[ ACTIVE-COORDINATOR ]"));

    withTopologyService(1, 1, topologyService -> assertTrue(topologyService.isActivated()));
    withTopologyService(1, 2, topologyService -> assertFalse(topologyService.isActivated()));

    // restrict activation to only node 2, which will become passive as node 1
    assertThat(configToolInvocation("activate", "-R", "-n", "my-cluster", "-s", "localhost:" + getNodePort(1, 2)),
        allOf(is(successful()), containsOutput("No license installed"), containsOutput("came back up")));

    waitUntil(out.getLog(1, 2), containsLog("Moved to State[ PASSIVE-STANDBY ]"));

    withTopologyService(1, 1, topologyService -> assertTrue(topologyService.isActivated()));
    withTopologyService(1, 2, topologyService -> assertTrue(topologyService.isActivated()));

    assertThat(getRuntimeCluster(1, 1), is(equalTo(getRuntimeCluster(1, 2))));
  }
```

So it is helpful in the use cases:

1. A user would want to activate the nodes but one by one on a single stripe
2. The attach command fails in the Nomad commit process (which can be repaired) so we have a destination cluster that contains the information about the added node, but sadly, the node that we want to add has not been activated or restarted. So we can in this case export the configuration of the cluster, and run the activate command on the passive node to add with `-R` so that the node will be activated with the same topology as the destination cluster and will become a passive.

### DIFF
--- a/dynamic-config/cli/config-tool/src/main/java/org/terracotta/dynamic_config/cli/config_tool/nomad/NomadManager.java
+++ b/dynamic-config/cli/config-tool/src/main/java/org/terracotta/dynamic_config/cli/config_tool/nomad/NomadManager.java
@@ -207,6 +207,8 @@ public class NomadManager<T> {
         InetSocketAddress address = getAddress();
         int stripeId = destinationCluster.getStripeId(address).getAsInt();
         CompletableFuture<AcceptRejectResponse> result = cache.computeIfAbsent(stripeId, sid -> {
+          LOGGER.info("Committing passive change to stripe ID: {}... (this operation is blocking and can take time in case a failover happens)", stripeId);
+
           LOGGER.trace("Sending commit message: {} to stripe ID: {}", message, stripeId);
           CompletableFuture<AcceptRejectResponse> c = new CompletableFuture<>();
           try {
--- a/client-message-tracker/pom.xml
+++ b/client-message-tracker/pom.xml
@@ -33,11 +33,13 @@
       <groupId>org.terracotta</groupId>
       <artifactId>entity-server-api</artifactId>
       <version>${terracotta-apis.version}</version>
+      <scope>provided</scope>
     </dependency>
     <dependency>
       <groupId>org.terracotta</groupId>
       <artifactId>packaging-support</artifactId>
       <version>${terracotta-apis.version}</version>
+      <scope>provided</scope>
     </dependency>
 
     <dependency>
--- a/common/nomad/src/main/java/org/terracotta/nomad/messages/AcceptRejectResponse.java
+++ b/common/nomad/src/main/java/org/terracotta/nomad/messages/AcceptRejectResponse.java
@@ -78,6 +78,6 @@ public class AcceptRejectResponse {
 
   @Override
   public String toString() {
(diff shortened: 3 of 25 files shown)


dynamic-config/cli/config-tool/src/main/java/org/terracotta/dynamic_config/cli/config_tool/nomad/NomadManager.java
```diff
@@ -207,6 +207,8 @@ public AcceptRejectResponse commit(CommitMessage message) throws NomadException
         InetSocketAddress address = getAddress();
         int stripeId = destinationCluster.getStripeId(address).getAsInt();
         CompletableFuture<AcceptRejectResponse> result = cache.computeIfAbsent(stripeId, sid -> {
+          LOGGER.info("Committing passive change to stripe ID: {}... (this operation is blocking and can take time in case a failover happens)", stripeId);
```
mobasherul: Committing nomad change or passive change ?

mathieucarbou: should be commmitting toology change

mathieucarbou: @mobasherul : you can merge https://github.com/Terracotta-OSS/terracotta-platform/pull/612

PR #615, opened 2020-03-26
Several fixes

mathieucarbou:
- Rollback the nodes to attach to their initial topology
- Do not restart the node to detach if prepare is failing because no action is done
- Detached nodes are reseted and restart back in diagnostic mode with an empty topology


mathieucarbou:
> Needs a rebase

will do :-) I need to add a 3rd commit

mathieucarbou:
@mobasherul @saurabhagas : PR is good to merge :-)

PR #617, opened 2020-03-27
A/P fixes and tasks

mathieucarbou:
- Make repair command work in case some nodes are offline or not activated
- Removing an offline node

Note: removing an offline node works, but I have left the test ignored because the expected behavior would be that the detached node is able so start in diagnostic mode when detecting that it has been removed. It is not the case now it starts active...

@mobasherul @saurabhagas : you can stil merge, because the 2 commits are OK, we just need to determine what should be the right behavior to to a reset of the node. (for another commit).


saurabhagas:
Will let @mobasherul have a look and merge.

PR #624, opened 2020-04-02
Allow voter count to be zero to be consistent with terracotta.xsd

saurabhagas:
Discovered this problem during config conversion of a tc-config with a zero voter count. We need to be compatible with terracotta.xsd, which restricts voter count to be a non-negative integer.


mathieucarbou:
@saurabhagas : I will restart the build, it didn't start

dynamic-config/cli/cli-support/src/main/java/org/terracotta/dynamic_config/cli/command/RemoteMainCommand.java
```diff
@@ -27,6 +27,9 @@
 @Parameters(commandNames = LocalMainCommand.NAME)
 public class RemoteMainCommand extends LocalMainCommand {
 
+  @Parameter(names = {"-e", "--entity-timeout"}, description = "Entity operation timeout. Default: 120s", converter = TimeUnitConverter.class)
+  private Measure<TimeUnit> entityOperationTimeout = Measure.of(120, TimeUnit.SECONDS);
```
mathieucarbou, line 31: @mobasherul : use `"-t", "5s", "-e", "15s"` in your tests: connect timeout 5sec and entity operation timeout 15s. The command should end then.

management/testing/integration-tests/src/test/java/org/terracotta/management/integration/tests/DiagnosticIT.java
```diff
@@ -54,29 +57,14 @@ public void cluster_state_dump() throws Exception {
...
-      assertThat(dump, containsString("cacheSize="));
-
-      // MapProvider
-      assertThat(dump, containsString("caches="));
-
-      // Common on all active entities
-      assertThat(dump, containsString("instance="));
-      assertThat(dump, containsString("clientCount="));
-      assertThat(dump, containsString("clients="));
+      try (Stream<String> lines = Files.lines(Paths.get(getClass().getResource("/sate-dump-partial.txt").toURI()))) {
```
saurabhagas, line 62: The file isn't that big - we could do a string equals comparison instead of iterating over the lines

management/testing/integration-tests/src/test/resources/sate-dump-partial.txt
```diff
@@ -0,0 +1,217 @@
...
+    org.terracotta.entity.PlatformConfiguration={
+      serverName=testServer0
+      host=localhost
+      tsaPort=
+      ExtendedConfigs={
+        org.terracotta.dynamic_config.server.configuration.service.ConfigChangeHandlerManagerImpl={
+          configChangeHandlers={
+            NODE_LOG_DIR=org.terracotta.dynamic_config.server.service.handler.ServerAttributeConfigChangeHandler
+            NODE_GROUP_BIND_ADDRESS=org.terracotta.dynamic_config.server.service.handler.ServerAttributeConfigChangeHandler
+            NODE_PUBLIC_HOSTNAME=ConfigChangeHandler#accept()
```
saurabhagas, line 102: Can we dump these settings in lower case?

management/testing/integration-tests/src/test/java/org/terracotta/management/integration/tests/DiagnosticIT.java
mathieucarbou: no: lines are not in order and there can be some lines in the real dump that we skip.
I have only kept some important lines

saurabhagas: Okay

management/testing/integration-tests/src/test/resources/sate-dump-partial.txt
saurabhagas: Also wondering if it's worth to dump config change handler. It's static information, and not part of server "state"

PR #625, opened 2020-04-02
Add more flexibility for ConfigChangeHandlerManager and ConfigChangeHandler for override in case we need, plus better logging to troubleshoot

mathieucarbou:


mathieucarbou:
included in #624

PR #626, opened 2020-04-02
Tests for attach detach in consistency mode

mobasherul:


mathieucarbou:
@mobasherul : you can rebase on master

dynamic-config/testing/system-tests/src/test/java/org/terracotta/dynamic_config/system_tests/activated/AttachInConsistency1x3IT.java
```diff
@@ -0,0 +1,155 @@
...
+    startNode(1, 3);
+    waitUntil(out.getLog(1, 3), containsLog("Started the server in diagnostic mode"));
+    assertThat(getUpcomingCluster("localhost", getNodePort(1, 3)).getNodeCount(), is(equalTo(1)));
+
+    //setup for failover in commit phase on active
+    String propertySettingString = "stripe.1.node." + activeId + ".tc-properties.failoverAddition=killAddition-commit";
+    assertThat(configToolInvocation("set", "-s", "localhost:" + getNodePort(1, 1), "-c", propertySettingString), is(successful()));
+
+    // active died and passive can't become active 
+    assertThat(
```
mobasherul, line 136: This is tricky situation were the two phase commit failed but addition change getting committed on passive and when starting active it is becoming passive and syncing config from active. So Attach is getting successfully applied .

mathieucarbou: yeah - the node gets repaired" because it was partially committed the sync process is able to see that the active has the change committed, so it gets repaired automatically.

PR #631, opened 2020-04-04
Updated Azure descriptors and Maven POMs to execute tests with a Java 11 JVM

mathieucarbou:


azure-pipelines-windows.yml
```diff
@@ -26,5 +26,15 @@ resources:
...
+    timeoutInMinutes: 90
     vmImage: 'windows-latest'
+    options: '-B'
+- template: build-templates/maven-common.yml@templates
+  parameters:
+    jdkVersion: '1.8'
+    jobName: 'Java11'
     timeoutInMinutes: 90
+    vmImage: 'windows-latest'
+    options: '-B -DtestJVM=???'
```
mathieucarbou, line 40: @akomakom : 2 comments:
1. The Windows build does not seem to trigger. See previous PR build when I added the `-windows` yml (https://dev.azure.com/TerracottaCI/terracotta/_build/results?buildId=3679&view=results)
2. Maven has less flexibility than Gradle. Would it be possible in the init scripts to export some system properties per JVM ? I.e. here it would be nice to be able to do something like: `'-B -DtestJVM=%JAVA11_HOME%\bin\java.exe'`

azure-pipelines.yml
```diff
@@ -26,4 +26,15 @@ resources:
...
     timeoutInMinutes: 90
+    vmImage: 'ubuntu-latest'
+    options: '-B'
+- template: build-templates/maven-common.yml@templates
+  parameters:
+    jdkVersion: '1.8'
+    jobName: 'Java11'
+    timeoutInMinutes: 90
+    vmImage: 'ubuntu-latest'
+    options: '-B -DtestJVM=/usr/lib/jvm/zulu-11-azure-amd64/bin/java'
```
mathieucarbou, line 40: @akomakom : here is what I have to do now... Ideally this line would be something like: `'-B -DtestJVM=$JAVA11_HOME/bin/java'`

mathieucarbou:
@saurabhagas : the PR already passed several times and will pass again. I am just wondering if we just merge it as-is and not wait for the improvements (see my comments above).
@akomakom should be able to help us next week I think to improve the yml file content.
For now, considering the paths are not versioned, this solution would work in a reliable way for now.

azure-pipelines-windows.yml
mathieucarbou: I have opened issue #635 for the work to make the Win builds triggered. There are not right now, even if the yml file is there.

mathieucarbou: I have opened issue #636 for the work to put env variables to have access to the different JDK versions

azure-pipelines.yml
mathieucarbou: I have opened issue #636 for the work to put env variables to have access to the different JDK versions

akomakom: Like this? https://github.com/ehcache/ehcache3/blob/master/azure-pipelines-windows.yml#L39 ?

akomakom: @mathieucarbou In gradle, we baked this logic into the main build.gradle.  In Maven, what you're looking for is toolchains support in `surefire`, which it doesn't have.  However, [maven-forge-plugin](https://mvnrepository.com/artifact/org.terracotta/maven-forge-plugin/1.2.19) does.  It extends surefire.  I can give you internal usage examples offline.

PR #633, opened 2020-04-04
Correct galvan server log dir path and startup script path on Windows

saurabhagas:
**Depends on https://github.com/Terracotta-OSS/terracotta-core/pull/1128** and a corresponding version bump


mathieucarbou:
@saurabhagas : as per my comment in #631, my suggestion would be to merge #631 first and then rebase this PR to build it with Java11 build also.

saurabhagas:
@mathieucarbou rebased.

PR #637, opened 2020-04-04
Recovery / Repair

mathieucarbou:
- Changing --diagnostic-mode to --repair-mode to better match use cases and documentation. Diagnostic mode is also in conflict with the naming we use for diagnostic port
- Recovery / Repair features and fixes
- Allows to reset a node


mathieucarbou:
@saurabhagas : rebased :-)

PR #639, opened 2020-04-05
Moving cluster-wide settings to cluster level

mathieucarbou:

PR #640, opened 2020-04-06
Increasing build time limit, which is already taking ~1h15

mathieucarbou:


mathieucarbou:
@saurabhagas : if this one pass, please merge it instead of #639. It contains the commit from #639 plus the timeout increase. Merging #640 will close #639 automatically.

mathieucarbou:
@akomakom : rebased. 
Just the timeout change, can be merged directly without waiting for the build.

mathieucarbou:
declining in favor of #642

PR #642, opened 2020-04-06
maven-forge-plugin via parent for failsafe/surefire toolchains

akomakom:
Removed redundant items from pom, using parent to define forge plugin behavior (disables surefire/failsafe). 

```bash
* mvn verify # default (1.8, set in parent pom)
* mvn verify -Djava.test.version=1.11 
```
(relies on valid toolchains.xml)

Also adds conditional **argLine** for test jvm depending on java version (this is existing functionality but without hardcoded profiles)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -19,14 +19,18 @@
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
   <modelVersion>4.0.0</modelVersion>
 
-  <groupId>org.terracotta</groupId>
+  <parent>
+    <groupId>org.terracotta</groupId>
+    <artifactId>terracotta-parent</artifactId>
+    <version>5.10</version>
+  </parent>
+
   <artifactId>platform-root</artifactId>
   <version>5.7-SNAPSHOT</version>
   <packaging>pom</packaging>
 
   <properties>
     <maven.version>3.2.5</maven.version>
-    <maven-forge-plugin.version>1.2.16</maven-forge-plugin.version>
     <slf4j.version>1.7.25</slf4j.version>
     <logback.version>1.2.3</logback.version>
     <terracotta-apis.version>1.7.0-pre1</terracotta-apis.version>
@@ -40,8 +44,6 @@
     <jackson.version>2.10.1</jackson.version>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
-    <java.version>1.8</java.version>
-    <toolchain.provider.version>1.8</toolchain.provider.version>
   </properties>
 
   <modules>
@@ -261,53 +263,11 @@
         <plugin>
           <groupId>org.apache.maven.plugins</groupId>
           <artifactId>maven-compiler-plugin</artifactId>
-          <version>3.8.1</version>
           <configuration>
-            <source>${java.version}</source>
-            <target>${java.version}</target>
             <compilerArgs>
               <arg>-Xlint:all</arg>
               <arg>-Werror</arg>
             </compilerArgs>
-          </configuration>
-        </plugin>
-        <plugin>
-          <groupId>org.apache.maven.plugins</groupId>
-          <artifactId>maven-failsafe-plugin</artifactId>
-          <version>2.22.2</version>
-          <configuration>
-            <trimStackTrace>false</trimStackTrace>
-          </configuration>
-        </plugin>
-        <plugin>
-          <groupId>org.apache.maven.plugins</groupId>
-          <artifactId>maven-surefire-plugin</artifactId>
-          <version>2.22.2</version>
-          <configuration>
-            <trimStackTrace>false</trimStackTrace>
-          </configuration>
-        </plugin>
-        <plugin>
-          <groupId>org.sonatype.plugins</groupId>
-          <artifactId>nexus-staging-maven-plugin</artifactId>
-          <version>1.6.1</version>
-          <extensions>true</extensions>
-          <configuration>
-            <nexusUrl>http://nexus.terracotta.eur.ad.sag</nexusUrl>
-            <serverId>terracotta-nexus-staging</serverId>
-            <skipNexusStagingDeployMojo>${skipDeploy}</skipNexusStagingDeployMojo>
-          </configuration>
-        </plugin>
-        <plugin>
-          <groupId>org.apache.maven.plugins</groupId>
-          <artifactId>maven-toolchains-plugin</artifactId>
-          <version>3.0.0</version>
-          <configuration>
-            <toolchains>
-              <jdk>
-                <version>${toolchain.provider.version}</version>
-              </jdk>
-            </toolchains>
           </configuration>
         </plugin>
         <plugin>
@@ -341,22 +301,7 @@
           <artifactId>maven-jaxb2-plugin</artifactId>
           <version>0.14.0</version>
         </plugin>
-        <plugin>
-          <groupId>com.github.spotbugs</groupId>
-          <artifactId>spotbugs-maven-plugin</artifactId>
-          <version>3.1.12.2</version>
-          <configuration>
-            <xmlOutput>true</xmlOutput>
-          </configuration>
-          <dependencies>
-            <!-- overwrite dependency on spotbugs if you want to specify the version of spotbugs -->
-            <dependency>
-              <groupId>com.github.spotbugs</groupId>
-              <artifactId>spotbugs</artifactId>
-              <version>4.0.0</version>
-            </dependency>
-          </dependencies>
-        </plugin>
+
         <plugin>
           <artifactId>maven-assembly-plugin</artifactId>
           <version>3.2.0</version>
@@ -408,10 +353,6 @@
 
     <plugins>
       <plugin>
-        <groupId>org.apache.maven.plugins</groupId>
-        <artifactId>maven-compiler-plugin</artifactId>
-      </plugin>
-      <plugin>
         <groupId>org.apache.felix</groupId>
         <artifactId>maven-bundle-plugin</artifactId>
         <version>2.5.4</version>
@@ -447,33 +388,40 @@
   </build>
 
   <profiles>
-    <!-- fast profile for local use which will disable the default activated profiles and speed up the build -->
-    <profile>
-      <id>fast</id>
-    </profile>
     <!-- default profile including toolchain, javadoc, findbugs, license check, enforcer, etc -->
     <profile>
       <id>slow</id>
       <activation>
-        <activeByDefault>true</activeByDefault>
+        <property>
+          <name>fast</name>
+          <value>!true</value>
+        </property>
       </activation>
       <build>
+        <pluginManagement>
+          <plugins>
+            <plugin>
+              <groupId>com.github.spotbugs</groupId>
+              <artifactId>spotbugs-maven-plugin</artifactId>
+              <version>3.1.12.2</version>
+              <configuration>
+                <xmlOutput>true</xmlOutput>
+              </configuration>
+              <dependencies>
+                <!-- overwrite dependency on spotbugs if you want to specify the version of spotbugs -->
+                <dependency>
+                  <groupId>com.github.spotbugs</groupId>
+                  <artifactId>spotbugs</artifactId>
+                  <version>4.0.0</version>
+                </dependency>
+              </dependencies>
+            </plugin>
+          </plugins>
+        </pluginManagement>
         <plugins>
           <plugin>
             <groupId>org.apache.maven.plugins</groupId>
-            <artifactId>maven-toolchains-plugin</artifactId>
-            <executions>
-              <execution>
-                <goals>
-                  <goal>toolchain</goal>
-                </goals>
-              </execution>
-            </executions>
-          </plugin>
-          <plugin>
-            <groupId>org.apache.maven.plugins</groupId>
             <artifactId>maven-enforcer-plugin</artifactId>
-            <version>1.4.1</version>
             <executions>
               <execution>
                 <goals>
@@ -485,7 +433,7 @@
                       <uniqueVersions>true</uniqueVersions>
                     </dependencyConvergence>
                     <requireJavaVersion>
-                      <version>[${java.version},)</version>
+                      <version>[${java.build.version},)</version>
                     </requireJavaVersion>
                     <requireMavenVersion>
                       <version>[${maven.version},)</version>
@@ -522,11 +470,6 @@
             </executions>
           </plugin>
           <plugin>
-            <groupId>org.sonatype.plugins</groupId>
-            <artifactId>nexus-staging-maven-plugin</artifactId>
-            <extensions>true</extensions>
-          </plugin>
-          <plugin>
             <inherited>false</inherited>
             <groupId>com.mycila</groupId>
             <artifactId>license-maven-plugin</artifactId>
@@ -538,84 +481,53 @@
               </execution>
             </executions>
           </plugin>
-        </plugins>
-      </build>
-    </profile>
-    <!-- profile handling support for java versions >=9  -->
-    <profile>
-      <id>azure_linux_java_11</id>
-      <activation>
-        <property>
-          <name>azure_linux_java_11</name>
-        </property>
-      </activation>
-      <properties>
-        <toolchain.provider.version>1.11</toolchain.provider.version>
-      </properties>
-      <build>
-        <plugins>
           <plugin>
-            <groupId>org.apache.maven.plugins</groupId>
-            <artifactId>maven-failsafe-plugin</artifactId>
+            <groupId>org.terracotta</groupId>
+            <artifactId>maven-forge-plugin</artifactId>
+            <!--
+              configuration added to what is inherited from parent
+              See antrun plugin in this pom.
+             -->
             <configuration>
-              <trimStackTrace>false</trimStackTrace>
-              <argLine>--illegal-access=permit</argLine>
-              <jvm>/usr/lib/jvm/zulu-11-azure-amd64/bin/java</jvm>
-              <environmentVariables>
-                <JAVA_HOME>/usr/lib/jvm/zulu-11-azure-amd64</JAVA_HOME>
-              </environmentVariables>
+              <argLine>${java.test.argLine}</argLine>
             </configuration>
           </plugin>
           <plugin>
+            <!--
+              Workaround for maven not being able to set a property conditionally based on parent property.
+
+              When testing with Java 9+ we need to pass a flag to the jvm (illegal-access).
+              When testing with Java 8, we must not, since it's not supported.
+
+              Since $java.test.version is set in the parent, we can't easily key a profile off the value
+              of this variable (profile activation doesn't resolve maven variables).  Using antrun instead to set
+              the value of $java.test.argLine and referencing it in maven-forge-definition.
+            -->
             <groupId>org.apache.maven.plugins</groupId>
-            <artifactId>maven-surefire-plugin</artifactId>
-            <configuration>
-              <trimStackTrace>false</trimStackTrace>
-              <argLine>--illegal-access=permit</argLine>
-              <jvm>/usr/lib/jvm/zulu-11-azure-amd64/bin/java</jvm>
-              <environmentVariables>
-                <JAVA_HOME>/usr/lib/jvm/zulu-11-azure-amd64</JAVA_HOME>
-              </environmentVariables>
-            </configuration>
-          </plugin>
-        </plugins>
-      </build>
-    </profile>
-    <profile>
-      <id>azure_windows_java_11</id>
-      <activation>
-        <property>
-          <name>azure_windows_java_11</name>
-        </property>
-      </activation>
-      <properties>
-        <toolchain.provider.version>1.11</toolchain.provider.version>
-      </properties>
-      <build>
-        <plugins>
-          <plugin>
-            <groupId>org.apache.maven.plugins</groupId>
-            <artifactId>maven-failsafe-plugin</artifactId>
-            <configuration>
-              <trimStackTrace>false</trimStackTrace>
-              <argLine>--illegal-access=permit</argLine>
-              <jvm>C:\program files\java\zulu-11-azure-jdk_11.33.15-11.0.4-win_x64\bin\java.exe</jvm>
-              <environmentVariables>
-                <JAVA_HOME>C:\program files\java\zulu-11-azure-jdk_11.33.15-11.0.4-win_x64</JAVA_HOME>
-              </environmentVariables>
-            </configuration>
-          </plugin>
-          <plugin>
-            <groupId>org.apache.maven.plugins</groupId>
-            <artifactId>maven-surefire-plugin</artifactId>
-            <configuration>
-              <trimStackTrace>false</trimStackTrace>
-              <argLine>--illegal-access=permit</argLine>
-              <jvm>C:\program files\java\zulu-11-azure-jdk_11.33.15-11.0.4-win_x64\bin\java.exe</jvm>
-              <environmentVariables>
-                <JAVA_HOME>C:\program files\java\zulu-11-azure-jdk_11.33.15-11.0.4-win_x64</JAVA_HOME>
-              </environmentVariables>
-            </configuration>
+            <artifactId>maven-antrun-plugin</artifactId>
+            <version>1.8</version>
+            <executions>
+              <execution>
+                <phase>validate</phase>
+                <goals>
+                  <goal>run</goal>
+                </goals>
+                <configuration>
+                  <exportAntProperties>true</exportAntProperties>
+                  <target>
+                    <!--
+                      only add the jvm param if we're testing on newer JDKs
+                      Note: this property should not have a default in properties section, otherwise
+                      antrun may not be able to overwrite its value (or at least won't display it correctly)
+                    -->
+                    <condition property="java.test.argLine" value="" else="--illegal-access=permit">
+                      <equals arg1="${java.test.version}" arg2="1.8"/>
+                    </condition>
+                    <echo message="Configured test argLine (for java ${java.test.version}): '${java.test.argLine}'"/>
+                  </target>
+                </configuration>
+              </execution>
+            </executions>
           </plugin>
         </plugins>
       </build>
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -24,31 +24,25 @@ resources:
       name: terracotta/terracotta
 
 jobs:
-  - template: build-templates/maven-common.yml@templates
-    parameters:
-      jdkVersion: '1.8'
-      jobName: 'linux_java_8'
-      timeoutInMinutes: 120
-      vmImage: 'ubuntu-latest'
-      options: '-B'
-  - template: build-templates/maven-common.yml@templates
-    parameters:
-      jdkVersion: '1.8'
-      jobName: 'linux_java_11'
-      timeoutInMinutes: 120
-      vmImage: 'ubuntu-latest'
-      options: '-B -Dazure_linux_java_11'
-  - template: build-templates/maven-common.yml@templates
-    parameters:
-      jdkVersion: '1.8'
-      jobName: 'windows_java_8'
-      timeoutInMinutes: 120
-      vmImage: 'windows-latest'
-      options: '-B'
-  - template: build-templates/maven-common.yml@templates
-    parameters:
-      jdkVersion: '1.8'
-      jobName: 'windows_java_11'
-      timeoutInMinutes: 120
-      vmImage: 'windows-latest'
-      options: '-B -Dazure_windows_java_11'
+- template: build-templates/maven-common.yml@templates
+  parameters:
+    jobName: 'Linux_Java_8'
+    timeoutInMinutes: 120
+- template: build-templates/maven-common.yml@templates
+  parameters:
+    jobName: 'Linux_Java_11'
+    timeoutInMinutes: 120
+    options: '-B -Djava.test.version=1.11'
+- template: build-templates/maven-common.yml@templates
+  parameters:
+    jobName: 'Windows_Java_8'
+    timeoutInMinutes: 180
+    vmImage: 'windows-latest'
+- template: build-templates/maven-common.yml@templates
+  parameters:
+    jobName: 'Windows_Java_11'
+    timeoutInMinutes: 180
+    vmImage: 'windows-latest'
+    options: '-B -Djava.test.version=1.11'
+
+
--- a/dynamic-config/testing/system-tests/pom.xml
+++ b/dynamic-config/testing/system-tests/pom.xml
@@ -147,8 +147,8 @@
         </executions>
       </plugin>
       <plugin>
-        <groupId>org.apache.maven.plugins</groupId>
-        <artifactId>maven-failsafe-plugin</artifactId>
+        <groupId>org.terracotta</groupId>
+        <artifactId>maven-forge-plugin</artifactId>
         <configuration>
           <systemPropertyVariables>
             <angela.rootDir>${project.build.directory}/angela</angela.rootDir>
@@ -157,18 +157,10 @@
             <angela.distribution>${project.version}</angela.distribution>
             <angela.kitInstallationDir>${project.build.directory}/platform-kit-${project.version}</angela.kitInstallationDir>
             <angela.java.vendor/> <!--empty to allow any vendor-->
-            <angela.java.version>${toolchain.provider.version}</angela.java.version>
+            <angela.java.version>${java.test.version}</angela.java.version>
             <!--<angela.java.opts>-agentlib:jdwp=transport=dt_socket,server=y,suspend=y,address=5005</angela.java.opts>-->
           </systemPropertyVariables>
         </configuration>
-        <executions>
-          <execution>
-            <goals>
-              <goal>integration-test</goal>
-              <goal>verify</goal>
-            </goals>
-          </execution>
-        </executions>
       </plugin>
     </plugins>
   </build>
--- a/management/sequence-generator/pom.xml
+++ b/management/sequence-generator/pom.xml
@@ -32,8 +32,8 @@
   <build>
     <plugins>
       <plugin>
-        <groupId>org.apache.maven.plugins</groupId>
-        <artifactId>maven-surefire-plugin</artifactId>
+        <groupId>org.terracotta</groupId>
+        <artifactId>maven-forge-plugin</artifactId>
         <configuration>
           <argLine>
             -Dorg.terracotta.management.sequence.NodeIdSource=org.terracotta.management.sequence.MyNodeIdSource
--- a/management/testing/integration-tests/pom.xml
+++ b/management/testing/integration-tests/pom.xml
@@ -167,8 +167,8 @@
         </executions>
       </plugin>
       <plugin>
-        <groupId>org.apache.maven.plugins</groupId>
-        <artifactId>maven-failsafe-plugin</artifactId>
+        <groupId>org.terracotta</groupId>
+        <artifactId>maven-forge-plugin</artifactId>
         <configuration>
           <systemPropertyVariables>
             <JAVA_HOME>/Library/Java/JavaVirtualMachines/adoptopenjdk-11.jdk/Contents/Home/</JAVA_HOME>
@@ -176,14 +176,6 @@
 <!--            <serverDebugPortStart>5005</serverDebugPortStart>-->
           </systemPropertyVariables>
         </configuration>
-        <executions>
-          <execution>
-            <goals>
-              <goal>integration-test</goal>
-              <goal>verify</goal>
-            </goals>
-          </execution>
-        </executions>
       </plugin>
     </plugins>
   </build>


pom.xml
```diff
@@ -254,55 +256,13 @@
...
               <arg>-Werror</arg>
             </compilerArgs>
           </configuration>
         </plugin>
-        <plugin>
-          <groupId>org.apache.maven.plugins</groupId>
-          <artifactId>maven-failsafe-plugin</artifactId>
-          <version>2.22.2</version>
-          <configuration>
-            <trimStackTrace>false</trimStackTrace>
```
mathieucarbou, line 272: Could you add it to the parent pom or put it back in our pom ? This is really helpful to see the stack traces in the console.

pom.xml
```diff
@@ -254,55 +256,13 @@
...
-          </configuration>
-        </plugin>
-        <plugin>
-          <groupId>org.apache.maven.plugins</groupId>
-          <artifactId>maven-surefire-plugin</artifactId>
-          <version>2.22.2</version>
-          <configuration>
-            <trimStackTrace>false</trimStackTrace>
```
mathieucarbou, line 280: Could you add it to the parent pom or put it back in our pom ? This is really helpful to see the stack traces in the console.

mathieucarbou:
There seems to be an incompatibility between PortChooser of Galvan, Azure and Windows. So any progress on activating the Windows build is blocked until this is solved: https://github.com/Terracotta-OSS/galvan/issues/215

pom.xml
akomakom: I'll add it to the parent in 5.8.  Should show up in the maven repo in an hour at most.

mathieucarbou:
@akomakom :
`[ERROR] Unknown lifecycle phase "mx3072m". You must specify a valid lifecycle phase or a goal in the format <plugin-prefix>:<goal> or <plugin-group-id>:<plugin-artifact-id>[:<plugin-version>]:<goal>. Available lifecycle phases are: validate, initialize, generate-sources, process-sources, generate-resources, process-resources, compile, process-classes, generate-test-sources, process-test-sources, generate-test-resources, process-test-resources, test-compile, process-test-classes, test, prepare-package, package, pre-integration-test, integration-test, post-integration-test, verify, install, deploy, pre-clean, clean, post-clean, pre-site, site, post-site, site-deploy. -> [Help 1]
`

akomakom:
@mathieucarbou FYI I've added automatic support for the maven wrapper on the infra side, so the latest build is running that way.

akomakom:
@mathieucarbou Interestingly, the latest run only had one Junit3 error in Windows+Java11 (the other windows branch didn't fail)

mathieucarbou:
@akomakom : you can rebase your PR on top of #645 at the moment. I did some changes in our test to better support the slowness of Azure for WIndows. All builds pass now, several times consecutively.

mathieucarbou:
@saurabhagas : quick note on this PR to explain to you the goal.

At the moment, in master we have the profiles activated to make the windows build work. The only issue is that the paths to the JVM are hard-coded. So even it that works, this is not the ideal solution.

@akomakom has started this PR to integrate the forge plugin: this plugin is able to look at the JVM path from the toolchain file, so we just need to tell the JDK version we want to use. We saw some issues with the forge plugin - this is not the same pllugin versions than we are using in master and the parent pom has changed, so there are still things to be fixed.

@saurabhagas : we will be able to merge this PR when:

1. Build will pass and it will be rebased
2. We will have the same features we have right now which mean:
  - [ ] A fast build: when we do `./mvnw clean install -DskipTests -Pfast`, the `-Pfast` profile __should only build jars and source jars__. See the current profile for details: no toolchain (extra time), no verification, no license check, no javadoc, etc. 
  - [ ] Separation between  tests and IT tests: we should be able to run `./mvnw -DskipTests` to __skip all tests__ and `./mvnw -DskipITs` to only keep failsafe tests (Angela and Galvan). Also, this should be compatible with the `-Pfast` profile: i.e. a command I often run is: `./mvnw clean install -DskipITs -Pfast`.

Currently the work in this PR is not achieve to complete these 2 things we need to not impact the daily development in the project.

Also, the parent pom is currently problematic since it forces some plugin execution on downstream projects, which is wrong since it let no choice for downstream project to organize and speed up their build. Parent pom should only declare dependency management sections, plugin management section and some non-default profiles in theory and this is up to the downstream poms to organize their build depending on the speed they need.

Example: a small project with only 1-2 modules might not care about executing everything. BUt in tc-platform here, this is a pretty BIG project so we need to have control over the plugin execution. The parent pom cannot enforce them at every time.

akomakom:
Since #645 is merged I rebased onto master.  I removed `<argLine>--illegal-access=permit</argLine>` ... let me know if it's important and perhaps it can be added to the global parent.

mathieucarbou:
> Since #645 is merged I rebased onto master. I removed `<argLine>--illegal-access=permit</argLine>` ... let me know if it's important and perhaps it can be added to the global parent.

Yes this is important otherwise the tests will fail with Java 11 because there are some forbidden reflective calls in core sadly....

akomakom:
@mathieucarbou @saurabhagas I believe that the last build was correctly configured but tests [are failing](https://dev.azure.com/TerracottaCI/terracotta/_build/results?buildId=4108&view=ms.vss-test-web.build-test-results-tab).   Let me know if anything needs tweaking.  

I got some speedup from caching the local maven repo, but the build still takes a while.

mathieucarbou:
> @mathieucarbou @saurabhagas I believe that the last build was correctly configured but tests [are failing](https://dev.azure.com/TerracottaCI/terracotta/_build/results?buildId=4108&view=ms.vss-test-web.build-test-results-tab). Let me know if anything needs tweaking.
> 
> I got some speedup from caching the local maven repo, but the build still takes a while.

When you see _Error starting agent localhost_ this is caused by angela that is not able to restart the Ignite process because a previous test has timed out... This happens sometimes bit not frequently on Windows builds, so we restart the build. Let's try.

akomakom:
You're an optimist. 
At 1.5 hours per build and 4 of 10 total executors occupied for that entire time, this is not ideal.
If this cannot be made more robust then perhaps we should look into separating linux and windows into distinct jobs so that "Re-run failed checks" in GitHub will only re-run windows in this case.

 
> When you see _Error starting agent localhost_ this is caused by angela that is not able to restart the Ignite process because a previous test has timed out... This happens sometimes bit not frequently on Windows builds, so we restart the build. Let's try.

akomakom:
I take that back... it looks like Azure is able to re-run failed jobs as it is (has never worked for me before).  That's not as bad as I thought.

pom.xml
```diff
@@ -404,6 +366,14 @@
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-compiler-plugin</artifactId>
       </plugin>
+      <plugin>
+        <groupId>org.terracotta</groupId>
+        <artifactId>maven-forge-plugin</artifactId>
+        <!-- configuration added to what is inherited from parent -->
+        <configuration>
+          <argLine>${java.test.argLine}</argLine>
```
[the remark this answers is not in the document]
akomakom: I had initially [done exactly that](https://github.com/Terracotta-OSS/terracotta-platform/pull/642/commits/8b6dba0686d56d3bdca65d1979d41639615d7b7a) 

The problem with this approach is that profile activation needs to happen based on property value, and maven doesn't support maven properties in profile activation, only system properties (-D).   So, while it looked nicer, it made the assumption that any time someone passed a -Djava.test.version=ANYTHING, it meant that the profile should be active.   The other option was to say 
```xml
<value>!1.8</value>
```
but that would then mean that not specifying `-Djava.test.version=X` would mean that it's active.   So, no perfect way, which is why I switched to antrun.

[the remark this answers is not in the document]
mathieucarbou: But why would anyone use `-Djava.test.version=ANYTHING` ? We do not have this use case, even in Ehcache: the only place this property is passed would be on jenkins to change the build behavior to run with java 11. 

If someone locally wants to run the tests with Java 11, he can do it through the IDE, or run a maven build with java 8, then switch the environnement to point to 11, then run mvn test.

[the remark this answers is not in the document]
GaryWKeim: That's a PITA. The -D is better.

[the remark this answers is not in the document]
akomakom: > But why would anyone use `-Djava.test.version=ANYTHING` ? We do not have this use case, even in Ehcache: the only place this property is passed would be on ~jenkins~ _Azure_ to change the build behavior to run with java 11.
> 
> If someone locally wants to run the tests with Java 11, he can do it through the IDE, or run a maven build with java 8, then switch the environnement to point to 11, then run mvn test.

We are using `-Djava.test.version` in our entire in-house maven project set at this point (everything that extends the internal forge-parent).  Maven has a toolchain feature, so we use for CI.  If you don't want to use this in your IDE, I don't see why you need to.  How is this getting in your way?

The reason that this is not used in Ehcache is because ehcache is **not using maven**.  Still, in ehcache we use `-DtestVM=java11Home`, which is user-experience-equivalent to `-Djava.test.version=1.11` (if you see a difference, please explain).  One uses toolchains.xml, the other uses gradle.properties, both environment prerequisites.

akomakom:
> * I have tried a `./mvnw clean install -DskipTests -Pfast`: I think we could get rid of the executions of `buildnumber-maven-plugin` and `maven-antrun-plugin`. Both take some time and are not required for a local speedy build.

I don't believe that removing antrun (or whatever we choose to use for that) is possible. This build needs to be capable of running on JDK 8 and 11, and depending on which one we're using for tests, the command line has to change.  Buildnumber plugin can probably be removed (will that really speed anything up?)

> * I have tried a `./mvnw clean install -DskipITs -Pfast`: not working.  This command should execute only unit tests (*Test) but not the *IT tests.  This was the idea behind having a separation between surefire tests and failsafe tests before. Forge does not support IT tests it seems...

That is not correct. I added forge support when we last discussed this (a couple of weeks ago). It's [configured here](https://github.com/Terracotta-OSS/terracotta-parent/blob/master/pom.xml#L149).  What you're looking for is a separate property, which I can set up.


> 
> If we want the equivalent of before we would need an sort of forge plugin that extends failsafe. Looking at the source code I saw that:
> 
> ```
> /**
>  * Mimics the failsafe original but allows the JVM to be configured
>  * from toolchains
>  */
> @Mojo(name = "integration-test", requiresProject = true, requiresDependencyResolution = ResolutionScope.TEST,
>         defaultPhase = LifecyclePhase.INTEGRATION_TEST, threadSafe = true)
> public class TerracottaFailsafeIntegrationTestMojo extends IntegrationTestMojo {
> ```
> 
> So where we were using failsafe, probably we should use that instead of forge ?

We already are.

akomakom:
@mathieucarbou Here is a sample output with timestamps so you can determine whether it's worth stripping out extra plugins:

```ada
[2020-04-14 08:59:35] [INFO] ----------< org.terracotta.dynamic-config:dynamic-config-api >----------
[2020-04-14 08:59:35] [INFO] Building Dynamic Config :: API 5.7-SNAPSHOT                       [9/89]
[2020-04-14 08:59:35] [INFO] --------------------------------[ jar ]---------------------------------
[2020-04-14 08:59:35] [INFO] 
[2020-04-14 08:59:35] [INFO] --- maven-enforcer-plugin:3.0.0-M2:enforce (enforce-java) @ dynamic-config-api ---
[2020-04-14 08:59:35] [INFO] Skipping Rule Enforcement.
[2020-04-14 08:59:35] [INFO] 
[2020-04-14 08:59:35] [INFO] --- maven-toolchains-plugin:1.1:toolchain (default) @ dynamic-config-api ---
[2020-04-14 08:59:35] [INFO] Required toolchain: jdk [ version='1.8' ]
[2020-04-14 08:59:35] [INFO] Found matching toolchain for type jdk: JDK[/usr/lib/jvm/zulu-8-amd64]
[2020-04-14 08:59:35] [INFO] 
[2020-04-14 08:59:35] [INFO] --- maven-antrun-plugin:1.8:run (default) @ dynamic-config-api ---
[2020-04-14 08:59:35] [INFO] Executing tasks
[2020-04-14 08:59:35] 
[2020-04-14 08:59:35] main:
[2020-04-14 08:59:35]      [echo] Configured test argLine: ''
[2020-04-14 08:59:35] [INFO] Executed tasks
[2020-04-14 08:59:35] [INFO] 
[2020-04-14 08:59:35] [INFO] --- buildnumber-maven-plugin:1.3:create (default) @ dynamic-config-api ---
[2020-04-14 08:59:35] [INFO] Executing: /bin/sh -c cd /home/user/workspace/sag/terracotta-platform/dynamic-config/api && git rev-parse --verify HEAD
[2020-04-14 08:59:35] [INFO] Working directory: /home/user/workspace/sag/terracotta-platform/dynamic-config/api
[2020-04-14 08:59:35] [INFO] Storing buildNumber: 034046ef0228ef749e17531d12e16f886c03d10b at timestamp: 1586869175518
[2020-04-14 08:59:35] [INFO] Storing buildScmBranch: maven-forge-plugin
[2020-04-14 08:59:35] [INFO] 
[2020-04-14 08:59:35] [INFO] --- maven-resources-plugin:3.1.0:resources (default-resources) @ dynamic-config-api ---
[2020-04-14 08:59:35] [INFO] Using 'UTF-8' encoding to copy filtered resources.
[2020-04-14 08:59:35] [INFO] skip non existing resourceDirectory /home/user/workspace/sag/terracotta-platform/dynamic-config/api/src/main/resources
[2020-04-14 08:59:35] [INFO] 
[2020-04-14 08:59:35] [INFO] --- maven-compiler-plugin:3.3:compile (default-compile) @ dynamic-config-api ---
[2020-04-14 08:59:35] [INFO] Toolchain in maven-compiler-plugin: JDK[/usr/lib/jvm/zulu-8-amd64]
[2020-04-14 08:59:35] [INFO] Nothing to compile - all classes are up to date
[2020-04-14 08:59:35] [INFO] 
[2020-04-14 08:59:35] [INFO] >>> maven-bundle-plugin:2.5.4:manifest (default) > process-classes @ dynamic-config-api >>>
[2020-04-14 08:59:35] [INFO] 
[2020-04-14 08:59:35] [INFO] --- maven-enforcer-plugin:3.0.0-M2:enforce (enforce-java) @ dynamic-config-api ---
[2020-04-14 08:59:35] [INFO] Skipping Rule Enforcement.
[2020-04-14 08:59:35] [INFO] 
[2020-04-14 08:59:35] [INFO] --- maven-toolchains-plugin:1.1:toolchain (default) @ dynamic-config-api ---
[2020-04-14 08:59:35] [INFO] Required toolchain: jdk [ version='1.8' ]
[2020-04-14 08:59:35] [INFO] Found matching toolchain for type jdk: JDK[/usr/lib/jvm/zulu-8-amd64]
[2020-04-14 08:59:35] [INFO] 
[2020-04-14 08:59:35] [INFO] --- maven-antrun-plugin:1.8:run (default) @ dynamic-config-api ---
[2020-04-14 08:59:35] [INFO] Executing tasks
[2020-04-14 08:59:35] 
[2020-04-14 08:59:35] main:
[2020-04-14 08:59:35]      [echo] Configured test argLine: ''
[2020-04-14 08:59:35] [INFO] Executed tasks
[2020-04-14 08:59:35] [INFO] 
[2020-04-14 08:59:35] [INFO] --- buildnumber-maven-plugin:1.3:create (default) @ dynamic-config-api ---
[2020-04-14 08:59:35] [INFO] Executing: /bin/sh -c cd /home/user/workspace/sag/terracotta-platform/dynamic-config/api && git rev-parse --verify HEAD
[2020-04-14 08:59:35] [INFO] Working directory: /home/user/workspace/sag/terracotta-platform/dynamic-config/api
[2020-04-14 08:59:35] [INFO] Storing buildNumber: 034046ef0228ef749e17531d12e16f886c03d10b at timestamp: 1586869175546
[2020-04-14 08:59:35] [INFO] Storing buildScmBranch: maven-forge-plugin
[2020-04-14 08:59:35] [INFO] 
[2020-04-14 08:59:35] [INFO] --- maven-resources-plugin:3.1.0:resources (default-resources) @ dynamic-config-api ---
[2020-04-14 08:59:35] [INFO] Using 'UTF-8' encoding to copy filtered resources.
[2020-04-14 08:59:35] [INFO] skip non existing resourceDirectory /home/user/workspace/sag/terracotta-platform/dynamic-config/api/src/main/resources
[2020-04-14 08:59:35] [INFO] 
```

pom.xml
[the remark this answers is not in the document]
mathieucarbou: there is a BIG difference: 
- toolchain is a file outside of the project eventually maintained by a build and points to jd versions where keys are eventually similar across machines but path different.
- gradle.properties is a project file and it does contain project properties. The properties in this file NEVER change depending ion the env because this file is in githib. If the build is altering this file, this is a hack. Tehre are ways in gradle to look at config files outside of the project. 
- In maven, we can also have a maven.properties and have it loaded at build startup (http://www.mojohaus.org/properties-maven-plugin/). We just do not use this feature right now, but perhaps we should then to have the __exact__ gradle equivalent and not use dirty hacks in the pom.xml file

mathieucarbou:
> @mathieucarbou Here is a sample output with timestamps so you can determine whether it's worth stripping out extra plugins:
> 
> ```ada
> [2020-04-14 08:59:35] [INFO] ----------< org.terracotta.dynamic-config:dynamic-config-api >----------
> [2020-04-14 08:59:35] [INFO] Building Dynamic Config :: API 5.7-SNAPSHOT                       [9/89]
> [2020-04-14 08:59:35] [INFO] --------------------------------[ jar ]---------------------------------
> [2020-04-14 08:59:35] [INFO] 
> [2020-04-14 08:59:35] [INFO] --- maven-enforcer-plugin:3.0.0-M2:enforce (enforce-java) @ dynamic-config-api ---
> [2020-04-14 08:59:35] [INFO] Skipping Rule Enforcement.
> [2020-04-14 08:59:35] [INFO] 
> [2020-04-14 08:59:35] [INFO] --- maven-toolchains-plugin:1.1:toolchain (default) @ dynamic-config-api ---
> [2020-04-14 08:59:35] [INFO] Required toolchain: jdk [ version='1.8' ]
> [2020-04-14 08:59:35] [INFO] Found matching toolchain for type jdk: JDK[/usr/lib/jvm/zulu-8-amd64]
> [2020-04-14 08:59:35] [INFO] 
> [2020-04-14 08:59:35] [INFO] --- maven-antrun-plugin:1.8:run (default) @ dynamic-config-api ---
> [2020-04-14 08:59:35] [INFO] Executing tasks
> [2020-04-14 08:59:35] 
> [2020-04-14 08:59:35] main:
> [2020-04-14 08:59:35]      [echo] Configured test argLine: ''
> [2020-04-14 08:59:35] [INFO] Executed tasks
> [2020-04-14 08:59:35] [INFO] 
> [2020-04-14 08:59:35] [INFO] --- buildnumber-maven-plugin:1.3:create (default) @ dynamic-config-api ---
> [2020-04-14 08:59:35] [INFO] Executing: /bin/sh -c cd /home/user/workspace/sag/terracotta-platform/dynamic-config/api && git rev-parse --verify HEAD
> [2020-04-14 08:59:35] [INFO] Working directory: /home/user/workspace/sag/terracotta-platform/dynamic-config/api
> [2020-04-14 08:59:35] [INFO] Storing buildNumber: 034046ef0228ef749e17531d12e16f886c03d10b at timestamp: 1586869175518
> [2020-04-14 08:59:35] [INFO] Storing buildScmBranch: maven-forge-plugin
> [2020-04-14 08:59:35] [INFO] 
> [2020-04-14 08:59:35] [INFO] --- maven-resources-plugin:3.1.0:resources (default-resources) @ dynamic-config-api ---
> [2020-04-14 08:59:35] [INFO] Using 'UTF-8' encoding to copy filtered resources.
> [2020-04-14 08:59:35] [INFO] skip non existing resourceDirectory /home/user/workspace/sag/terracotta-platform/dynamic-config/api/src/main/resources
> [2020-04-14 08:59:35] [INFO] 
> [2020-04-14 08:59:35] [INFO] --- maven-compiler-plugin:3.3:compile (default-compile) @ dynamic-config-api ---
> [2020-04-14 08:59:35] [INFO] Toolchain in maven-compiler-plugin: JDK[/usr/lib/jvm/zulu-8-amd64]
> [2020-04-14 08:59:35] [INFO] Nothing to compile - all classes are up to date
> [2020-04-14 08:59:35] [INFO] 
> [2020-04-14 08:59:35] [INFO] >>> maven-bundle-plugin:2.5.4:manifest (default) > process-classes @ dynamic-config-api >>>
> [2020-04-14 08:59:35] [INFO] 
> [2020-04-14 08:59:35] [INFO] --- maven-enforcer-plugin:3.0.0-M2:enforce (enforce-java) @ dynamic-config-api ---
> [2020-04-14 08:59:35] [INFO] Skipping Rule Enforcement.
> [2020-04-14 08:59:35] [INFO] 
> [2020-04-14 08:59:35] [INFO] --- maven-toolchains-plugin:1.1:toolchain (default) @ dynamic-config-api ---
> [2020-04-14 08:59:35] [INFO] Required toolchain: jdk [ version='1.8' ]
> [2020-04-14 08:59:35] [INFO] Found matching toolchain for type jdk: JDK[/usr/lib/jvm/zulu-8-amd64]
> [2020-04-14 08:59:35] [INFO] 
> [2020-04-14 08:59:35] [INFO] --- maven-antrun-plugin:1.8:run (default) @ dynamic-config-api ---
> [2020-04-14 08:59:35] [INFO] Executing tasks
> [2020-04-14 08:59:35] 
> [2020-04-14 08:59:35] main:
> [2020-04-14 08:59:35]      [echo] Configured test argLine: ''
> [2020-04-14 08:59:35] [INFO] Executed tasks
> [2020-04-14 08:59:35] [INFO] 
> [2020-04-14 08:59:35] [INFO] --- buildnumber-maven-plugin:1.3:create (default) @ dynamic-config-api ---
> [2020-04-14 08:59:35] [INFO] Executing: /bin/sh -c cd /home/user/workspace/sag/terracotta-platform/dynamic-config/api && git rev-parse --verify HEAD
> [2020-04-14 08:59:35] [INFO] Working directory: /home/user/workspace/sag/terracotta-platform/dynamic-config/api
> [2020-04-14 08:59:35] [INFO] Storing buildNumber: 034046ef0228ef749e17531d12e16f886c03d10b at timestamp: 1586869175546
> [2020-04-14 08:59:35] [INFO] Storing buildScmBranch: maven-forge-plugin
> [2020-04-14 08:59:35] [INFO] 
> [2020-04-14 08:59:35] [INFO] --- maven-resources-plugin:3.1.0:resources (default-resources) @ dynamic-config-api ---
> [2020-04-14 08:59:35] [INFO] Using 'UTF-8' encoding to copy filtered resources.
> [2020-04-14 08:59:35] [INFO] skip non existing resourceDirectory /home/user/workspace/sag/terracotta-platform/dynamic-config/api/src/main/resources
> [2020-04-14 08:59:35] [INFO] 
> ```

Like I said, with the local -Pfast option, we don't want the build number plugin and the antrun plugin : they bring nothing except slow down the build. -Pfast: just compilation, jars, felix and source jars. Minimal things so that we can develop locally, run system tests, etc. Keep in mind in with maven, this is not like gradle taht is smart. When we want to re-tun the tests with Maven, we have to rebuild prety much nearly the whole project. That is why the mvn install -Pfast is really important for a big project because the devlpement time is highly impacted.

pom.xml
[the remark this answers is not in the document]
akomakom: You are thinking of a different gradle.properties file.  Observe:
```
/data/jenkins-slave/.gradle/gradle.properties
/data/jenkins-slave/.m2/toolchains.xml
```
All of our environments (including what you'd need to build ehcache3) already require these.  The only reason `testVM=java11Home` works is because the build host has the aforemenioned gradle.properties with this line:
```properties
java11Home=/jdk/zulu-1.11
```

I do not understand your reference to the maven plugin.  What problem are you trying to solve?  Do you want to reinvent toolchains?

akomakom:
> Like I said, with the local -Pfast option, we don't want the build number plugin and the antrun plugin : they bring nothing except slow down the build. -Pfast: just compilation, jars, felix and source jars. Minimal things so that we can develop locally, run system tests, etc. Keep in mind in with maven, this is not like gradle taht is smart. When we want to re-tun the tests with Maven, we have to rebuild prety much nearly the whole project. That is why the mvn install -Pfast is really important for a big project because the devlpement time is highly impacted.

Did you look at the timestamps?

mathieucarbou:
> > Like I said, with the local -Pfast option, we don't want the build number plugin and the antrun plugin : they bring nothing except slow down the build. -Pfast: just compilation, jars, felix and source jars. Minimal things so that we can develop locally, run system tests, etc. Keep in mind in with maven, this is not like gradle taht is smart. When we want to re-tun the tests with Maven, we have to rebuild prety much nearly the whole project. That is why the mvn install -Pfast is really important for a big project because the devlpement time is highly impacted.
> 
> Did you look at the timestamps?

Yes but it does not measure anything. In a big project, plugin executions are repeated. The correct way to measure the impact of a plugin would be to use something like this: https://blog.sebastian-daschner.com/entries/analyze_maven_build_time

in my computer, ` ./mvnw clean install -DskipTests -Pfast` takes several seconds in a total of ~100 modules to execute these plugins, which does not count the time to download and boostrap them each time by maven.

pom.xml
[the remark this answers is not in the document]
mathieucarbou: since you already have the logic in place to write a property, why not using that instead and get rid of the forge plugin. We do not need the forge plugin. You are already creating a property file with a mapping for the JVM to use. We just have to use these properties in surefire and failsafe in a profile then ?

Only compilation requires toolchain.

Forge is a hack to read the JVM path from the toolchain to run the test but we do not need tollchain to run the test, just a place where a mapping exist that we can use.

akomakom:
Thanks for that, that's interesting.  I ran the build with this extension, the build took 1:20 but the extension only accounted for 38 seconds: (I aggregated the totals on mojo)

|                                                              |                  | 
|--------------------------------------------------------------|------------------| 
| Mojo                                                         | Sum - Time       | 
| buildnumber-maven-plugin:create (default)                    | 0.514            | 
| maven-antrun-plugin:run (default)                            | 0.815            | 
| maven-assembly-plugin:single (platform-kit)                  | 2.785            | 
| maven-bundle-plugin:manifest (default)                       | 2.842            | 
| maven-clean-plugin:clean (default-clean)                     | 0.254            | 
| maven-compiler-plugin:compile (default-compile)              | 0.21             | 
| maven-compiler-plugin:testCompile (default-testCompile)      | 27.382           | 
| maven-dependency-plugin:copy (copy-test-entity)              | 0.004            | 
| maven-dependency-plugin:unpack (unpack-kit)                  | 0.424            | 
| maven-enforcer-plugin:enforce (enforce-java)                 | 0.036            | 
| maven-forge-plugin:integration-test (default-verify)         | 0.135            | 
| maven-forge-plugin:test (default-test)                       | 0.427            | 
| maven-forge-plugin:verify (default-verify)                   | 0.028            | 
| maven-install-plugin:install (default-install)               | 0.423            | 
| maven-jar-plugin:jar (default-jar)                           | 0.699            | 
| maven-jaxb2-plugin:generate (default)                        | 0.013            | 
| maven-resources-plugin:copy-resources (copy-logging-config)  | 0.004            | 
| maven-resources-plugin:resources (default-resources)         | 0.055            | 
| maven-resources-plugin:testResources (default-testResources) | 0.073            | 
| maven-shade-plugin:shade (default)                           | 0.093            | 
| maven-source-plugin:jar (attach-sources)                     | 0.615            | 
| maven-source-plugin:jar-no-fork (default)                    | 0.182            | 
| maven-surefire-plugin:test (default-test)                    | 0.103            | 
| maven-toolchains-plugin:toolchain (default)                  | 0.049            | 
| Total Result                                                 | 38.1650000000001 | 
 

So if that's accurate, removing toolchains and enforcer will buy you a whole `0.085 seconds`

mathieucarbou:
> Thanks for that, that's interesting. I ran the build with this extension, the build took 1:20 but the extension only accounted for 38 seconds: (I aggregated the totals on mojo)
> 	
> Mojo 	Sum - Time
> buildnumber-maven-plugin:create (default) 	0.514
> maven-antrun-plugin:run (default) 	0.815
> maven-assembly-plugin:single (platform-kit) 	2.785
> maven-bundle-plugin:manifest (default) 	2.842
> maven-clean-plugin:clean (default-clean) 	0.254
> maven-compiler-plugin:compile (default-compile) 	0.21
> maven-compiler-plugin:testCompile (default-testCompile) 	27.382
> maven-dependency-plugin:copy (copy-test-entity) 	0.004
> maven-dependency-plugin:unpack (unpack-kit) 	0.424
> maven-enforcer-plugin:enforce (enforce-java) 	0.036
> maven-forge-plugin:integration-test (default-verify) 	0.135
> maven-forge-plugin:test (default-test) 	0.427
> maven-forge-plugin:verify (default-verify) 	0.028
> maven-install-plugin:install (default-install) 	0.423
> maven-jar-plugin:jar (default-jar) 	0.699
> maven-jaxb2-plugin:generate (default) 	0.013
> maven-resources-plugin:copy-resources (copy-logging-config) 	0.004
> maven-resources-plugin:resources (default-resources) 	0.055
> maven-resources-plugin:testResources (default-testResources) 	0.073
> maven-shade-plugin:shade (default) 	0.093
> maven-source-plugin:jar (attach-sources) 	0.615
> maven-source-plugin:jar-no-fork (default) 	0.182
> maven-surefire-plugin:test (default-test) 	0.103
> maven-toolchains-plugin:toolchain (default) 	0.049
> Total Result 	38.1650000000001
> 
> So if that's accurate, removing toolchains and enforcer will buy you a whole `0.085 seconds`

Well, the build takes only 30sec in master... It takes 300% master's time on this PR. 
I did the same aggregation you did with excel over our ~100 modules and on my computer, the buildnumber plugin account for ~2 seconds.

akomakom:
So are we discussing a 1 minute build time change on a project in which tests take >1 hour?  And, assuming that this is really important, which plugin do you believe accounts for the difference, if not toolchains, enforcer, or buildnumber?

mathieucarbou:
> So are we discussing a 1 minute build time change on a project in which tests take >1 hour? And, assuming that this is really important, which plugin do you believe accounts for the difference, if not toolchains, enforcer, or buildnumber?

no you didn't get it!

The project can take 1 hour on Azure. That is ok, this is a full build.

__What we need as devleppers is to have a project build time that is short__ This is the point I try to make since the beginning here, That is why we have this `-Pfast` mode that avoids executing the plugins not needed. We need to have the ability to build a project fast because when you want to run the system tests in your IDE, try a fix, etc, yo uneed to rebuild everything. Maven does not have a build cache like gradle. So we use this mode to reduce the build time considerably to only be able to compile, package, build a kit and install.

akomakom:
@mathieucarbou I think I understand your goals.  I can make these changes, although most likely you'll have to invoke `-Dfast` instead of `-Pfast` going forward (profile activation with parent poms is a mess)

mathieucarbou:
> @mathieucarbou I think I understand your goals. I can make these changes, although most likely you'll have to invoke `-Dfast` instead of `-Pfast` going forward (profile activation with parent poms is a mess)

no problem :-) as long as for the development we have a very quick way of building the project :-)

mathieucarbou:
> > @mathieucarbou I think I understand your goals. I can make these changes, although most likely you'll have to invoke `-Dfast` instead of `-Pfast` going forward (profile activation with parent poms is a mess)
> 
> no problem :-) as long as for the development we have a very quick way of building the project :-)

Side note this is even better using properties because: `-Pfast` relied on the fact that when a user specifies a profile on the CLI, Maven automatically disables the profiles that are activated by default (`<activeByDefault>true</activeByDefault>`). Which is why I have in master a default profile called slow with everything inside, which is the default build, and I have a fast profile that only contains the compilation and packaging, and the activation of this profile would automatically deactivate any other profiles with ``<activeByDefault>true</activeByDefault>``

But this is not good because it can lead to issues if there are some required profiles in sub modules...

So definitely we should go with properties :-)

akomakom:
@mathieucarbou you can check the current commit for speed with and without `-Dfast`

mathieucarbou:
> @mathieucarbou you can check the current commit for speed with and without `-Dfast`



> @mathieucarbou you can check the current commit for speed with and without `-Dfast`

Awesome!!! Fast is now pretty fast!!!

- `./mvnw clean install -DskipTests -Dfast` => `[INFO] Total time:  26.822 s`
- `./mvnw clean install -DskipITs -Dfast` => `[INFO] Total time:  02:23 min` + only run unit test 👍

pom.xml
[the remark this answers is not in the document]
akomakom: The properties file is for gradle, because gradle doesn't have a notion of toolchains.  Maven does, so we're using it.  Reading a JVM path from toolchains.xml (via the toolchains plugin API) is not a hack - that's what the toolchains plugin does and thus that is what all the core plugins do.  This includes surefire/failsafe, they too honor toolchains.  What forge adds is the ability to use a different toolchain-specified JDK from the main maven toolchain-specified JDK (ie compile with 8, test with 11).

pom.xml
```diff
@@ -334,22 +296,7 @@
...
-          <dependencies>
-            <!-- overwrite dependency on spotbugs if you want to specify the version of spotbugs -->
-            <dependency>
-              <groupId>com.github.spotbugs</groupId>
-              <artifactId>spotbugs</artifactId>
-              <version>4.0.0</version>
-            </dependency>
-          </dependencies>
-        </plugin>
+
```
mathieucarbou, line 299: extra space

pom.xml
```diff
@@ -404,6 +351,7 @@
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-compiler-plugin</artifactId>
       </plugin>
+
```
mathieucarbou, line 354: extra space

pom.xml
```diff
@@ -436,37 +384,45 @@
           </execution>
         </executions>
       </plugin>
+
```
mathieucarbou, line 387: extra space

pom.xml
```diff
@@ -531,88 +482,42 @@
...
+                    <echo message="Configured test argLine: '${java.test.argLine}'"/>
+                  </target>
+                </configuration>
+              </execution>
+            </executions>
           </plugin>
         </plugins>
       </build>
     </profile>
+
```
mathieucarbou, line 520: extra space

pom.xml
```diff
@@ -531,88 +482,42 @@
...
+                  <goal>run</goal>
+                </goals>
+                <configuration>
+                  <exportAntProperties>true</exportAntProperties>
+                  <target>
+                    <!-- only add the jvm param if we're testing on newer JDKs -->
+                    <condition property="java.test.argLine" value="" else="--illegal-access=permit">
+                      <equals arg1="${java.test.version}" arg2="1.8"/>
+                    </condition>
+                    <echo message="Configured test argLine: '${java.test.argLine}'"/>
```
mathieucarbou, line 511: to remove or comment

pom.xml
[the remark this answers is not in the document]
mathieucarbou: my point was just to say: 

- we already have the logic in place to feed a property file with the jvm path in azure
- in master, we already have the ability to put a property in the <jvm> attributes of surefire / failsafe.

So I am just saying that it would have probably been quicker and involved less modification to load a property file and use those properties in place of the hard-coded path, then all these changes with forge and parent pom.

At the end, what we need, is to run the tests with a different JVM. We can achieve that goal without involving the toolchain plugin :-)

And this mimics also closer what gradle is doing in Ehcahce: they change the JVM to use to run the tests based on a property file.

akomakom:
Made the changes, but [tests are failing](https://dev.azure.com/TerracottaCI/terracotta/_build/results?buildId=4134&view=ms.vss-test-web.build-test-results-tab)

pom.xml
```diff
@@ -538,84 +481,53 @@
...
+                  <target>
+                    <!--
+                      only add the jvm param if we're testing on newer JDKs
+                      Note: this property should not have a default in properties section, otherwise
+                      antrun may not be able to overwrite its value (or at least won't display it correctly)
+                    -->
+                    <condition property="java.test.argLine" value="" else="--illegal-access=permit">
+                      <equals arg1="${java.test.version}" arg2="1.8"/>
+                    </condition>
+                    <echo message="Configured test argLine (for java ${java.test.version}): '${java.test.argLine}'"/>
```
mathieucarbou, line 526: the echo is still there

akomakom: It was malfunctioning until recently, so I left it intentionally
If you're sure that having the variable set incorrectly will be obvious in the test failures, I can remove it.

pom.xml
```diff
@@ -538,84 +481,53 @@
...
+              See antrun plugin in this pom.
+             -->
             <configuration>
-              <trimStackTrace>false</trimStackTrace>
-              <argLine>--illegal-access=permit</argLine>
-              <jvm>C:\program files\java\zulu-11-azure-jdk_11.33.15-11.0.4-win_x64\bin\java.exe</jvm>
-              <environmentVariables>
-                <JAVA_HOME>C:\program files\java\zulu-11-azure-jdk_11.33.15-11.0.4-win_x64</JAVA_HOME>
-              </environmentVariables>
+              <argLine>${java.test.argLine}</argLine>
```
mathieucarbou, line 492: `<trimStackTrace>false</trimStackTrace>` is missing compared to master

pom.xml
mathieucarbou: - If this property is incorrectly set, we will see it with the java 11 build.

But this makes me think of a potential issue...

The responsibility of setting this property should be up to the project's pom who uses the forge plugin. This is what we do in `management/sequence-generator/pom.xml` where the arg line is set for example. So the global one is override.

So more generally, there should be an easy way for the project using forge to be able to declare some profiles for java 11 and put the necessary arguments to correctly run the tests. They could be some system props, java args, or anything.

So instead of using the antrun plugin which could easily lead to mistakes especially in some modules where we need to redefine the argLine, could we instead have the ability as projects and or submodules to be able to activate a profile based on the property passed ?

Something like that:

```xml
    <profile>
      <id>testJVM11</id>
      <activation>
        <property>
          <name>java.test.version</name>
          <value>1.11</value>
        </property>
      </activation>
      <build>
        <plugins>
          <plugin>
            <configuration>
              <argLine>--illegal-access=permit</argLine>
            </configuration>
          </plugin>
        </plugins>
      </build>
    </profile>
```

patch:

```diff
Index: pom.xml
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
--- pom.xml	(revision f7d6f6c8318cd83a97461290d86268414f218489)
+++ pom.xml	(date 1586903370812)
@@ -484,50 +484,30 @@
           <plugin>
             <groupId>org.terracotta</groupId>
             <artifactId>maven-forge-plugin</artifactId>
-            <!--
-              configuration added to what is inherited from parent
-              See antrun plugin in this pom.
-             -->
-            <configuration>
-              <argLine>${java.test.argLine}</argLine>
-            </configuration>
           </plugin>
+        </plugins>
+      </build>
+    </profile>
+    <profile>
+      <id>testJVM11</id>
+      <activation>
+        <property>
+          <name>java.test.version</name>
+          <value>1.11</value>
+        </property>
+      </activation>
+      <build>
+        <plugins>
           <plugin>
-            <!--
-              Workaround for maven not being able to set a property conditionally based on parent property.
-
-              When testing with Java 9+ we need to pass a flag to the jvm (illegal-access).
-              When testing with Java 8, we must not, since it's not supported.
-
-              Since $java.test.version is set in the parent, we can't easily key a profile off the value
-              of this variable (profile activation doesn't resolve maven variables).  Using antrun instead to set
-              the value of $java.test.argLine and referencing it in maven-forge-definition.
-            -->
-            <groupId>org.apache.maven.plugins</groupId>
-            <artifactId>maven-antrun-plugin</artifactId>
-            <version>1.8</version>
-            <executions>
-              <execution>
-                <phase>validate</phase>
-                <goals>
-                  <goal>run</goal>
-                </goals>
-                <configuration>
-                  <exportAntProperties>true</exportAntProperties>
-                  <target>
-                    <!--
-                      only add the jvm param if we're testing on newer JDKs
-                      Note: this property should not have a default in properties section, otherwise
-                      antrun may not be able to overwrite its value (or at least won't display it correctly)
-                    -->
-                    <condition property="java.test.argLine" value="" else="--illegal-access=permit">
-                      <equals arg1="${java.test.version}" arg2="1.8"/>
-                    </condition>
-                    <echo message="Configured test argLine (for java ${java.test.version}): '${java.test.argLine}'"/>
-                  </target>
-                </configuration>
-              </execution>
-            </executions>
+            <groupId>org.terracotta</groupId>
+            <artifactId>maven-forge-plugin</artifactId>
+            <configuration>
+              <argLine>--illegal-access=permit</argLine>
+              <jvm>/Users/matc/.jenv/versions/11.0/bin/java</jvm>
+              <environmentVariables>
+                <JAVA_HOME>/Users/matc/.jenv/versions/11.0</JAVA_HOME>
+              </environmentVariables>
+            </configuration>
           </plugin>
         </plugins>
       </build>
```

I verified and this seems to work because the forge and toolchain seem to be triggered with the right JDK. Sadly because of the other issue I have reported the tests are still running with the 1.8 java home.

```log
[INFO] --- maven-forge-plugin:1.2.32:test (default-test) @ integration-tests ---
[INFO] Setting surefire's jvm to /Users/matc/.jenv/versions/11.0/bin/java from toolchain JDK[/Users/matc/.jenv/versions/11.0], requirements: {version=1.11}
[INFO] Toolchain in maven-surefire-plugin: JDK[/Users/matc/.jenv/versions/1.8]
[WARNING] Toolchains are ignored, 'jvm' parameter is set to /Users/matc/.jenv/versions/11.0/bin/java
```

mathieucarbou:
Closing since this is resolved by PR https://github.com/Terracotta-OSS/terracotta-platform/pull/663 in commit 791a3d4cdc5e86117d47e1b0876e1dc2b2f6e709.

Solution implemented by using the GMaven plugin to read the toolchains.xml file to extract the JVM path for the requested JVM version, which will be used for the tests,
Solution works for any java version on the toolchain file: 1.8, 1.9, 1.11, etc.
Solution also keeps the same plugin versions as we had in tc-platform before.

Consequently, the issue  (ability to run the tests by looking up the JVM path in the toolchains) is  solved at the moment without involving any parent pom, plugin version changes and forge integration.

If someone sees a missing use case or an issue not resolved, or a valid use case requiring the use of one of these things above, feel free to open another issue describing the problem / use case to solve. The current issue we needed to solve is solved at the moment by using a simple plugin. So except if there is another need of something else or issue somewhere, __then let's not spend much more dev and review time on this__ We can open an issue / story if we see other issue or use cases and a new PR without all this long conversation ;-) We just have to keep in mind the following to verify any work.

In any case, if there is a change of implementation, __here are the requirements and things to verify for this project__ to have parity with master:

### Fast build

This project is big and development and testing time is highly impacted by Maven build time since there is no gradle cache.
So we need to be able several times per day to build the project __fast__.

`./mvnw clean install -DskipTests -Dfast`

The command should only do the minimal required steps:

- compilation steps
- felix
- packaging (jars and source jars - to use snapshots in downstream projects)

__No toolchain, verfication, license check, findbugs, etc.__

The command needs to be fast: __less than 30 seconds.__

### Unit tests vs IT tests

We should be able to run the unit tests only and all tests:

- `./mvnw clean verify -DskipTests -Dfast` => should skip all tests
- `./mvnw clean verify -DskipITs -Dfast` => should skip IT Galvan and Angela tests but run unit tests (__last ~2-3min__)
- `./mvnw clean verify` => should run everything, like on Azure. __Lasts ~1h20min__

### Plugin config

- Use `<trimStackTrace>false</trimStackTrace>`
- Use `<argLine>--illegal-access=permit</argLine>` (only for Java 11, because core is doing some illegal reflection calls)

### Test JVM

We should be able to run the tests with one of the JVM on the toolchain but compilation done in 1.8

__Examples__

Run a test with Java 11 (will look at your toolchains.xml file to fine the JVM):

```
./mvnw verify -f management/testing/integration-tests/pom.xml -Dit.test=DiagnosticIT -Djava.test.version=1.11
```

Run a test with Java 8 (will look at your toolchains.xml file to fine the JVM):

```
./mvnw verify -f management/testing/integration-tests/pom.xml -Dit.test=DiagnosticIT -Djava.test.version=1.8
```

Run a test with Java 8 (will look at your toolchains.xml file to fine the JVM - 1.8 is teh default):

```
./mvnw verify -f management/testing/integration-tests/pom.xml -Dit.test=DiagnosticIT
```

Run a test by using the default JAVA_HOME found from your shell. -Dfast has no enforcement at all and does not use toolchain. 
-Dfast goal is to compile and package fast.

```
./mvnw verify -f management/testing/integration-tests/pom.xml -Dit.test=DiagnosticIT -Dfast
```

__The same applies for Angela tests in dynamic-config/testing/system-tests and they should be verified as well__

__Verification__

Look inside the servers logs of the nodes starte by Angela and Galvan.
The server logs contain the list of system properties and JVM details used to start the servers.
The version should match the one requested.

- For Angela tests: `target/angela/work/XYZ/logs/stripe1/node-1-1/terracotta.server.log`
- For Galvan tests: `target/galvan/XYZ/stripe1/testServer0/logs/terracotta.server.log`

Look for something like that in the logs:

```
java.runtime.name             : OpenJDK Runtime Environment
java.runtime.version          : 11.0.6+10
java.specification.name       : Java Platform API Specification
java.specification.vendor     : Oracle Corporation
java.specification.version    : 11
```

### IDE support

After having done a `./mvnw clean verify -DskipTests -Dfast` to build the project and kit, you should be able to go in your IDE and right-click run the tests for example.

This must be verified both both Angela tests and management tests, by trying ot run the `DiagnosticIT` test. 
This tests is in both modules.

PR #645, opened 2020-04-06
Temporary activating windows builds and Java 11 builds by hard-coding JVM paths in profiles

mathieucarbou:
- To be able to run Windows builds
- To be able to run Java 11  builds correctly in Galvan (needs JAVA_HOME to be set)
- fix the failing tests.

This is not the ideal solution. Ideally, we would need #642, but it is not working yet.

So this is a temporary replacement.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -38,17 +38,17 @@ jobs:
       timeoutInMinutes: 120
       vmImage: 'ubuntu-latest'
       options: '-B -Dazure_linux_java_11'
-#  - template: build-templates/maven-common.yml@templates
-#    parameters:
-#      jdkVersion: '1.8'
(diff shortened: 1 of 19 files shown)


mathieucarbou:
@saurabhagas @akomakom : tehre is an incompatibility between PortChooser of Galvan, Azure and Windows. So any progress on activating the Windows build is blocked until this is solved: https://github.com/Terracotta-OSS/galvan/issues/215

mathieucarbou:
@mobasherul @saurabhagas : I was able to make all windows builds pass. 
I am restarting the build another time again, but I think all the tests that were subject to timing or issues are resolved. The Azure windows VM are really really slow compared to linux ones.
Could you please approve ?
Thanks!

PR #647, opened 2020-04-07
Dependency scope fixes

mathieucarbou:
(See https://github.com/ehcache/ehcache3/pull/2756)

### DIFF
--- a/common/structures/pom.xml
+++ b/common/structures/pom.xml
@@ -33,6 +33,7 @@
     <dependency>
       <groupId>com.fasterxml.jackson.core</groupId>
       <artifactId>jackson-annotations</artifactId>
+      <scope>provided</scope>
     </dependency>
 
     <dependency>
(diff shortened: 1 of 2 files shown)


common/structures/pom.xml
```diff
@@ -33,6 +33,7 @@
     <dependency>
       <groupId>com.fasterxml.jackson.core</groupId>
       <artifactId>jackson-annotations</artifactId>
+      <scope>provided</scope>
```
mathieucarbou, line 36: only modules that are using this module + jackson databind need the annotation jar, and this jar will come as a dependency from jackson anyway.

chrisdennis: Shouldn't this kind of 'only needed for certain things' dependency be an 'optional' dependency and not a provided one?

mathieucarbou: no it is a dependency that is provided by jackson jars when using jackson. Like kit is providing some api libs. This is not optional, otherwise it would mean that we can use jackons databind without this jar, which is not the case

PR #657, opened 2020-04-13
Fix #656: Many issues when attaching / detaching unconfigured nodes

### LINKED ISSUE #656: Detach command problems for unconfigured nodes
The new detach command UX and behavior is both confusing and wrong. The command expects the nodes to be removed to be already shut down, but the behavior is odd either way.

**Case 1: Nodes being detached are shut down**
Detach command fails with "com.terracotta.connection.api.DetailedConnectionException: java.util.concurrent.TimeoutException: localhost:9410=Connection refused".

**Case 2: Nodes being detached are not shut down**
The nodes in question should be shut down by the command upon passing `-f` option, but it doesn't happen if the nodes are not activated.

mathieucarbou:


mathieucarbou:
@saurabhagas : can you approve the PR please so that we can meege as soon as build pass.
It is highly likely that I will have to restart a Windows build if one is failing because of some timeout or slowness.

PR #666, opened 2020-04-15
maven-forge-plugin via parent for failsafe/surefire toolchains

akomakom:
Re-attempt of #642

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -19,7 +19,12 @@
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
   <modelVersion>4.0.0</modelVersion>
 
-  <groupId>org.terracotta</groupId>
+  <parent>
+    <groupId>org.terracotta</groupId>
+    <artifactId>terracotta-parent</artifactId>
+    <version>5.13</version>
+  </parent>
+
   <artifactId>platform-root</artifactId>
   <version>5.7-SNAPSHOT</version>
   <packaging>pom</packaging>
@@ -39,9 +44,6 @@
     <jackson.version>2.10.1</jackson.version>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
-    <java.build.version>1.8</java.build.version>
-    <!-- Use -Djava.test.version=1.11 for example to run tests om Java 11 JVM -->
-    <java.test.version>${java.build.version}</java.test.version>
   </properties>
 
   <modules>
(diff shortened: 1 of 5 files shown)


pom.xml
```diff
@@ -261,72 +263,15 @@
...
           <groupId>org.apache.maven.plugins</groupId>
           <artifactId>maven-compiler-plugin</artifactId>
-          <version>3.8.1</version>
           <configuration>
-            <source>${java.build.version}</source>
-            <target>${java.build.version}</target>
             <compilerArgs>
               <arg>-Xlint:all</arg>
               <arg>-Werror</arg>
             </compilerArgs>
```
mathieucarbou, line 270: @chrisdennis : should we move that to the parent POM, now that all projects should compile without warnings ? @myronkscott Don't know about core if it will hurt or not ? ;-)

pom.xml
```diff
@@ -432,234 +355,186 @@
...
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-source-plugin</artifactId>
+        <version>3.2.1</version>
+        <executions>
+          <execution>
+            <goals>
+              <goal>jar-no-fork</goal>
+            </goals>
+          </execution>
+        </executions>
```
mathieucarbou, line 392: Why we need to move these 2 plugins from the plugin management section (used to declare and configure plugins) ? 
Could we just have their declaration and configuration in plugin management like the other ones, and their usage in <build> section, like it is also done in sub modules ?

pom.xml
```diff
@@ -432,234 +355,186 @@
...
-    <!--
-      Default profile is slow.
-      It includes: including toolchain, javadoc, findbugs, license check, enforcer, etc.
-      Use any Maven command with -Dfast to speed up the build.
-      -Dfast can be used daily and will just do compilation and packaging
-    -->
+    <!-- default profile including toolchain, javadoc, findbugs, license check, enforcer, etc -->
     <profile>
-      <id>default</id>
+      <id>slow</id>
```
mathieucarbou, line 400: could we change to `default` to match the comment and the fact that this is actually a default profile, slow being one of its characteristic ;-)

pom.xml
```diff
@@ -261,72 +263,15 @@
...
-              <goals>
-                <goal>toolchain</goal>
-              </goals>
-            </execution>
-          </executions>
-        </plugin>
-        <plugin>
+          <inherited>false</inherited>
           <groupId>com.mycila</groupId>
           <artifactId>license-maven-plugin</artifactId>
```
mathieucarbou, line 276: <artifactId>license-maven-plugin</artifactId> could be moved to the plugin management section of the profile

pom.xml
```diff
@@ -432,234 +355,186 @@
...
+              <dependencies>
+                <!-- overwrite dependency on spotbugs if you want to specify the version of spotbugs -->
+                <dependency>
+                  <groupId>com.github.spotbugs</groupId>
+                  <artifactId>spotbugs</artifactId>
+                  <version>4.0.0</version>
+                </dependency>
+              </dependencies>
+            </plugin>
+          </plugins>
```
mathieucarbou, line 426: I completely forgot that we could have a plugin management section per profile 👍 
Why only spotbugs is there ? Could we have also the other ones declared and moved here too (i.e. license) ?

pom.xml
```diff
@@ -432,234 +355,186 @@
...
+            <!--
+              configuration added to what is inherited from parent
+              See antrun plugin in this pom.
+             -->
             <configuration>
-              <jvm>${java.test.bin}</jvm>
-              <environmentVariables>
-                <JAVA_HOME>${java.test.home}</JAVA_HOME>
-              </environmentVariables>
+              <argLine>${java.test.argLine}</argLine>
```
mathieucarbou, line 499: this can be removed: BC and Chris confirmed that there is no argLine to pass.

pom.xml
```diff
@@ -432,234 +355,186 @@
...
+            -->
             <groupId>org.apache.maven.plugins</groupId>
-            <artifactId>maven-surefire-plugin</artifactId>
-            <configuration>
-              <jvm>${java.test.bin}</jvm>
-              <environmentVariables>
-                <JAVA_HOME>${java.test.home}</JAVA_HOME>
-              </environmentVariables>
-            </configuration>
+            <artifactId>maven-antrun-plugin</artifactId>
```
mathieucarbou, line 514: plugin can be removed: BC and Chris confirmed that there is no argLine to pass.

pom.xml
akomakom: I checked, and tc-core does not compile with `-Werror`.   However, I could put this in the parent based on a property (or two properties) so that it can be turned off.   Or we could leave that up to the individual projects to specify explicitly.

pom.xml
akomakom: @mathieucarbou I see two parts to this plugin, one in pluginManagement and the other in plugins.  The part in plugins merely specifies an execution, but it also has a suspicious <inherited>false ... Documentation suggests that "check" is automatically bound to `verify`, so I don't know why any of this is in the pom in the first place.  Do you know?

pom.xml
mathieucarbou: I agree with you. Since it has been decided that all project now should compile with no warnings, these options should be there by default. If a project cannot comply, even temporary, then it should be able to deactivate.

I don't think we need some properties because core could just override compile plugin and control these options, like setting `-Xlint:none` instead of `-Xlint:all`

pom.xml
akomakom: That must have been a bad merge

pom.xml
mathieucarbou: yesI did this plugin ;-) `<inherited>false<inherited>` is required for the license plugin in case you have a multi-module project and you want to execute it only once form the paent pom across all the project files. This is also required because of that: `<header>header.txt</header>` If the plugin executes in some sub-modules, then it won't find the header.
The license plugin has some pre-defined headers already, APACHE being suppported, but sadly we are overriting the apache license to add: `Copyright Terracotta, Inc.`. I don't even know if this is legal to alter the apache license like that... I think it breaks its compatibility becasue of the copyright.
The license plugin is also able to take teh license from any URI or even a classpath, when license is put in a jar as a plugin dependency.
So depending how it is configured and which type of project, inherited=false is required or not.
For our OSS projects, I think we should in theory go with the normal apache supported license without providing our own. I don't know where this one comes from but what is sure is that Copyrighting a source file as it is one in the current license is not compatible with the terms of the apache license.

pom.xml
akomakom: Depends on what you'd like to be able to do with `-Dfast`... if you want a **valid** jar/kit built, then no.

pom.xml
akomakom: Happy to hear that

pom.xml
mathieucarbou: @akomakom to add, here are the license of plugins plus template that comes with the license plugin: https://github.com/mycila/license-maven-plugin/blob/master/license-maven-plugin/src/main/resources/com/mycila/maven/plugin/license/templates/. The template already supports some placeholders.
Can be accessed like that:
`<header>com/mycila/maven/plugin/license/templates/APACHE-2.txt</header>`

pom.xml
akomakom: :+1:

pom.xml
mathieucarbou: I was speaking about moving the plugin declaration and confoguration from the `<plugins>` section of the profile only into the `<pluginManagement>` section of the profile only.

Here, we have a profile with both `pluginManagement>` and `plugins`. 
So my question was: why did you applied a clean recipe (separation between plugin declaration/config and plugin usage) only to spotBugs and not the other ones ?

Also, my you have kept license plugin in the main pluginManagement section and not moved it in the profile pluginManagement section ?

pom.xml
akomakom: I have no idea whether we can switch from a tc license to apache, but perhaps we can move this plugin definition to terracotta-parent (using tc-core definition as it appears to have a superset of exclusions)?

mathieucarbou: yes this plugin is really configured per project.
But in the parent pom, you can enforce its execution though, and set an id to the execution section: in case a sub project wants to disable the plugin execution, he will be able to thanks to the id.

akomakom: So I'm looking at platform and core... and core doesn't appear to declare any execution for the license plugin at this time.  So either it's not necessary, or it doesn't work in core (I'm guessing - the latter)

mathieucarbou: It is because tc-platform project has been created more recently - we added license plugin with Anthony. core project was already there. But perhaps Myron/Sairam would be interested in having that because they sometimes commit files without headers I saw ;-)

akomakom: What I meant was that core [does mention the license plugin](https://github.com/Terracotta-OSS/terracotta-core/blob/master/pom.xml#L87) but does not define any executions.  I just removed a license header from source and it still compiled :)

akomakom: ... which is one of the reasons I want to move common elements to parent, so that things are not easily overlooked and do not fall out of sync.  (from an infra perspective).

akomakom: @mathieucarbou @myronkscott should license checking happen at:
* compile (fail fast)
* verify  (check any possible generated files)

mathieucarbou: I designed the plugin for an execution during verify phase. And generated files (in target folder) should be excluded from the check: there are some default exclusions for everything inside target/**. Except if explicitly changed.

akomakom: Now I'm puzzled.  It seems to me that license check at compile would be the same as license check at verify with exclusions... except that there won't need to be any exclusions... unless I'm misunderstanding.

pom.xml
akomakom: That works

pom.xml
akomakom: Here is my plan: https://github.com/Terracotta-OSS/terracotta-parent/pull/13  Exclusions seem to work for core and platform, and I'll be making a PR for core because it never had license plugin enabled until now, so half the headers were wrong or missing.  Currently license execution is bound to `compile`, but I'm not married to that.

mathieucarbou: put some comments over there ;-)

pom.xml
```diff
@@ -251,144 +253,16 @@
...
-          </dependencies>
-          <executions>
-            <execution>
-              <goals>
-                <goal>check</goal>
-              </goals>
-            </execution>
-          </executions>
-        </plugin>
+
```
mathieucarbou, line 265: there are some extra lines left around this plugin after the removal

pom.xml
```diff
@@ -613,57 +372,54 @@ log.info("java.test.bin => " + project.properties['java.test.bin'])
         </property>
       </activation>
       <build>
+        <pluginManagement>
+          <plugins>
+
+
+          </plugins>
+        </pluginManagement>
```
mathieucarbou, line 380: `<pluginManagement>` section can be removed

pom.xml
```diff
@@ -613,57 +372,54 @@ log.info("java.test.bin => " + project.properties['java.test.bin'])
...
+            <executions>
+              <execution>
+                <goals>
+                  <goal>jar</goal>
+                </goals>
+              </execution>
+            </executions>
           </plugin>
+
+
```
mathieucarbou, line 422: some extra lines there

pom.xml
```diff
@@ -39,9 +44,6 @@
     <jackson.version>2.10.1</jackson.version>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
```
mathieucarbou, line 46: oh ! just saw that these 2 should come from the top also. we could move them in parent pom.

akomakom: One is already in parent, adding the other.

pom.xml
mathieucarbou: blanks likes still in the diff ;-)

akomakom:
:clap:

PR #668, opened 2020-04-16
Attach Detach with Voter running

mobasherul:
Attach Detach with Voter running

### DIFF
--- a/dynamic-config/testing/system-tests/pom.xml
+++ b/dynamic-config/testing/system-tests/pom.xml
@@ -77,6 +77,11 @@
       <version>${logback.version}</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.terracotta.internal</groupId>
+      <artifactId>voter</artifactId>
+      <scope>test</scope>
+    </dependency>
   </dependencies>
 
   <build>
@@ -158,6 +163,7 @@
             <angela.kitInstallationDir>${project.build.directory}/platform-kit-${project.version}</angela.kitInstallationDir>
             <angela.java.vendor/> <!--empty to allow any vendor-->
             <angela.java.version>${java.test.version}</angela.java.version>
+            <org.terracotta.voter.topology.fetch.interval>9000</org.terracotta.voter.topology.fetch.interval>
             <!--<angela.java.opts>-agentlib:jdwp=transport=dt_socket,server=y,suspend=y,address=5005</angela.java.opts>-->
           </systemPropertyVariables>
         </configuration>
--- a/dynamic-config/testing/system-tests/src/test/java/org/terracotta/dynamic_config/system_tests/activated/AttachCommandWithVoter1x3IT.java
+++ b/dynamic-config/testing/system-tests/src/test/java/org/terracotta/dynamic_config/system_tests/activated/AttachCommandWithVoter1x3IT.java
@@ -1,0 +1,179 @@
+/*
+ * Copyright Terracotta, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.terracotta.dynamic_config.system_tests.activated;
+
+import org.junit.Before;
+import org.junit.Ignore;
+import org.junit.Test;
+import org.terracotta.dynamic_config.api.model.FailoverPriority;
+import org.terracotta.dynamic_config.test_support.ClusterDefinition;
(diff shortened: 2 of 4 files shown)


dynamic-config/testing/system-tests/pom.xml
```diff
@@ -77,6 +77,12 @@
       <version>${logback.version}</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.terracotta.internal</groupId>
+      <artifactId>voter</artifactId>
+      <version>5.7.0-pre25</version>
```
mathieucarbou, line 83: the dependency should be put in the parent pom and the version shouldn't be hardcoded. Yo ucan look at how its done from the other core dependencies. Projects are using them without verison tag,and dependency is defined in the parent pom.

dynamic-config/testing/system-tests/src/test/java/org/terracotta/dynamic_config/system_tests/activated/AttachCommandWithVoter1x3IT.java
```diff
@@ -0,0 +1,174 @@
...
+    Set<String> expectedRes = new HashSet<>();
+    expectedRes.add(getNode(1, activeId).getHostPort());
+    expectedRes.add(getNode(1, passiveId).getHostPort());
+    expectedRes.add(getNode(1, 3).getHostPort());
+
+    Thread.sleep(TOPOLOGY_FETCH_INTERVAL);
+    assertThat(activeVoter.getExistingTopology(), is(expectedRes));
+    assertThat(activeVoter.getHeartbeatFutures().size(), is(3));
+
+    activeVoter.stop();
```
mathieucarbou, line 142: use a try/finally to make sure the voter is stopped in case of a test failure

dynamic-config/testing/system-tests/src/test/java/org/terracotta/dynamic_config/system_tests/activated/AttachCommandWithVoter1x3IT.java
```diff
@@ -0,0 +1,174 @@
...
+    waitForDiagnostic(1, 3);
+    assertThat(getUpcomingCluster("localhost", getNodePort(1, 3)).getNodeCount(), is(equalTo(1)));
+
+    //Kill active so other passive becomes active by the vote of voter
+    stopNode(1, activeId);
+    waitForActive(1, passiveId);
+
+    // Not able to connect to the new active 
+    // Perhaps because of two concurrent diagnostic connection one from voter for heartbeating and other from config tool
+    // Investigate ?
```
mathieucarbou, line 130: damn! thanks for finding that! I have assigned this issue to you and put it in the sprint. Same for TDB-4947.

pom.xml
```diff
@@ -123,6 +123,12 @@
         <artifactId>client-runtime</artifactId>
         <version>${terracotta-core.version}</version>
       </dependency>
+      <dependency>
+        <groupId>org.terracotta.internal</groupId>
+        <artifactId>voter</artifactId>
+        <version>${terracotta-core.version}</version>
+        <scope>test</scope>
```
mathieucarbou, line 130: no scope when defining a dependency. scope is when the dependency is used.

dynamic-config/testing/system-tests/src/test/java/org/terracotta/dynamic_config/system_tests/activated/DetachCommandWithVoter1x2IT.java
```diff
@@ -0,0 +1,118 @@
...
+import static org.terracotta.dynamic_config.test_support.util.AngelaMatchers.successful;
+import static org.terracotta.voter.ActiveVoter.TOPOLOGY_FETCH_TIME_PROPERTY;
+
+@ClusterDefinition(nodesPerStripe = 2, autoActivate = true)
+public class DetachCommandWithVoter1x2IT extends DynamicConfigIT {
+  private static final long TOPOLOGY_FETCH_INTERVAL = 11000L;
+
+  static {
+    System.setProperty(TOPOLOGY_FETCH_TIME_PROPERTY, "9000");
+  }
```
mathieucarbou, line 44: We should never have system properties statically set like that  in a test because it impacts the whole test JVM.
Can this property be set another way ?

dynamic-config/testing/system-tests/src/test/java/org/terracotta/dynamic_config/system_tests/activated/AttachCommandWithVoter1x3IT.java
```diff
@@ -0,0 +1,184 @@
...
+import static org.terracotta.dynamic_config.test_support.util.AngelaMatchers.successful;
+import static org.terracotta.voter.ActiveVoter.TOPOLOGY_FETCH_TIME_PROPERTY;
+
+@ClusterDefinition(nodesPerStripe = 3, autoActivate = false)
+public class AttachCommandWithVoter1x3IT extends DynamicConfigIT {
+  private static final long TOPOLOGY_FETCH_INTERVAL = 11000L;
+
+  static {
+    System.setProperty(TOPOLOGY_FETCH_TIME_PROPERTY, "9000");
+  }
```
mathieucarbou, line 46: We should never have system properties statically set like that  in a test because it impacts the whole test JVM.
Can this property be set another way ?

PR #675, opened 2020-04-18
Angela Junit Rule (based on #678)

mathieucarbou:
Introducing an Angela Junit rule that could further be extracted

This rule is responsible for resource allocation (angela init, port allocation, etc) and resource cleanup related to angela plus adds several methods to access topology info.

Based on #678

@mobasherul : FYI, I had to fix a lot of unstable tests for consistency (timing issues or outcomes not taken in consideration)


mathieucarbou:
Included in #678

PR #676, opened 2020-04-18
[TDB-4843] Skip I/O when parsing data directories from XML

mathieucarbou:
This is a bugfix to avoid having directories created in the working directory when the migration CLI is parsing the xml files to convert them in a config repo


mathieucarbou:
Replaced by #677

PR #678, opened 2020-04-19
Dynamic config fixes

mathieucarbou:
Here is the list of fixes in this PR:

- [TDB-4843] Skip I/O when parsing data directories from XML
- [TDB-4953] convertor => converter 
- Concurrently run Galvan test (1 fork VM per core) 
- Increases build timeouts to 180 min
- Make LockingPortChooser a singleton
- Added ExtendedTestRule (this skeleton ensures exceptions thrown by after() do not hide test failures) and make use of ExtendedTestRule in Galvan (for platform), TmpDir and PortLockingRule.
- Fixing many problems in Galvan use of port locking:
  * reserved ports were not released
  * introduced an interface to control port allocation through the builder API, to allow people to use another kind of port locker
  * make sure the old port chooser is not used in BasicHarnessEntry
- Ensure Angela context is closed (no Ignite process is left opened), and other resources are cleaned when a test times out (https://github.com/Terracotta-OSS/angela/issues/29)
- Fixing unstable / non working consistency tests
- DynamicConfigIT refactoring to prepare for Angela Rule:
  * Renaming node names
  * Added final keyword where override is not possible
  * Fixed ConfigRepositoryGenerator to support stripeIds and port allocation
  * Fixing path that wre only Unix-like
- [TDB-4954] Introduce an Angela Junit rule that could further be extracted:
  * to control Angela lifecycle
  * to control Angela cleanup
  * to control port allocation and access node information


mathieucarbou:
@saurabhagas @mobasherul : this PR is ready to be merged :-) 
**ALL TESTS PASSED.** 
The Failure you see now is because Azure was not able to run the step _Attach log files to build_. So sadly Azure is failing. But if you look inside, the Maven build was successful.

PR #679, opened 2020-04-19
Making Angela use the lockingPortChooser mechanism

mathieucarbou:
**Rebased on #678**

**Depends on: https://github.com/Terracotta-OSS/angela/pull/42**

**Only look at last commit: https://github.com/Terracotta-OSS/terracotta-platform/pull/679/commits/bd7a3299ec101b27f4540a676baeb2793e2cc08a** (other commits are part of #678)

This PR updates the Angela Junit Rule so that it can provide ports by using the LockingPort mechanism that supports better running test in different forked JVM.

### DIFF
--- a/common/port-locking/src/main/java/org/terracotta/port_locking/LockingPortChooser.java
+++ b/common/port-locking/src/main/java/org/terracotta/port_locking/LockingPortChooser.java
@@ -24,7 +24,7 @@ public class LockingPortChooser {
     this.portLocker = portLocker;
   }
 
-  public MuxPortLock choosePorts(int portCount) {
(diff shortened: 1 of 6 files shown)


mathieucarbou:
@aurbroszniowski @saurabhagas : this PR works an allows the use of the port locking mechanism from tc-platform instead of Angela's default port chooser, by implementing our PortProvider interface. It supports different coordination of port reservation across forked JVM which we will need when running tests concurrently with Angela soon.

I will update it with the next Angela version with https://github.com/Terracotta-OSS/angela/pull/42 will be merged.

PR #680, opened 2020-04-20
tc-port-lock should be created with rw-rw-rw- so that other JVM spawn by other users can access the file in rw mode

mathieucarbou:
### THIS IS A FIX WE NEED TO AVOID RELEASE FAILURES OR BUILD FAILURES UNTIL WE HAVE A NEW PORT LOCKER MECHANISM.

Currently, some releases or build might fail because of file permissions issues.

### THIS PR HAS NOT THE GOAL OF FIXING ALL PERMISSIONS ISSUES WE HAVE (especially not in Windows) - BUT REDUCE THEM, AND AT LEAST ENSURE WE DO NOT HAVE RELEASE JOBS FAILING.

We can merge this quick fix to prevent that.
Merging it does not harm it helps us.
Not merging it makes us stick with these issues.

It is understood that it does not solve all the problems this port locker has. It just fixes some issues we run at the moment until we have a better port locker.

### DIFF
--- a/common/port-locking/src/main/java/org/terracotta/port_locking/GlobalFilePortLocker.java
+++ b/common/port-locking/src/main/java/org/terracotta/port_locking/GlobalFilePortLocker.java
@@ -23,13 +23,21 @@ import java.io.IOException;
 import java.io.RandomAccessFile;
 import java.nio.channels.FileChannel;
 import java.nio.channels.FileLock;
+import java.nio.file.FileAlreadyExistsException;
+import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
+import java.nio.file.attribute.PosixFilePermission;
+import java.util.Set;
+
+import static java.nio.file.attribute.PosixFilePermissions.asFileAttribute;
+import static java.nio.file.attribute.PosixFilePermissions.fromString;
 
 public class GlobalFilePortLocker implements PortLocker {
   private static final Logger LOGGER = LoggerFactory.getLogger(GlobalFilePortLocker.class);
   private static final String TEMP_DIRECTORY_PROPERTY = "java.io.tmpdir";
   private static final String PORT_LOCK_FILENAME = "tc-port-lock";
+  private static final Set<PosixFilePermission> PERMISSIONS = fromString("rw-rw-rw-");
 
   private final File portLockFile;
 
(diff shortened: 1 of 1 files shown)


common/port-locking/src/main/java/org/terracotta/port_locking/GlobalFilePortLocker.java
```diff
@@ -76,4 +79,21 @@ public PortLock tryLockPort(int port) {
       throw new PortLockingException("Error while trying to lock port", e);
     }
   }
+
+  private static File createOrReuse(Path portLockFilePath) throws IOException {
+    File file;
+    try {
+      try {
+        file = Files.createFile(portLockFilePath, asFileAttribute(PERMISSIONS)).toFile();
+      } catch (UnsupportedOperationException e) {
```
saurabhagas, line 88: Instead of this, we could check if setting POSIX permissions is supported

common/port-locking/src/main/java/org/terracotta/port_locking/GlobalFilePortLocker.java
```diff
@@ -76,4 +79,21 @@ public PortLock tryLockPort(int port) {
...
+        // in case we cannot atomically create and set these permissions / rwx not supported we retry to create normally
+        file = Files.createFile(portLockFilePath).toFile();
+      }
+      LOGGER.info("Created port lock file: {}", file.getAbsolutePath());
+    } catch (FileAlreadyExistsException e) {
+      file = portLockFilePath.toFile();
```
saurabhagas, line 94: What if the file exists but doesn't have the right permissions?

common/port-locking/src/main/java/org/terracotta/port_locking/GlobalFilePortLocker.java
mathieucarbou: This one is shorter:
Most of the cases, it will succeed. If it is not supported, it will be in all cases 1 operation done.
If we check the perms before, then most of the cases we will have an operation done for nothing so best case would be perm check + file creation.

common/port-locking/src/main/java/org/terracotta/port_locking/GlobalFilePortLocker.java
mathieucarbou: It is the current case, and it might be acceptable also: this will be a system where only a user will have rw access to the file.

Having the permissions set is a best effort but if they cannot be set, it will be like it was before.  So if another user tries to access the file from another user, it will fail, and this is OK.

What are the chances that we are running in some build environment that are not supporting POSIX file perms and where we also have different users accessing the same file ?

mathieucarbou:
@saurabhagas : FYI, `5.7.4-pre17` release in progress to have a version including this change

PR #681, opened 2020-04-21
Multi ignite cluster support

mathieucarbou:
Code updated for multi ignite support to run tests concurrently on separate JVM.

See: https://github.com/Terracotta-OSS/angela/pull/47

Note: do not merge yet, work in progress.

### Commands to run to test:
Fast build of the project:
`> ./mvnw clean install -DskipTests -Dfast -T4C`
Then run concurrently the tests in the diagnostic package (fastest tests) 
`> ./mvnw verify -f dynamic-config/testing/system-tests/ -Dit.test=org.terracotta.dynamic_config.system_tests.diagnostic.**`


pom.xml
```diff
@@ -39,7 +39,7 @@
     <terracotta-core.version>5.7.0-pre25</terracotta-core.version>
     <tripwire.version>1.0.0-pre7</tripwire.version>
     <galvan.version>1.5.0-pre3</galvan.version>
-    <angela.version>3.0.20</angela.version>
+    <angela.version>3.0.21-SNAPSHOT</angela.version>
```
mathieucarbou, line 42: @saurabhagas : need a new release with https://github.com/Terracotta-OSS/angela/pull/47

dynamic-config/testing/system-tests/pom.xml
```diff
@@ -160,6 +160,7 @@
         <groupId>org.terracotta</groupId>
         <artifactId>maven-forge-plugin</artifactId>
         <configuration>
+          <forkCount>1C</forkCount>
```
mathieucarbou, line 163: @saurabhagas : I wasn't able to test that on Azure... Since we have only 2 cores, it might create more slowdowns because 2 tests running concurrently, each test spawning ~4-6 processes, might be too hard.
So if it is too hard on Azure, we could comment and only use it locally.

mathieucarbou:
@saurabhagas : other problem: when running tests concurrently, there are some steps in Angela that are repeating for all the classes:  kit download and installation.
So on my computer, I have 8 angela tests starting to download and setup the things into `terracotta-platform/dynamic-config/testing/system-tests/target/platform-kit-5.7-SNAPSHOT`. So obviously i end up with a lot of errors due to corrupted jar files / missing classes, etc.

pom.xml
saurabhagas: Changed to `3.0.21`. @mobasherul will release an angela version once the angela PR is green.

mathieucarbou:
@saurabhagas @mobasherul : I have reverted the last commit to disable concurrent tests with Angela. I will open a new PR in Angela / tc-platform.
Let's merge and release that because this brings a lot of changes in tc-ee

saurabhagas:
All builds passed a while ago. Status is not refreshed here. Merging.

@mathieucarbou please don't release a new version yet though.

PR #682, opened 2020-04-21
Use files utility instead of Files.move as it tends to fails on Windows

saurabhagas:


mathieucarbou:
@saurabhagas :
- I had to rebase your commit : your branch was not rebsed
- I had to restart the build: the 2 windows builds were not finishing, even after 3 hours.

PR #683, opened 2020-04-22
Galvan tests were not running with <forkCount>1C</forkCount>

mathieucarbou:
`<forkCount>1C</forkCount>` was in the wrong xml section of the plugin 🙄

### DIFF
--- a/management/testing/integration-tests/pom.xml
+++ b/management/testing/integration-tests/pom.xml
@@ -175,8 +175,8 @@
         <groupId>org.terracotta</groupId>
         <artifactId>maven-forge-plugin</artifactId>
         <configuration>
+          <forkCount>1C</forkCount>
           <systemPropertyVariables>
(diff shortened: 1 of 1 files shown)


management/testing/integration-tests/pom.xml
```diff
@@ -175,8 +175,8 @@
         <groupId>org.terracotta</groupId>
         <artifactId>maven-forge-plugin</artifactId>
         <configuration>
+          <forkCount>1C</forkCount>
```
mathieucarbou, line 178: @saurabhagas : works very well locally, it speeds up a lot.
But since we have only 2 cores on Azure, it might create more slowdowns because 2 tests running concurrently, each test spawning ~2 processes, might be too hard and could cause more tests timeouts.
So if it is too hard on Azure, we could comment and only use it locally.

PR #689, opened 2020-04-27
Fixing validation in offheap and datadir change handlers

mathieucarbou:
Validation inside config handlers must be done against the topology that is upcoming. 
This topology also gets updated by the previous validated change in case of a multi setting nomad change.


mathieucarbou:
i am currently running all the tests locally...

PR #694, opened 2020-05-04
Fixing configuration validation for node hostname, port and name:

mathieucarbou:
- settings must all be eagerly resolved
- no placeholders should be allowed in a cluster configuration, since a machine (client or server) cannot resolve in place of each server
- missing validation for duplicated addresses
- missing validation for missing required settings


mathieucarbou:
@saurabhagas : could you please help me remember what was the use case you were testing already ? 
We saw a stack trace regarding the wrong  or missing validation around node hostname/port/name.
I would like to be sure it is handled.
Thanks!

saurabhagas:
> @saurabhagas : could you please help me remember what was the use case you were testing already ?
> We saw a stack trace regarding the wrong or missing validation around node hostname/port/name.
> I would like to be sure it is handled.
> Thanks!

@mathieucarbou I don't know what you're referring to..

saurabhagas:
Let's add the whole team (myself, you and @mobasherul) in all PRs/communications going forward. We all need to be aware of the problems and the fixes being proposed.

saurabhagas:
Declining this in favor of https://github.com/Terracotta-OSS/terracotta-platform/pull/695.

PR #698, opened 2020-05-04
Update sample config

jhouserizer:
Add a sample set of properties, and include it in the kit.

### DIFF
--- a/kit/src/main/assembly/platform-kit.xml
+++ b/kit/src/main/assembly/platform-kit.xml
@@ -32,7 +32,7 @@
       <outputDirectory>.</outputDirectory>
       <directory>${project.build.directory}/terracotta-kit-${terracotta-core.version}/terracotta-${terracotta-core.version}</directory>
       <excludes>
-        <exclude>server/conf/**</exclude>
+        <exclude>server/conf/**/*.xml</exclude>
         <exclude>server/plugins/lib/default-configuration-*.jar</exclude>
       </excludes>
     </fileSet>
--- a/kit/src/main/kit/server/conf/tc-server.properties
+++ b/kit/src/main/kit/server/conf/tc-server.properties
@@ -1,0 +1,145 @@
+#
+# Copyright Terracotta, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+# A set of basic configuration properties sufficient to run a simple 
+# single-node cluster
+#
+# This file will not get used unless it is specifically passed to the 
+# start-tc-server.sh script, upon which the properties here-in will be used to
+# configure and start a Terracotta server (after which that server's config
+# repository will contain the configuration, and this file should not be used
+# for future starts of that server)
+#
+# These properties could all be sent through config-tool, or directly on the 
+# start-tc-server command  line, but some users will find convenience in having
+# them in a file
+#
+# See documentation for a more complete list of properties and a more complete
+# description of them.
+
+
+# The name of the cluster the node will be part of
+cluster-name=tc-cluster
+
+# In the case of inter-node communication failures, should the cluster favor:
+#
+# - 'availability' of service (at the potential loss of data consistency)
+# - 'consistency' of data (at the potential loss of availability of service)
+#
+# ##### IMPORTANT: Storage use cases normally *require* 'consistency'  ######
+#                  Data loss can occur otherwise.
+#
+# Caching use cases may prefer 'availabilty'
+#
+# With the consistency option you specify how many additional (non-server node)
+# voters will participate in elections/quorum decisions. The number of nodes in 
+# a stripe + the number of additional voters should be an odd number.
+# So, for a two-node stripe you should run configure additional voter, i.e.:
+# failover-priority=consistency:1
+#
+failover-priority=availability
+
+
+# Configure the responsiveness of servers when detecting
+# unresponsive/disconnected clients.  Longer values are more tolerant of
+# transient delays in communication
+#client-lease-duration=15s
+
+# Indicates how much time (in seconds) a passive node taking over a failed 
+# active node will wait for pre-existing clients to reconnect.
+#client-reconnect-window=120s
+
+
+# (Enterprise edition only)
+# The directory containing security configuration artifacts
+# (certificates, credentials, etc.)
+#security-dir=/some/secure/path/terracotta-security
+
+# (Enterprise edition only)
+# The directory where security related audit logs will be written
+#security-audit-log-dir=/some/secure/path/terracotta-security--audit-logs
+
+# (Enterprise edition only)
+# The authentication mechanism to be used.  Possible options include:
+# 'file', 'ldap', 'certificate'
+#security-authc=file
+
+# (Enterprise edition only)
+# Activate SSL on listening ports
+#security-ssl-tls=false
+
+# (Enterprise edition only)
+# Activate the use of a file containing a white list of addresses (or address
+# ranges)
+#security-whitelist=false
+
+
+# A comma-separated list of offheap (in-memory) resourse pools in which data
+# data can be stored.
+# Entries are associations of a logical name for the resource to its size
+# in bytes (e.g. 'mypool:500MB,mystorage:250GB,mycaches:4GB')
+offheap-resources=main:500MB
+
+### Define stripes, and nodes within the stripes
+
+# ## Stripe 1, Node 1
+
+# The name to give the node
+stripe.1.node.1.node-name=default-node1
+
+# The RFC 1123 compliant hostname (or a valid IP address) of the node
+stripe.1.node.1.node-hostname=localhost
+
+# The address and port to bind the main communication port to
+# defaults are:
+#stripe.1.node.1.node-bind-address=0.0.0.0
+#stripe.1.node.1.node-port=9410
+
+# The address and port to bind the group communication port to
+# defaults are:
+#stripe.1.node.1.node-group-bind-address=0.0.0.0
+#stripe.1.node.1.node-group-port=9430
+
+# The directory in which to store node metadata (persistent state, etc.)
+stripe.1.node.1.node-metadata-dir=%H/terracotta/default-node1/metadata
+
+# A comma-separated list of directories in which stored data can be persisted.
+# Entries are mappings of a logical name for the directory to the path of the 
+# directory
+stripe.1.node.1.data-dirs=main:%H/terracotta/default-node1/user-data
+
+# The directory in which to write this node's logs
+stripe.1.node.1.node-log-dir=%H/terracotta/default-node1/logs
+
+# Base directory where backups will be placed
+#stripe.1.node.1.node-backup-dir=%H/terracotta/default-node1/backups
+
+
+# ## Stripe 1, Node 2
+
+#stripe.1.node.2.node-name=default-node2
+#stripe.1.node.2.node-hostname=myserver2.example.com
+#stripe.1.node.2.node-bind-address=0.0.0.0
+#stripe.1.node.2.node-port=9410
+#stripe.1.node.2.node-group-bind-address=0.0.0.0
+#stripe.1.node.2.node-group-port=9430
+        
+#stripe.1.node.2.node-metadata-dir=%H/terracotta/default-node2/metadata
+#stripe.1.node.2.data-dirs=main:%H/terracotta/default-node2/user-data
+#stripe.1.node.2.node-log-dir=%H/terracotta/default-node2/logs
+#stripe.1.node.2.node-backup-dir=%H/terracotta/default-node2/backups
+
+


kit/src/main/assembly/platform-kit.xml
```diff
@@ -32,7 +32,7 @@
       <outputDirectory>.</outputDirectory>
       <directory>${project.build.directory}/terracotta-kit-${terracotta-core.version}/terracotta-${terracotta-core.version}</directory>
       <excludes>
-        <exclude>server/conf/**</exclude>
+        <exclude>server/conf/**/*.xml</exclude>
```
jhouserizer, line 35: Can't entirely remove this because it comes in from core

chrisdennis:
Did we make a decision on how this properties file would get used? Are we happy for it to be a 'manual' addition to the launch command for the server?

kit/src/main/kit/server/conf/tc-server.properties
```diff
@@ -0,0 +1,128 @@
...
+# These properties could all be sent through config-tool, or directly on the 
+# start-tc-server command  line, but some users will find convenince in having 
+# them in a file
+#
+# See documentation for a more complete list of properties and a more complete
+# description of them.
+
+
+# The name of the cluster the node will be part of
+cluster-name=default-simple-cluster
```
mathieucarbou, line 20: careful: if we document to set it, the node(s) will start pre-activated so the user won't have a chance to build it's cluster (or multi-stripe cluster) remotely since the "diagnostic mode" is then by-passed.

kit/src/main/kit/server/conf/tc-server.properties
```diff
@@ -0,0 +1,128 @@
...
+# So, for a two-node stripe you should run one additional voter slot, e.g.: 
+# failover-priority=consistency:1
+#
+failover-priority=availability
+
+
+# Configure responsiveness of servers detecting unresponsive/disconnected 
+# clients.  Shorter values increase responsiveness, but longer values are 
+# more tolerant of transient delays in communication
+#client-lease-duration=20s
```
mathieucarbou, line 43: The default value is `150s`, not `20s`

kit/src/main/kit/server/conf/tc-server.properties
jhouserizer: The point is to have something that starts...

kit/src/main/kit/server/conf/tc-server.properties
jhouserizer: i went by what was in the node-tool asciidoc.  will fix this file.

kit/src/main/kit/server/conf/tc-server.properties
```diff
@@ -0,0 +1,128 @@
...
+
+# Indicates how much time (in seconds) a passive node taking over a failed 
+# active node will wait for pre-existing clients to reconnect.
+#client-reconnect-window=120s
+
+
+# The directory containing security configuration artifacts
+# (certificates, credentials, etc.)
+#security-dir=terracotta-security
```
mathieucarbou, line 52: for OSS, the way the config works, security and backup settings shouldn't be "seen" by the user because they are optional and if not used, won't be displayed in config.

So we could have 2 versions of this file: one for OSS, one fro EE, since OSS does not support backup & security.

kit/src/main/kit/server/conf/tc-server.properties
```diff
@@ -0,0 +1,128 @@
...
+# Activate the use of a file containing a white list of addresses (or address
+# ranges)
+#security-whitelist=false
+
+
+# A comma-separated list of offheap (in-memory) recourse pools in which data
+# data can be stored.
+# Entries are associations of a logical name for the resource to its size
+# in megabytes or gigabytes (e.g. 'mypool:500MB,mystorage:250GB,mycaches:4GB')
+offheap-resources=main:500MB
```
mathieucarbou, line 73: The default that was discussed to set 1 year ago is `main:512MB` (`Setting` class)

jhouserizer: yes, but this isn't documenting the default, as with some of the other properties. it is actually specified.

kit/src/main/kit/server/conf/tc-server.properties
```diff
@@ -0,0 +1,128 @@
...
+#stripe.1.node.1.node-bind-address=0.0.0.0
+#stripe.1.node.1.node-port=9410
+
+# The address and port to bind the group communication port to
+# defaults are:
+#stripe.1.node.1.node-group-bind-address=0.0.0.0
+#stripe.1.node.1.node-group-port=9430
+
+# Location of the node's configuration repository
+stripe.1.node.1.node-repository-dir=terracotta-config
```
mathieucarbou, line 96: `node-repository-dir` is not in the config. This is only a CLI option to point to a config directory, from where to read or save the config ;-) So cannot be in the config.

kit/src/main/kit/server/conf/tc-server.properties
```diff
@@ -0,0 +1,128 @@
...
+# The name of the cluster the node will be part of
+cluster-name=default-simple-cluster
+
+# In the case of node failures within the stripe, in the case that a 
+# split-brain could occur, should the cluster favor: 
+#
+# - 'availability' of service (at the cost of potential lost data consistency)
+# - 'consistency' of data (at the potential loss of availability of service)
+#
+# ##### IMPORTANT: Storage use cases require 'consistency'  ######
```
chrisdennis, line 28: `Require` might be too strong. `may` and `likely` are redundant.

kit/src/main/kit/server/conf/tc-server.properties
jhouserizer: I was just discovering that with some more testing.

kit/src/main/kit/server/conf/tc-server.properties
```diff
@@ -0,0 +1,128 @@
...
+# Caching use cases may likely prefer 'availabilty'
+#
+# With the consistency option you specify how many additional (non-server node)
+# voters will participate in elections/quorum decisions. The number of nodes in 
+# a stripe + the number of additional voters should be an odd number.
+# So, for a two-node stripe you should run one additional voter slot, e.g.:
```
chrisdennis, line 34: `slot` is unecessary.

chrisdennis, line 34: You mean `i.e.` not `e.g.`

kit/src/main/kit/server/conf/tc-server.properties
```diff
@@ -0,0 +1,128 @@
...
+# With the consistency option you specify how many additional (non-server node)
+# voters will participate in elections/quorum decisions. The number of nodes in 
+# a stripe + the number of additional voters should be an odd number.
+# So, for a two-node stripe you should run one additional voter slot, e.g.: 
+# failover-priority=consistency:1
+#
+failover-priority=availability
+
+
+# Configure responsiveness of servers detecting unresponsive/disconnected
```
chrisdennis, line 40: This feels awkward with the two instances of `responsiveness` and no `when`.

kit/src/main/kit/server/conf/tc-server.properties
```diff
@@ -0,0 +1,128 @@
...
+# The address and port to bind the group communication port to
+# defaults are:
+#stripe.1.node.1.node-group-bind-address=0.0.0.0
+#stripe.1.node.1.node-group-port=9430
+
+# Location of the node's configuration repository
+stripe.1.node.1.node-repository-dir=terracotta-config
+
+# The directory in which to store node metadata (persistent state, etc.)
+stripe.1.node.1.node-metadata-dir=terracotta-data
```
mathieucarbou, line 96: The default tha twas discussed to use it: `%H/terracotta/metadata` (`Setting` class)

But frankly, I personally  don't like the default we have right now... I would really prefer to have a default relative to the working directory, which is what you did.

Using %H/... for defaults is really bad since this puts folders at the same location for all the nodes on the same server.

This is something we could change in the Setting class, to have defaults relative to the working directory, and align them with this file.

WDYT @jhouserizer @saurabhagas ? Change is really fast to do.\ to get rid of the %H/... in our default values.

kit/src/main/kit/server/conf/tc-server.properties
jhouserizer: I actually started out with two properties files - one with just a few properties that starts, and one with examples of all properties.  the group decided it was better to conflate them, and have the "extras" commented out  (as this file has).

mathieucarbou: ok :-)

kit/src/main/kit/server/conf/tc-server.properties
jhouserizer: That seems to relate to the 'larger' kit conversation.   @chrisdennis  already has a simple properties file going into the ehcache kit.

kit/src/main/kit/server/conf/tc-server.properties
jhouserizer: I would prefer most a %T (or some other character) that means "relative to the tc home dir".

kit/src/main/kit/server/conf/tc-server.properties
jhouserizer: (removed with the latest commit)

kit/src/main/kit/server/conf/tc-server.properties
mathieucarbou: ok!

kit/src/main/kit/server/conf/tc-server.properties
mathieucarbou: good idea! and actually we have that! server scripts are started with the system prop: `tc.install-root` which points to `<kit-folder>/server`.

So that means we could have something like:

```
stripe.1.node.1.node-metadata-dir=%(tc.install-root)/data/server
stripe.1.node.1.data-dirs=main:%(tc.install-root)/data/user/main
stripe.1.node.1.log-directory=%(tc.install-root)/data/logs
...
```

mathieucarbou:
@jhouserizer : PR will need to be rebased ;-)

kit/src/main/kit/server/conf/tc-server.properties
```diff
@@ -0,0 +1,128 @@
...
+
+# Activate SSL on listening ports
+#security-ssl-tls=false
+
+# Activate the use of a file containing a white list of addresses (or address
+# ranges)
+#security-whitelist=false
+
+
+# A comma-separated list of offheap (in-memory) recourse pools in which data
```
chrisdennis, line 69: `resource`

kit/src/main/kit/server/conf/tc-server.properties
```diff
@@ -0,0 +1,128 @@
...
+# data can be stored.
+# Entries are associations of a logical name for the resource to its size
+# in megabytes or gigabytes (e.g. 'mypool:500MB,mystorage:250GB,mycaches:4GB')
```
chrisdennis, line 72: I think it can be pretty much any unit...

kit/src/main/kit/server/conf/tc-server.properties
jhouserizer: Seems like if we had the collection %H, %T, and %W   (home dir,  tc home, working dir) then we'd cover all bases of user's desires,  and %W probably ought to be default.

kit/src/main/kit/server/conf/tc-server.properties
mathieucarbou: yes, from B to PB ;-)

jhouserizer: Yes, but do I really want to go that detailed in the properties file?

kit/src/main/kit/server/conf/tc-server.properties
jhouserizer: woops!

kit/src/main/kit/server/conf/tc-server.properties
jhouserizer: seems like %(tc.install-root) is misnamed, since it's not the install root, but rather the root of the server stuff

mathieucarbou: I completely agree for the name! 
@myronkscott probably knows for what it is used and if we can change ii since it comes from core scripts.

jhouserizer: start-tc-server script currently does this:

```
    "${JAVA_HOME}/bin/java" $JAVA_COMMAND_ARGS -Xms256m -Xmx2049m \
	-XX:+HeapDumpOnOutOfMemoryError \
        -Dtc.install-root="${TC_SERVER_DIR}" \
        ${JAVA_OPTS} \
        -cp "${TC_SERVER_DIR}/lib/tc.jar:${TC_SERVER_DIR}/lib/:." \
        com.tc.server.TCServerMain "$@"
```

Is that a mistake what it does with "tc.install-root" ?  @mathieucarbou  @myronkscott @chrisdennis    ... or is it just a bad name?

jhouserizer: I guess it's been that way since 10.0 ... so it's just an unfortunate name.

mathieucarbou:
`[WARNING] Missing header in: /home/vsts/work/1/s/kit/src/main/kit/server/conf/tc-server.properties`

https://dev.azure.com/TerracottaCI/terracotta/_build/results?buildId=4483&view=logs&j=a9de3ca1-eec4-53fb-16f6-e5f9f7af9dd3&t=e00a0aaa-1564-5171-8d0e-0ec30f6b5e84&l=13653

mathieucarbou:
> @jhouserizer : PR will need to be rebased ;-)

was just saying that it needs to be rebased since I have merged Myon's PR today. Probably also commits fixed-up :P - `git fetch upstream && git rebase -i upstream/master`.

kit/src/main/kit/server/conf/tc-server.properties
```diff
@@ -0,0 +1,138 @@
...
+# active node will wait for pre-existing clients to reconnect.
+#client-reconnect-window=120s
+
+
+# The directory containing security configuration artifacts
+# (certificates, credentials, etc.)
+#security-dir=terracotta-security
+
+# The directory where security related audit logs will be written
+#security-audit-log-dir=terracotta-security-logs
```
saurabhagas, line 69: Could be named to "terracotta-audit-logs", and made to use a location other than the current working directory, as would be the case usually. (as security directory need to be access controlled because they have credentials etc.)

kit/src/main/kit/server/conf/tc-server.properties
```diff
@@ -0,0 +1,138 @@
...
+# Configure the responsiveness of servers when detecting
+# unresponsive/disconnected clients.  Longer values are more tolerant of
+# transient delays in communication
+#client-lease-duration=15s
+
+# Indicates how much time (in seconds) a passive node taking over a failed 
+# active node will wait for pre-existing clients to reconnect.
+#client-reconnect-window=120s
+
+
```
saurabhagas, line 63: I would keep all the security-related setting under its own section saying something like "The following section describes security-related settings".

Note that security (among other features) are EE-only, so we shouldn't keep them in the OSS kit (unless we want to tease the users, and tell them what they're missing out on).

If we skip security and other EE settings from this sample file (which we SHOULD), we'd need to have a similar config file in the EE kit, containing this section.

kit/src/main/kit/server/conf/tc-server.properties
```diff
@@ -0,0 +1,138 @@
...
+# ## Stripe 1, Node 2
+
+#stripe.1.node.2.node-name=default-node2
+#stripe.1.node.2.node-hostname=myserver2.example.com
+#stripe.1.node.2.node-bind-address=0.0.0.0
+#stripe.1.node.2.node-port=9410
+#stripe.1.node.2.node-group-bind-address=0.0.0.0
+#stripe.1.node.2.node-group-port=9430
+        
+#stripe.1.node.1.node-metadata-dir=%(tc.install-root)/default-node2/metadata
```
saurabhagas, line 133: This seems like a stray section, as all of these have been included above

kit/src/main/kit/server/conf/tc-server.properties
```diff
@@ -0,0 +1,138 @@
...
+stripe.1.node.1.data-dirs=main:%(tc.install-root)/default-node1/user-data
+
+# The directory in which to write this node's logs
+stripe.1.node.1.node-log-dir=%(tc.install-root)/default-node1/logs
+
+# Base directory where backups will be placed
+#stripe.1.node.1.node-backup-dir=%(tc.install-root)/default-node1/backups
+
+
+# ## Stripe 1, Node 2
```
saurabhagas, line 124: Could add a note that addition of a node adds HA to the system.

kit/src/main/kit/server/conf/tc-server.properties
```diff
@@ -0,0 +1,138 @@
...
+#stripe.1.node.1.node-bind-address=0.0.0.0
+#stripe.1.node.1.node-port=9410
+
+# The address and port to bind the group communication port to
+# defaults are:
+#stripe.1.node.1.node-group-bind-address=0.0.0.0
+#stripe.1.node.1.node-group-port=9430
+
+# The directory in which to store node metadata (persistent state, etc.)
+stripe.1.node.1.node-metadata-dir=%(tc.install-root)/default-node1/metadata
```
saurabhagas, line 110: This would need `tc.install-root` to be set right? I don't think we need this - saying `default-node1/metadata` suffices IMO

saurabhagas:
A few comments from my side. Some minor, others not so. When you update the PR, please squash the commits too. I'll have another pass at the file to see if it all makes sense.

kit/src/main/kit/server/conf/tc-server.properties
```diff
@@ -0,0 +1,138 @@
...
+# These properties could all be sent through config-tool, or directly on the 
+# start-tc-server command  line, but some users will find convenience in having
+# them in a file
+#
+# See documentation for a more complete list of properties and a more complete
+# description of them.
+
+
+# The name of the cluster the node will be part of
+cluster-name=default-simple-cluster
```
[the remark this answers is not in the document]
jhouserizer: ok!

kit/src/main/kit/server/conf/tc-server.properties
```diff
@@ -0,0 +1,138 @@
...
+
+# In the case of node failures within the stripe, should the cluster favor:
```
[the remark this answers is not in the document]
jhouserizer: good point!

kit/src/main/kit/server/conf/tc-server.properties
```diff
@@ -0,0 +1,138 @@
...
+#
+# - 'availability' of service (at the potential loss of data consistency)
+# - 'consistency' of data (at the potential loss of availability of service)
+#
+# ##### IMPORTANT: Storage use cases normally require 'consistency'  ######
```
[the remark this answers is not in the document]
jhouserizer: Suggestions?   I'm not trying to have the property file replace documentation, but trying to help users get by as they can, and know what they should (it's important) to look to docs for.

kit/src/main/kit/server/conf/tc-server.properties
myronkscott: that is a property that was used for TIMs.  That property is not currently used.

kit/src/main/kit/server/conf/tc-server.properties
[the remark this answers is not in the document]
saurabhagas: Something like "TC Server when being used for store use cases NEEDS TO have failover-priority set to consistency. A failover-priority configuration of availability will be deemed unsupported"?

kit/src/main/kit/server/conf/tc-server.properties
```diff
@@ -0,0 +1,138 @@
...
+# So, for a two-node stripe you should run configure additional voter, i.e.:
+# failover-priority=consistency:1
+#
+failover-priority=availability
+
+
+# Configure the responsiveness of servers when detecting
+# unresponsive/disconnected clients.  Longer values are more tolerant of
+# transient delays in communication
+#client-lease-duration=15s
```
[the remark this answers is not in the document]
jhouserizer: That's sort of the industry-wide practice for config files  (commented out defaults).

kit/src/main/kit/server/conf/tc-server.properties
jhouserizer: typo actually, these should be for node.2 ... thanks for catching.

kit/src/main/kit/server/conf/tc-server.properties
jhouserizer: %(tc.install-root)  is defined by the start script, so it exists.

I'm not sold on using it though.

kit/src/main/kit/server/conf/tc-server.properties
jhouserizer: so we could rename it  and / or change it to:  "-Dtc.install-root="${TC_SERVER_DIR}/.."  so that it is accurate (and then potentially be of use)?

jhouserizer: @myronkscott  ?

kit/src/main/kit/server/conf/tc-server.properties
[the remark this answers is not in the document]
saurabhagas: I was also pointing at the fact that default client-lease-duration is 150s. So, if we plan to use commented-out defaults, this needs to change.

kit/src/main/kit/server/conf/tc-server.properties
```diff
@@ -0,0 +1,145 @@
...
+#
+
+# A set of basic configuration properties sufficient to run a simple 
+# single-node cluster
+#
+# This file will not get used unless it is specifically passed to the 
+# start-tc-server.sh script, upon which the properties here-in will be used to
+# configure and start a Terracotta server (after which that server's config
+# repository will contain the configuration, and this file should not be used
+# for future starts of that server)
```
saurabhagas, line 24: Would be civil to end each sentence with a period ;-)

kit/src/main/kit/server/conf/tc-server.properties
```diff
@@ -0,0 +1,145 @@
...
+# unresponsive/disconnected clients.  Longer values are more tolerant of
+# transient delays in communication
+#client-lease-duration=15s
+
+# Indicates how much time (in seconds) a passive node taking over a failed 
+# active node will wait for pre-existing clients to reconnect.
+#client-reconnect-window=120s
+
+
+# (Enterprise edition only)
```
saurabhagas, line 66: This section could be moved at the end, since it's getting in the way of settings relevant for OSS

kit/src/main/kit/server/conf/tc-server.properties
```diff
@@ -0,0 +1,145 @@
...
+# Activate the use of a file containing a white list of addresses (or address
+# ranges)
+#security-whitelist=false
+
+
+# A comma-separated list of offheap (in-memory) resourse pools in which data
+# data can be stored.
+# Entries are associations of a logical name for the resource to its size
+# in bytes (e.g. 'mypool:500MB,mystorage:250GB,mycaches:4GB')
+offheap-resources=main:500MB
```
saurabhagas, line 94: The default is `main:512MB`

kit/src/main/kit/server/conf/tc-server.properties
```diff
@@ -0,0 +1,145 @@
...
+#stripe.1.node.1.node-bind-address=0.0.0.0
+#stripe.1.node.1.node-port=9410
+
+# The address and port to bind the group communication port to
+# defaults are:
+#stripe.1.node.1.node-group-bind-address=0.0.0.0
+#stripe.1.node.1.node-group-port=9430
+
+# The directory in which to store node metadata (persistent state, etc.)
+stripe.1.node.1.node-metadata-dir=%H/terracotta/default-node1/metadata
```
saurabhagas, line 117: These settings would need a revisit once the naming changes are merged.

kit/src/main/kit/server/conf/tc-server.properties
```diff
@@ -0,0 +1,145 @@
...
+# A comma-separated list of directories in which stored data can be persisted.
+# Entries are mappings of a logical name for the directory to the path of the 
+# directory
+stripe.1.node.1.data-dirs=main:%H/terracotta/default-node1/user-data
+
+# The directory in which to write this node's logs
+stripe.1.node.1.node-log-dir=%H/terracotta/default-node1/logs
+
+# Base directory where backups will be placed
+#stripe.1.node.1.node-backup-dir=%H/terracotta/default-node1/backups
```
saurabhagas, line 128: (Enterprise edition only)?

I would keep the EE settings together under a section called "Enterprise settings" and mention all EE settings there.

kit/src/main/kit/server/conf/tc-server.properties
```diff
@@ -0,0 +1,145 @@
...
+# - 'availability' of service (at the potential loss of data consistency)
+# - 'consistency' of data (at the potential loss of availability of service)
+#
+# ##### IMPORTANT: Storage use cases normally *require* 'consistency'  ######
+#                  Data loss can occur otherwise.
+#
+# Caching use cases may prefer 'availabilty'
+#
+# With the consistency option you specify how many additional (non-server node)
+# voters will participate in elections/quorum decisions. The number of nodes in
```
saurabhagas, line 48: election/quorum -> election?

I'd remove the lines which start from "the number of nodes in ...". We're documenting best-practices here, which we don't need to.

kit/src/main/kit/server/conf/tc-server.properties
jhouserizer: Civil schmivel.  :D

saurabhagas:
Needs to be updated now corresponding to the latest changes.

saurabhagas:
Included in https://github.com/Terracotta-OSS/terracotta-platform/pull/709

PR #699, opened 2020-05-04
Set default path relative to the working directory instead of %H/terracotta

mathieucarbou:
This PR follows some discussions from some months ago and here (https://github.com/Terracotta-OSS/terracotta-platform/pull/698#pullrequestreview-405265201).

Having the default values for paths related to the `%H/terracotta` folder causes pain such as:

- failures when **starting several nodes** on the same machine since directories easily overlap (issue TDB-4985)
- forcing to redefine these settings in tests **ALWAYS** and if we forget one, we might end up with some test failures because 2 tests will use or re-use the same folders. **These test issues are really hard to troubleshoot.**
- this is a wrong default because our samples are focusing on **not using %H**
- using %H might lead to the server **reusing any old / crappy folder** that the user might have and this might fail.

We could also have these properties relative to the kit home folder (`%(tc.install-root)/..`) but:
- the name is not so good
- this would create an absolute path - so conflicts could still happen if starting 2 nodes from the same kit folder
- having data inside a kit installation is not so good.

So with this PR, a terracotta folder will be created inside the working directory (instead of %H) and the layout will be:

```
terracotta
   |---repository
   |--- logs
   |--- metadata
   |--- user-data
            |--- main
            |--- second
```


jhouserizer:
:(  I'm sad this was already merged.

We had some good team discussion about this today, with some some reasoning about why this shouldn't (quite) be this way.

What we (~12 of us in the leads meeting) decided:

1) If nothing is specified (defaults are used), then creating/using a 'terracotta' directory off of the user's home directory (with appropriate sub-directories for metadata, logs, repository, etc.) should be used.   Default if nothing specified is off of home dir (but organized under one dir).

2) If relative paths are used when specifying paths, then they should be relative to the working directory

jhouserizer:
Also "repository" is vague, why not "config-repository" or similar?  Similarly "node-repository-dir" is also rather mysterious and vague sounding, but probably that ship has sailed.

start-tc-server.sh needs to be pointed to the config to use.   telling users to point it to a "repository" or a "node-repository"  isn't as helpful (in helping them understand the concept of what they're doing) as telling them to point to a  "config-repository".

mathieucarbou:
>     1. terracotta

This is the behavior we had before @jhouserizer . **But people were complaining about this behavior for the reasons cited in the PR description.** We had an issue opened for a while to find a solution for that. Plus we had a lot of problems during tests, manual tryouts, and so on, that people / client will have when testing the new KIT. 

This is fine for me to revert. I don't mind which solution we choose. What I don't mind though, is the change of direction and the fact that we are driven into different ways. I would like to point that that people were complaining about the solution we had before in the demo.

So this kind of endless discussion needs to be settled once and for all and we need to accept the pros and cons for the solution we chose.

Here is my opinion (and probably @saurabhagas will share it since we encountered these issues with using absolute paths as defaults):

> We are changing the way the server is configured and started in dynamic config. There is no point in sticking with what we had before which was more adapted to a static solution. More flexibility in configuration and startup (which we have now) will obviously lead to more errors / mistakes from he user if we default to absolute directories instead of relative ones. Since the long time we are in the project and use it, I am confident this choice might lead to more problematic feedbacks from users, because we experiences such issues too when using absolute path by default.

mathieucarbou:
> Also "repository" (as a directory name) is vague, why not "config-repository" or similar? Similarly "node-repository-dir" is also rather mysterious and vague sounding, but probably that ship has sailed.
> 
> start-tc-server.sh needs to be pointed to the config to use. telling users to point it to a "repository" or a "node-repository" isn't as helpful (in helping them understand the concept of what they're doing) as telling them to point to a "config-repository".

yes, completely agree with you. this name and concept was introduced way before I was in the project. To me, this is something that wasn't clear also for me but I get used to it.

This is not too late to rename though... We just need to know whether we do it or not.

FYI, the name repository probably comes from that fact that inside this folder there is a `conf` folder, `sanskrit` folder and `license` folder. `conf` contains the cluster config so we could have something like:

```
terracotta
  - configuration (instead of repository)
    - sanskirt
    - cluster
    - license
```

PR #700, opened 2020-05-04
[BUGFIX] Prevent failure when concurrent forked JVM create the target/galvan folder

mathieucarbou:
You could end up with Windows builds failing with this error:

```
java.lang.IllegalArgumentException: Cluster directory could not be created: target\galvan
	at org.terracotta.testing.rules.BasicExternalCluster.<init>(BasicExternalCluster.java:96)
	at org.terracotta.testing.rules.BasicExternalClusterBuilder.build(BasicExternalClusterBuilder.java:155)
	at org.terracotta.management.integration.tests.AbstractSingleTest.<init>(AbstractSingleTest.java:46)
	at org.terracotta.management.integration.tests.ClientCacheRemoteManagementIT.<init>(ClientCacheRemoteManagementIT.java:34)
```

Example: https://dev.azure.com/TerracottaCI/terracotta/_build/results?buildId=4484&view=ms.vss-test-web.build-test-results-tab&runId=1009560&resultId=100872&paneView=debug

### DIFF
--- a/galvan-platform-support/src/main/java/org/terracotta/testing/rules/BasicExternalCluster.java
+++ b/galvan-platform-support/src/main/java/org/terracotta/testing/rules/BasicExternalCluster.java
@@ -92,7 +92,7 @@ class BasicExternalCluster extends Cluster {
       }
     } else {
       boolean didCreateDirectories = clusterDirectory.toFile().mkdirs();
-      if (!didCreateDirectories) {
+      if (!didCreateDirectories && !Files.exists(clusterDirectory)) {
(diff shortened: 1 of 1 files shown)


mathieucarbou:
> I'm sure that this will work, but would [FileUtils](http://commons.apache.org/proper/commons-io/javadocs/api-2.5/org/apache/commons/io/FileUtils.html#forceMkdir(java.io.File)) be cleaner?

we don't want to force the creation, just ensure that the directory exist already, or has been created.
If the same operation is done concurrently by different VM, we need to double check since creation and check can happen concurrently.

PR #701, opened 2020-05-05
Small IT test fixes + version bump-ups

myronkscott:
tc-core version bump.

PR #703, opened 2020-05-06
Naming change for configuration directory

mathieucarbou:
### As discussed in chatroom

**Naming change for configuration directory:**

- CLI: `--node-repository-dir=%H/terracotta/repository` => `--node-config-dir=%H/terracotta/config`

- File structure becomes:

```
%H/terracotta/config
   |-- cluster
   |     |-- node1.1.properties
   |     |-- node1.2.properties
   |-- changes
   |     |-- lock
   |     |-- append.log
   |-- license
   |     |-- licence.xml
```

instead of:

```
%H/terracotta/repository
   |-- config
   |     |-- node1.1.properties
   |     |-- node1.2.properties
   |-- sanskrit
   |     |-- lock
   |     |-- append.log
   |-- license
   |     |-- licence.xml
```


saurabhagas:
I'd prefer using `history` instead of `changes`.

dynamic-config/cli/config-converter/src/main/java/org/terracotta/dynamic_config/cli/config_converter/ConversionFormat.java
```diff
@@ -18,7 +18,7 @@
 import com.beust.jcommander.converters.EnumConverter;
 
 public enum ConversionFormat {
-  REPOSITORY,
+  DIRECTORY,
```
saurabhagas, line 21: Don't quite like it. "Directory" is just vague. How about "config-dir" or "config-directory", or "configuration-directory"?

jhouserizer:
> The downside of calling it "config directory" is that users might think it's a place they have to go to to change configuration. "Repository" was good in the sense that, it was a store house of information, which the user didn't have to worry about, until things went wrong.
> 
> Also, I'd prefer using `history` instead of `changes`.
> 
> But if it's agreed upon with the bigger audience, I'm okay with keeping these changes. After all, there are only two hard things in computer science - naming and cache invalidation, and we do both :-D



> users might think it's a place they have to go to to change configuration

Good point.  I was actually pondering do we need to add a file to the directory named "DO_NOT_EDIT_THESE_FILES.txt" or something.  

We may want to do that even if we change the name of the folder, and the user can name the folder anything they want if they don't go with the default.

How about naming the folder "internal-config"

mathieucarbou:
> The downside of calling it "config directory" is that users might think it's a place they have to go to to change configuration. "Repository" was good in the sense that, it was a store house of information, which the user didn't have to worry about, until things went wrong.
> 
> Also, I'd prefer using `history` instead of `changes`.
> 
> But if it's agreed upon with the bigger audience, I'm okay with keeping these changes. After all, there are only two hard things in computer science - naming and cache invalidation, and we do both :-D

Personally, I would prefer not going over that again if we can stick with what we have now...

mathieucarbou:
> DO_NOT_EDIT_THESE_FILES

Why ?

Is the user expected to play into any data directory ? Also, oes the user expected to play into the server metadata directory ? Or put something into the logs directory ?

Why this folder would be special compared to other ones ?

mathieucarbou:
PR rebased

dynamic-config/cli/config-converter/src/main/java/org/terracotta/dynamic_config/cli/config_converter/ConversionFormat.java
mathieucarbou: because this is a format.

I would return the question then. Why we didn't use config-properties, or config-props ?

I have kept the same logic: properties or directory are just different format outputs for the converted configuration.

mathieucarbou:
@saurabhagas @jhouserizer : I will let you edit this PR and continue it once the naming will have decided. Clearly, this is not settled as I thought it was yesterday after our discussion. 
So I will stop spending time on that,

dynamic-config/cli/config-converter/src/main/java/org/terracotta/dynamic_config/cli/config_converter/ConversionFormat.java
mathieucarbou: Addition: this change to me is not part of this PR.

This PR goal is not remove the repository "naming".

If we want to change the output format descriptions, we can raise another PR after this one, or add anotehr commit.

jhouserizer:
I think the current stat of this if fine for merge (if checks pass, etc.).

We're really into the weeds now with talks of further change.

However, i think i will raise a request to get the contents of the cluster folder (the properties files) to have a header comment in them that warns the user not to attempt editing the file.

mathieucarbou:
> I think the current state of this if fine for merge (if checks pass, etc.).
> 
> We're really into the weeds now with talks of further change.
> 
> However, i think i will raise a request to get the contents of the cluster folder (the properties files) to have a header comment in them that warns the user not to attempt editing the file.

yes! this is totally feasible, we already supports having a header but we leaves it to blank at the moment. @saurabhagas will sync with you and @GaryWKeim for these points and the pre-activation flags we would want to add :-)

saurabhagas:
@myronkscott your changes from #701 included here

saurabhagas:
@mathieucarbou fixed a failing test (in config conversion tool, added just before your PR). Also added 2 commits from Myron's PR to avoid having to rebase

mathieucarbou:
> @mathieucarbou fixed a failing test (in config conversion tool, added just before your PR). Also added 2 commits from Myron's PR to avoid having to rebase

Removed added commits because build is not passing with these additions like in PR #701.

PR #707, opened 2020-05-07
[URG ] Angela fixes + converter test fix

mathieucarbou:
- use Angela 3.0.25 (fixes PortManager introduction) - DO NOT USE v3.0.24
- Close NomadClient 
- Fix convertor code test

PR #708, opened 2020-05-07
Server logging

myronkscott:


mathieucarbou:
@myronkscott : with @saurabhagas we saw some issues in a converter test + converter code that was committed (nomad client not closed, so lock not released), which is making a test fail on Windows.

All is fixes in #707: angela + these fixes.

I suggest you rebase your PR on my branch. If you do so and yours pass, we can merge everything together, which will speed up the process. Otherwise you will have to rebase again...

PR #709, opened 2020-05-07
Fixes

mathieucarbou:
1. This reverts commit a13e04b6fb43024a18e80dae445852594cdc4b2c. This commit wrongly appeared in @myronkscott's PR after he rebased on #707
2. Add a warning in properties file inside config repository
3. Add config properties sample

### DIFF
--- a/dynamic-config/server/configuration-repository/src/main/java/org/terracotta/dynamic_config/server/configuration/nomad/persistence/FileConfigStorage.java
+++ b/dynamic-config/server/configuration-repository/src/main/java/org/terracotta/dynamic_config/server/configuration/nomad/persistence/FileConfigStorage.java
@@ -86,7 +86,8 @@ public class FileConfigStorage implements ConfigStorage<NodeContext> {
       nonDefaults.setProperty("this.name", String.valueOf(config.getNodeName()));
 
       StringWriter out = new StringWriter();
-      Props.store(out, nonDefaults, "Configuration for node '" + config.getNodeName() + "' in stripe ID " + config.getStripeId());
+      String comments = "THIS FILE IS INTENDED FOR BOOK-KEEPING PURPOSES ONLY, AND IS NOT SUPPOSED TO BE EDITED. DO NOT ATTEMPT TO MODIFY.";
+      Props.store(out, nonDefaults, comments);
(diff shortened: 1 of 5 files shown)


dynamic-config/server/configuration-repository/src/main/java/org/terracotta/dynamic_config/server/configuration/nomad/persistence/FileConfigStorage.java
```diff
@@ -86,7 +86,8 @@ public void saveConfig(long version, NodeContext config) throws ConfigStorageExc
       nonDefaults.setProperty("this.name", String.valueOf(config.getNodeName()));
 
       StringWriter out = new StringWriter();
-      Props.store(out, nonDefaults, "Configuration for node '" + config.getNodeName() + "' in stripe ID " + config.getStripeId());
+      String comments = "THIS FILE IS INTENDED FOR BOOK-KEEPING PURPOSES ONLY, AND IS NOT SUPPOSED TO BE EDITED. DO NOT ATTEMPT TO MODIFY.";
```
saurabhagas, line 89: Apparently, there's no way to add multiple-comments using the `Properties::store` API. The stripe ID and node name are already included in the file so it didn't feel wrong to replace it with this warning message.

saurabhagas:
@mathieucarbou rebased the PR. Added a commit.

dynamic-config/server/configuration-repository/src/main/java/org/terracotta/dynamic_config/server/configuration/nomad/persistence/FileConfigStorage.java
mathieucarbou: Yes thsi is right. The stripe id and node name and id were added after: this config repo file is an "enhanced" version of the cluster properties, which also contains the necessary info to recreate the node context. The comment was there before these fields were added if I remember.

saurabhagas: Yeah, that's what I recall as well.

PR #716, opened 2020-05-11
TDB-4990: Refactoring of json wiring in dynamic config

mathieucarbou:
- **re-enabling `-Werror`**

- **Removed all Jackson annotations from model classes that are used on client classpath**

- **Removed all Nomad related classes from client classpath**

- **Removed all json-related modules from client classpath (common-json-support)**

- **Removed any jackson module scanning (classgraph or jackson module discovery)**

- **Removed statically accessed Json class**

- **Module SPLIT**: `-model` == topology classes VS `-api` == Nomad classes and services. Firs tone is used by topology entity and will in client classpath, the second one is the DC api used for CLI, plugins and services

- Each module now provides (**in provided scope**) an isolated jackson module that is only used for the consumers that need to know how to map the classes in these modules

- Correctly configured compiler flags on root pom with `-Xlint:all -Xlint:-path -Werror` (PR sent in parent pom: https://github.com/Terracotta-OSS/terracotta-parent/pull/15). `-Xlint:-path` is required because of the manifest entries generated, which point to jars on the same folders (required for plugins/api and plugins/lib, but creates compiler warnings)

**Tested the downstream EE project with this PR: KIT builds fine, with `-Werror`, and no jackson libs or Nomad classes leak in client classpath**


mathieucarbou:
@saurabhagas @mobasherul : PR ready to be reviewed & merged. EE PR is already sent also. We just need a release train.

saurabhagas:
Don't understand the need for the new `Substitutor` class - it just checks if the input contains a substitution parameter, and doesn't substitute anything. If we do need this class, we could rename it to `SubstitutionChecker` or something along those lines.

saurabhagas:
I'm not wondering that if nomad is removed from the client-classpath, did we even need to remove the annotations?

saurabhagas:
I merged the PR. The questions can be addressed later. Not starting a release train because there's no need for another release currently.

mathieucarbou:
> Don't understand the need for the new `Substitutor` class - it just checks if the input contains a substitution parameter, and doesn't substitute anything. If we do need this class, we could rename it to `SubstitutionChecker` or something along those lines.

to separate this check from the interface. The interface is a service - in api module. The check utility class is used by the model classes.

- api module won't go in clients
- model will go in client's classpath

mathieucarbou:
> I'm not wondering that if nomad is removed from the client-classpath, did we even need to remove the annotations?

nomad is not the only use of jackson. The diagnostic layer makes use of jackson.

Also, the removal of the annotations from the model classes is necessary as explained already to get rid of the x:lint warnings and be able to use -Werror.

mathieucarbou:
> I merged the PR. The questions can be addressed later. Not starting a release train because there's no need for another release currently.

Thansk for the merge. 
Yes there is the need... EE PR is waiting for these changes.
I will raise an Ehcache PR.

PR #717, opened 2020-05-11
Tweaked test timeouts

saurabhagas:
- Individual operation timeout should not be the same as global test timeout - introduced a per-operation timeout
- Use the operation timeout for config tool and diagnostic connections

### DIFF
--- a/dynamic-config/testing/support/src/main/java/org/terracotta/dynamic_config/test_support/DynamicConfigIT.java
+++ b/dynamic-config/testing/support/src/main/java/org/terracotta/dynamic_config/test_support/DynamicConfigIT.java
@@ -23,6 +23,7 @@ import org.junit.runner.Description;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.terracotta.angela.client.config.ConfigurationContext;
+import org.terracotta.angela.client.config.TsaConfigurationContext;
 import org.terracotta.angela.client.filesystem.RemoteFolder;
 import org.terracotta.angela.client.support.junit.AngelaRule;
 import org.terracotta.angela.client.support.junit.NodeOutputRule;
@@ -33,6 +34,8 @@ import org.terracotta.angela.common.dynamic_cluster.Stripe;
 import org.terracotta.angela.common.tcconfig.License;
 import org.terracotta.angela.common.tcconfig.TerracottaServer;
 import org.terracotta.angela.common.topology.Topology;
+import org.terracotta.common.struct.Measure;
+import org.terracotta.common.struct.TimeUnit;
 import org.terracotta.diagnostic.client.DiagnosticService;
 import org.terracotta.diagnostic.client.DiagnosticServiceFactory;
 import org.terracotta.dynamic_config.api.model.Cluster;
@@ -57,6 +60,8 @@ import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.time.Duration;
+import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.List;
 import java.util.OptionalInt;
 import java.util.Properties;
@@ -68,7 +73,6 @@ import static java.util.stream.IntStream.rangeClosed;
 import static org.hamcrest.Matchers.equalTo;
 import static org.hamcrest.Matchers.is;
 import static org.junit.Assert.assertThat;
-import org.terracotta.angela.client.config.TsaConfigurationContext;
 import static org.terracotta.angela.client.config.custom.CustomConfigurationContext.customConfigurationContext;
 import static org.terracotta.angela.client.support.hamcrest.AngelaMatchers.successful;
 import static org.terracotta.angela.common.AngelaProperties.DISTRIBUTION;
@@ -87,9 +91,12 @@ import static org.terracotta.utilities.io.Files.ExtendedOption.RECURSIVE;
 import static org.terracotta.utilities.test.matchers.Eventually.within;
 
 public class DynamicConfigIT {
-  private static final Logger LOGGER = LoggerFactory.getLogger(DynamicConfigIT.class);
-
   protected static final boolean WIN = System.getProperty("os.name").toLowerCase().startsWith("windows");
+
+  private static final Logger LOGGER = LoggerFactory.getLogger(DynamicConfigIT.class);
+  private static final Duration DEFAULT_TEST_TIMEOUT = Duration.ofMinutes(2);
+  private static final Duration CONN_TIMEOUT = Duration.ofSeconds(30);
+  private static final Duration ASSERT_TIMEOUT = Duration.ofMinutes(1);
 
   protected final TmpDir tmpDir;
   protected final AngelaRule angela;
@@ -98,7 +105,7 @@ public class DynamicConfigIT {
   @Rule public RuleChain rules;
 
   public DynamicConfigIT() {
-    this(Duration.ofSeconds(120));
+    this(DEFAULT_TEST_TIMEOUT);
   }
 
   public DynamicConfigIT(Duration testTimeout) {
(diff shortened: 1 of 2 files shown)


dynamic-config/testing/support/src/main/java/org/terracotta/dynamic_config/test_support/DynamicConfigIT.java
```diff
@@ -358,19 +372,19 @@ private Properties generateProperties() {
...
   }
 
   protected final void waitUntil(NodeOutputRule.NodeLog result, Matcher<NodeOutputRule.NodeLog> matcher) {
-    waitUntil(() -> result, matcher, timeout);
+    waitUntil(() -> result, matcher, OPERATION_TIMEOUT);
   }
 
   protected final <T> void waitUntil(Supplier<T> callable, Matcher<T> matcher) {
-    waitUntil(callable, matcher, timeout);
+    waitUntil(callable, matcher, OPERATION_TIMEOUT);
```
mathieucarbou, line 383: it cannot be a constant: it needs to be a variable so that it can be changed per class and subclass also i nEE

dynamic-config/testing/support/src/main/java/org/terracotta/dynamic_config/test_support/DynamicConfigIT.java
```diff
@@ -217,7 +223,15 @@ protected ConfigToolExecutionResult activateCluster(String name) throws TimeoutE
   }
 
   protected ConfigToolExecutionResult configToolInvocation(String... cli) {
-    return angela.tsa().configTool(getNode(1, 1)).executeCommand(cli);
+    ConfigTool configTool = angela.tsa().configTool(getNode(1, 1));
+    if (!Arrays.asList(cli).contains("-t")) {
+      String[] enhancedCli = new String[cli.length + 2];
+      enhancedCli[0] = "-t";
```
mathieucarbou, line 229: operation timeout is `-r`, not `-t`. `-t` is connection timeout. For `-t`, we can re-used the existing `timeout` variable if you need to set one

saurabhagas:
@mathieucarbou comments handled. PR good to go.

dynamic-config/testing/support/src/main/java/org/terracotta/dynamic_config/test_support/DynamicConfigIT.java
```diff
@@ -217,7 +226,48 @@ protected ConfigToolExecutionResult activateCluster(String name) throws TimeoutE
   }
 
   protected ConfigToolExecutionResult configToolInvocation(String... cli) {
-    return angela.tsa().configTool(getNode(1, 1)).executeCommand(cli);
+    List<String> enhancedCli = new ArrayList<>(cli.length);
+    List<String> configToolOptions = getConfigToolOptions(cli);
+
+    boolean addedOptions = false;
```
saurabhagas, line 232: Options like `-r` can be present multiple times in the passed `cli` param - one for the request timeout, and the other for specific-command options (e.g. `get` command has a `-r` option). So, the idea is to check if something like `-r` was passed as the config tool option (not the command option), and add it only if it wasn't.

dynamic-config/testing/support/src/main/java/org/terracotta/dynamic_config/test_support/DynamicConfigIT.java
```diff
@@ -217,7 +226,48 @@ protected ConfigToolExecutionResult activateCluster(String name) throws TimeoutE
...
+    String timeout = Measure.of(OP_TIMEOUT.getSeconds(), TimeUnit.SECONDS).toString();
```
mathieucarbou, line 233: I really don't like the use of hard-coded non extensible constants here... Subclasses SHOULD be able to control that.

This could be replaced at least by:

`String timeout = getOperationTimeout().toString();`

and have a 

`protected Measure<TimeUnit> getOperationTimeout() { return OP_TIMEOUT; }` in DynamicCOnfgiIT

dynamic-config/testing/support/src/main/java/org/terracotta/dynamic_config/test_support/DynamicConfigIT.java
mathieucarbou: ok... this becomes complex and hacky :-(

dynamic-config/testing/support/src/main/java/org/terracotta/dynamic_config/test_support/DynamicConfigIT.java
```diff
@@ -358,19 +408,19 @@ private Properties generateProperties() {
   // =========================================
 
   protected final void waitUntil(ConfigToolExecutionResult result, Matcher<ConfigToolExecutionResult> matcher) {
-    waitUntil(() -> result, matcher, timeout);
+    waitUntil(() -> result, matcher, OP_TIMEOUT);
```
mathieucarbou, line 411: use an overridable `getOperationTimeout()` method instead

dynamic-config/testing/support/src/main/java/org/terracotta/dynamic_config/test_support/DynamicConfigIT.java
```diff
@@ -358,19 +408,19 @@ private Properties generateProperties() {
...
   }
 
   protected final void waitUntil(NodeOutputRule.NodeLog result, Matcher<NodeOutputRule.NodeLog> matcher) {
-    waitUntil(() -> result, matcher, timeout);
+    waitUntil(() -> result, matcher, OP_TIMEOUT);
```
mathieucarbou, line 415: use an overridable `getOperationTimeout()` method instead

dynamic-config/testing/support/src/main/java/org/terracotta/dynamic_config/test_support/DynamicConfigIT.java
```diff
@@ -358,19 +408,19 @@ private Properties generateProperties() {
...
   }
 
   protected final <T> void waitUntil(Supplier<T> callable, Matcher<T> matcher) {
-    waitUntil(callable, matcher, timeout);
+    waitUntil(callable, matcher, OP_TIMEOUT);
```
mathieucarbou, line 419: use an overridable `getOperationTimeout()` method instead

dynamic-config/testing/support/src/main/java/org/terracotta/dynamic_config/test_support/DynamicConfigIT.java
```diff
@@ -414,8 +464,8 @@ protected final Cluster getUpcomingCluster(String host, int port) throws Excepti
     try (DiagnosticService diagnosticService = DiagnosticServiceFactory.fetch(
         InetSocketAddress.createUnresolved(host, port),
         getClass().getSimpleName(),
-        Duration.ofSeconds(30),
-        Duration.ofSeconds(30),
+        OP_TIMEOUT,
+        OP_TIMEOUT,
```
mathieucarbou, line 468: use an overridable `getOperationTimeout()` method instead

dynamic-config/testing/support/src/main/java/org/terracotta/dynamic_config/test_support/DynamicConfigIT.java
```diff
@@ -429,8 +479,8 @@ protected final Cluster getRuntimeCluster(String host, int port) throws Exceptio
...
```
mathieucarbou, line 483: use an overridable `getOperationTimeout()` method instead

dynamic-config/testing/support/src/main/java/org/terracotta/dynamic_config/test_support/DynamicConfigIT.java
```diff
@@ -444,8 +494,8 @@ protected final void withTopologyService(String host, int port, Consumer<Topolog
...
```
mathieucarbou, line 498: use an overridable `getOperationTimeout()` method instead

dynamic-config/testing/support/src/main/java/org/terracotta/dynamic_config/test_support/DynamicConfigIT.java
saurabhagas: We already have a flavor which takes a `Duration` argument. I don't see the need for `OP_TIMEOUT` to be explicitly overridden - if someone wants a custom timeout, they can use that flavor.

dynamic-config/testing/support/src/main/java/org/terracotta/dynamic_config/test_support/DynamicConfigIT.java
saurabhagas: Done everywhere.

dynamic-config/testing/support/src/main/java/org/terracotta/dynamic_config/test_support/DynamicConfigIT.java
saurabhagas: Anyway, done everywhere.

mathieucarbou: > We already have a flavor which takes a `Duration` argument. I don't see the need for `OP_TIMEOUT` to be explicitly overridden - if someone wants a custom timeout, they can use that flavor.

Well, if a sub-test need to use a bigger, would you rather have all the +20 waituntil assertions using another parameter and updated, which makes the assertion reading bad, a lot of changes, or you would prefer to give the ability to the class to just override 1 method, to set the timeout, and use normal assertions at every place normally ?

mathieucarbou:
@saurabhagas : PR rebased

PR #719, opened 2020-05-11
Separate stdout vs stderr logging in tools 

### LINKED ISSUE #718: config-tool command output is not scripting friendly
We should make sure to separate 'data' from 'errors and/or diagnostics' in the output of the `config-tool`. This is particularly important if someone tries to use the output of a `get` or `export` command piped directly to another command.

mathieucarbou:


mathieucarbou:
> It seems odd to force the sys-out through a logger for minimal benefit. One false move in the logging setup and we can end up breaking the output of all the commands. It would seem to me that the following would be a safer and more easily understood setup.
> 
>     1. Configure a single appender that pushes all logging output to `System.err`
> 
>     2. Write all "command output" directly to System.out bypassing the logging framework completely.
> 
>     3. Have `--verbose` simply modify the thresholds for the `System.err` logging.
> 
> 
> This greatly simplifies the setup and gives the nice benefit of 'protecting' the `System.out` of the process from munging by the logger and from pollution when you turn on `--verbose`.

The current solution does what you describe, and much more, since it allows also to have different logging patterns. This is also the solution used at Tc since years, we didn't reinvent the wheel : same pattern to go through a logging for the tools was there.

I agree with you though that it does not provide a clean API separation for the developers between logging and command output.

This PR solves the issue you have described, but perhaps you can open another one so tha twe change the way logging and outputting is done in all the terracotta tools (this is a broader change)

mathieucarbou:
@saurabhagas @chrisdennis : I am pausing this issue because a lot of tests are failing. This is caused by an angela issue (https://github.com/Terracotta-OSS/angela/issues/17): angela is not correctly redirecting the err stream of the processes. So we cannot correctly assert on stderr. So all tests that were doing some assertions based on the stdout output, that now is sent to stderr, are consequently failing.

mathieucarbou:
@chrisdennis @saurabhagas : PR should pass now.

- I have reverted back to the behavior we had, to log everything to `System.out` through logback
- The get/export command shouldn't display any info logs outside of the ones for the output

This should fix #718 so that we can pipe get and export.

I think we can merge this, and open other issues regarding the other points raised by @chrisdennis :

1. Clean API separation between command output and logging. This apply not only to config tool but all tools.
2. Separation between stdout/stderr and support it in Angela

mathieucarbou:
> Although this does fix the problem with the polluted output of the get and export commands I'm not sure it's worth tearing this issue in half. I think I'd rather see us just fix the whole damn thing properly in one shot by decidubg what belongs in sys-out and what belongs in sys-err (and what mechanism we want to have for distinguishing the two in the code) rather than splitting this in to two separate issues.

I think it's much more a PM decision (@jhouserizer ) since now with this issue we are fixing the UX output. The code change that you are asking for can be done after the XPC.

mathieucarbou:
> > Although this does fix the problem with the polluted output of the get and export commands I'm not sure it's worth tearing this issue in half. I think I'd rather see us just fix the whole damn thing properly in one shot by decidubg what belongs in sys-out and what belongs in sys-err (and what mechanism we want to have for distinguishing the two in the code) rather than splitting this in to two separate issues.
> 
> I think it's much more a PM decision (@jhouserizer ) since now with this issue we are fixing the UX output. The code change that you are asking for can be done after the XPC, and it won't change the UX output.



> Although this does fix the problem with the polluted output of the get and export commands I'm not sure it's worth tearing this issue in half. I think I'd rather see us just fix the whole damn thing properly in one shot by decidubg what belongs in sys-out and what belongs in sys-err (and what mechanism we want to have for distinguishing the two in the code) rather than splitting this in to two separate issues.

- I will only leave in this PR the proposal code to split the log outputs between STDOUT vs STDERR
- I have moved to #722 the issues that need to be merged (keeping the exact same logging system we have ATM) regarding wrong logging levels, extra lines added, angela not properly redirecting stderr, hidden warnings and errors from not declared loggers, etc.

PR #713, opened 2020-05-08
switch to in-house slaves for linux

akomakom:
**Good news**: Builds take <30 minutes on Linux
**Bad news**:
* at `forkCount=20` tests frequently fail (timeouts in server start/stop).  
* at `forkCount=30`, tests consistently fail
* at `forkCount=10`, tests seem to fail as well (but it's just as fast as with 20)
* at `forkCount=6`, builds take ~32 minutes.
* at `forkCount=2`, builds take 45-50 minutes.

Slaves have 40 cores.

PR #720, opened 2020-05-11
Made forkCount updatable through CLI with -Dtest.parallel.forks=N

mathieucarbou:
forkCount was not updatable since fixed in the POM directly.

With this PR (which is a part from #713 actually) this is possible.

Example locally:

```
> ./mvnw verify -f management/testing/integration-tests/ -Dtest.parallel.forks=6
```

```
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  02:09 min
[INFO] Finished at: 2020-05-11T17:56:31-04:00
[INFO] ------------------------------------------------------------------------
```

And...

```
> ./mvnw verify -f dynamic-config/testing/system-tests/ -Dtest.parallel.forks=1C
```

```
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  15:57 min
[INFO] Finished at: 2020-05-11T18:18:51-04:00
[INFO] ------------------------------------------------------------------------
```

### DIFF
--- a/dynamic-config/testing/system-tests/pom.xml
+++ b/dynamic-config/testing/system-tests/pom.xml
@@ -159,7 +159,7 @@
         <groupId>org.terracotta</groupId>
         <artifactId>maven-forge-plugin</artifactId>
         <configuration>
-          <forkCount>2</forkCount>
+          <forkCount>${test.parallel.forks}</forkCount>
           <systemPropertyVariables>
             <angela.rootDir>${project.build.directory}/angela</angela.rootDir>
             <angela.skipUninstall>true</angela.skipUninstall>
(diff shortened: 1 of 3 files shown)


akomakom:
Duplicates #713 which also enables this in azure-pipelines. 
Note also that `forkCount` is already (theoretically) a user property in the surefire plugin and thus may have unexpected behavior (though in my testing, -DforkCount is ignored)

mathieucarbou:
> forkCount

#713 is to run on jenkins node so I would prefer to decouple both.

This PR is to have the behavior we had some weeks ago, before we have fixed forkCount in the POM. Property was then not working anymore.

I will update this PR to use the same name you have in #713

PR #721, opened 2020-05-12
Fix angela server memory settings for dynamic tests

myronkscott:
The wrong prefixes are set for angela tests.  These settings will reduce memory requirements on server startup.


saurabhagas:
Subsumed in https://github.com/Terracotta-OSS/terracotta-platform/pull/713. Closing.

mathieucarbou:
@saurabhagas : reopening. THis PR has also some benefits on its own and #713 is not totally ready yet.

mathieucarbou:
I have re-opened and rebased yourPR @myronkscott .

PR #728, opened 2020-05-14
Doc corrections

saurabhagas:


docs/src/main/asciidoc/dynamic-config/MigrationFrom10xTo107.adoc
```diff
@@ -83,39 +83,37 @@ Options:
  -u,--upgrade-compatiblity     enable rolling upgrade compatiblity mode
 
 Dynamic Configuration Options:
+    -u, --audit-log-dir
```
saurabhagas, line 97: Pasted the latest output of `start-tc-server` help command. The diff is just reordering of the options (options are sorted based on the long name)

PR #732, opened 2020-05-19
Allow startup using a node name when using a config properties file

saurabhagas:
Allow startup using a node name when using a config properties file

### DIFF
--- a/dynamic-config/server/configuration-provider/src/main/java/org/terracotta/dynamic_config/server/configuration/startup/Options.java
+++ b/dynamic-config/server/configuration-provider/src/main/java/org/terracotta/dynamic_config/server/configuration/startup/Options.java
@@ -136,7 +136,7 @@ public class Options {
   @Parameter(names = {"-N", "--" + CLUSTER_NAME})
   private String clusterName;
 
-  @Parameter(names = {"-l", "--" + LICENSE_FILE})
+  @Parameter(names = {"-l", "--" + LICENSE_FILE}, hidden = true)
   private String licenseFile;
 
   @Parameter(names = {"-D", "--" + REPAIR_MODE})
@@ -164,7 +164,6 @@ public class Options {
    * as the key and user-specified-value as the value.
    *
    * @param jCommander jCommander instance
-   * @return the constructed map
    */
   private void extractTopologyOptions(CustomJCommander jCommander) {
     Collection<String> userSpecifiedOptions = jCommander.getUserSpecifiedOptions();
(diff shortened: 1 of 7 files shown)


dynamic-config/model/src/main/java/org/terracotta/dynamic_config/api/model/Setting.java
```diff
@@ -83,7 +83,7 @@
       NODE,
       fromNode(Node::getNodeName),
       intoNode(Node::setNodeName),
-      of(GET, CONFIG),
+      of(GET, SET, CONFIG),
```
mathieucarbou, line 86: `SET` is sued to describe a setting that can be changed through config-tool set command. If this is the goal, some list below should probably be updated (i.e. requrements, etc), and some extensive unit tests added because right now, the same + stripeId is used as a couple to identify a node, so this will completely break the detection we have in Configuratio class and DyanmicConfigServiceImpl to be able to determine which node we were referring to.
If the name can be changed, we have also to get rid at ALL he places (including NodeContext) where we can refer to a node by using the stripeId / name couple.

Anyway ;-)
I will let you test and discover all the problematic use cases and consequences this has.
The code as it is right now it really not designed to allow a node name change.

dynamic-config/server/configuration-provider/src/main/java/org/terracotta/dynamic_config/server/configuration/startup/Options.java
```diff
@@ -136,7 +136,7 @@
   @Parameter(names = {"-N", "--" + CLUSTER_NAME})
   private String clusterName;
 
-  @Parameter(names = {"-l", "--" + LICENSE_FILE})
+  @Parameter(names = {"-l", "--" + LICENSE_FILE}, hidden = true)
```
mathieucarbou, line 139: should be moved to a separate commit and explain why

dynamic-config/server/configuration-provider/src/main/java/org/terracotta/dynamic_config/server/configuration/startup/Options.java
```diff
@@ -188,53 +187,47 @@ private void extractTopologyOptions(CustomJCommander jCommander) {
...
+      filteredOptions.remove(addDashDash(NODE_NAME));
+      filteredOptions.remove("-n");
       filteredOptions.remove(addDashDash(NODE_CONFIG_DIR));
+      filteredOptions.remove("-r");
 
       if (filteredOptions.size() != 0) {
         throw new ParameterException(
             String.format(
-                "'%s' parameter can only be used with '%s', '%s', '%s', '%s' and '%s' parameters",
+                "'%s' parameter can only be used with '%s', '%s' and '%s' parameters",
```
mathieucarbou, line 207: why removing hostname/port support ? We should be able to use -f with hostname/port or with nodeName, as we want.
The idea behind hostname/port is the usage in cloud, where node names could be unknowm.

saurabhagas: Done

dynamic-config/server/configuration-provider/src/main/java/org/terracotta/dynamic_config/server/configuration/startup/Options.java
saurabhagas: The changes were all in the same file - will be hard to fish out the changes for this. The commit message contains an explanation for this though.

dynamic-config/model/src/main/java/org/terracotta/dynamic_config/api/model/Setting.java
saurabhagas: Reverted, as per discussion over email.

PR #733, opened 2020-05-21
Disallow duplicate node names

saurabhagas:


dynamic-config/model/src/main/java/org/terracotta/dynamic_config/api/service/ClusterValidator.java
```diff
@@ -76,28 +76,29 @@ private void validateFailoverSetting() {
...
+        .map(Map.Entry::getKey)
+        .findAny()
+        .ifPresent(nodeName -> {
+          throw new MalformedClusterException("Found duplicate node name: " + nodeName);
+        });
   }
 
   private void validateServerSettings() {
-    validate(node -> node.getDataDirs().keySet(), "Data directory names of all nodes should match");
+    Collection<Object> settings = cluster.getNodes().stream()
```
saurabhagas, line 96: Just inlined the method.

PR #736, opened 2020-05-29
Cluster validation should happen after all Nomad processors have executed

mathieucarbou:
Cluster validation should happen after all Nomad processors have executed, not only after 1 specific change has been applied. This is to support multiple changes sent at once in the same Nomad transaction.

### DIFF
--- a/dynamic-config/testing/system-tests/src/test/java/org/terracotta/dynamic_config/system_tests/activated/SetCommand1x2IT.java
+++ b/dynamic-config/testing/system-tests/src/test/java/org/terracotta/dynamic_config/system_tests/activated/SetCommand1x2IT.java
@@ -35,4 +35,12 @@ public class SetCommand1x2IT extends DynamicConfigIT {
     assertThat(configToolInvocation("get", "-s", "localhost:" + getNodePort(), "-c", "client-reconnect-window"),
         allOf(hasExitStatus(0), containsOutput("client-reconnect-window=10s")));
   }
+
+  @Test
+  public void testCluster_setDataDirs_postActivation() throws Exception {
+    assertThat(configToolInvocation("set", "-s", "localhost:" + getNodePort(),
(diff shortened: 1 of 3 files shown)


dynamic-config/testing/system-tests/src/test/java/org/terracotta/dynamic_config/system_tests/activated/SetCommand1x2IT.java
```diff
@@ -35,4 +36,15 @@ public void testCluster_setClientReconnectWindow_postActivation() throws Excepti
     assertThat(configToolInvocation("get", "-s", "localhost:" + getNodePort(), "-c", "client-reconnect-window"),
         allOf(hasExitStatus(0), containsOutput("client-reconnect-window=10s")));
   }
+
+  @Test
+  public void testCluster_setDataDirs_postActivation() throws Exception {
```
saurabhagas, line 41: Adding a test where data-dir is being set one just 1 node can be added. This test would be expected to fail.

It would be great if we could print a more helpful message though (hinting at the fact that data-dir names need to match across the cluster, and the same can be achieved usind multiple `-c` options).

mathieucarbou: I will let you add it ;-)

mathieucarbou:
FYI - alls DC system tests passed locally

dynamic-config/testing/system-tests/src/test/java/org/terracotta/dynamic_config/system_tests/activated/SetCommand1x2IT.java
saurabhagas: Filed an issue and assigned to @mobasherul.

PR #737, opened 2020-06-01
Adding Data-Dir tests and improving some logging.

mobasherul:
Adding Data-Dir tests and improving some logging.

### DIFF
--- a/dynamic-config/cli/cli-support/src/main/java/org/terracotta/dynamic_config/cli/command/LocalMainCommand.java
+++ b/dynamic-config/cli/cli-support/src/main/java/org/terracotta/dynamic_config/cli/command/LocalMainCommand.java
@@ -38,8 +38,7 @@ public class LocalMainCommand extends Command {
     Logger rootLogger = (Logger) LoggerFactory.getLogger(ROOT_LOGGER_NAME);
 
     if (verbose) {
-      ConsoleAppender<ILoggingEvent> appender = (ConsoleAppender<ILoggingEvent>) rootLogger.getLoggerContext()
-          .getLogger("org.terracotta.dynamic_config.cli").getAppender("STDOUT");
+      ConsoleAppender<ILoggingEvent> appender = (ConsoleAppender<ILoggingEvent>) rootLogger.getAppender("STDOUT");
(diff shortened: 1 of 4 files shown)


dynamic-config/cli/cli-support/src/main/java/org/terracotta/dynamic_config/cli/command/LocalMainCommand.java
```diff
@@ -38,8 +38,7 @@ public void run() {
     Logger rootLogger = (Logger) LoggerFactory.getLogger(ROOT_LOGGER_NAME);
 
     if (verbose) {
-      ConsoleAppender<ILoggingEvent> appender = (ConsoleAppender<ILoggingEvent>) rootLogger.getLoggerContext()
-          .getLogger("org.terracotta.dynamic_config.cli").getAppender("STDOUT");
+      ConsoleAppender<ILoggingEvent> appender = (ConsoleAppender<ILoggingEvent>) rootLogger.getAppender("STDOUT");
```
mobasherul, line 41: we can then use this from enterprise as well.

myronkscott: I suggest holding this PR until the latest tc-apis is available.  ServerEnv.getServer().console(String msg, Object...) has been added.

mobasherul: ok but this is client side logging

myronkscott: Oops, sorry.  Should this really be dipping into logback?  It would be nice if logging, especially at this level could stick to slf4j.  Maybe define a console logger in the logback config and use that when verbose is requested.

mobasherul: It was earlier done that way but in verbose mode we use the same appender but just change the pattern for that appender. So basically this console appender is already there in the logback file.

PR #739, opened 2020-06-03
Changing tools/config-converter -> tools/upgrade

mobasherul:


dynamic-config/cli/pom.xml
```diff
@@ -34,8 +34,8 @@
   <modules>
     <module>cli-support</module>
     <module>config-tool</module>
-    <module>config-converter</module>
-    <module>config-converter-oss</module>
+    <module>upgrade-tools</module>
+    <module>upgrade-tools-oss</module>
```
mathieucarbou, line 38: why tools and not tool ?

galvan-platform-support/src/main/java/org/terracotta/testing/config/ConfigFileStartupBuilder.java
```diff
@@ -84,7 +84,7 @@ private Path convertToConfigFile() {
     }
 
     List<String> command = new ArrayList<>();
-    String scriptPath = getAbsolutePath(Paths.get("tools", "config-converter", "bin", "config-converter"));
+    String scriptPath = getAbsolutePath(Paths.get("tools", "upgrade", "bin", "config-converter"));
```
mathieucarbou, line 87: upgrade or migrate ? upgrade would closely match an upgrade process from version A to B (like OS upgrade) handling more than config (i.e. the kit could be an upgrade). but in the case of one piece of software is it really an upgrade, especially that some use cases things cannot be "migrated" ?

dynamic-config/cli/pom.xml
mobasherul: we will add one more script for creating permanent entity in enterprise.

mobasherul:
> Some META-INF/services files need to be renamed also

I renamed the META-INF/services for 

> Some META-INF/services files need to be renamed also

My resource/META-iNF/services file has been changed for TcConfigMapper locally but somehow git can't identify that as change or what and is not adding as change.

dynamic-config/cli/pom.xml
jhouserizer: because there are more than one.

jhouserizer: e.g. the tool haque is making to migrate data / entities.

galvan-platform-support/src/main/java/org/terracotta/testing/config/ConfigFileStartupBuilder.java
jhouserizer: there was a fair amount of debate about this, but because the folder will contain multiple things, and the contents will change over time between releases, and 'upgrade' and 'migrate' are ambiguous,  we went with this.

PR #743, opened 2020-06-04
Fix config

myronkscott:
test configuration helper was not compatible with the configuration builder.  

only report a log location if the server has been configured.


dynamic-config/server/configuration-provider/src/main/java/org/terracotta/dynamic_config/server/configuration/startup/StartupConfiguration.java
```diff
@@ -312,7 +312,7 @@ public void setClientReconnectWindow(int value) {
 
       @Override
       public File getLogsLocation() {
-        return substitutor.substitute(pathResolver.resolve(node.getNodeLogDir())).toFile();
+        return (unConfigured) ? null : substitutor.substitute(pathResolver.resolve(node.getNodeLogDir())).toFile();
```
mathieucarbou, line 315: 👍

PR #744, opened 2020-06-10
change location of config-tool, move to new angela version that knows…

jhouserizer:
… about that.

### DIFF
--- a/kit/src/main/kit/tools/bin/config-tool.bat
+++ b/kit/src/main/kit/tools/bin/config-tool.bat
@@ -20,7 +20,7 @@ setlocal EnableExtensions EnableDelayedExpansion
 pushd "%~dp0.."
 set "CONFIG_TOOL_DIR=%CD%"
 popd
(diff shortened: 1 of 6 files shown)


kit/src/main/kit/tools/bin/config-tool.bat
```diff
@@ -20,7 +20,7 @@ setlocal EnableExtensions EnableDelayedExpansion
 pushd "%~dp0.."
 set "CONFIG_TOOL_DIR=%CD%"
 popd
-set "CP=%CONFIG_TOOL_DIR%\lib\*;%CONFIG_TOOL_DIR%\..\common\lib\*"
+set "CP=%CONFIG_TOOL_DIR%\lib\*"
```
chrisdennis, line 23: Should this be called the %TOOLS_DIR% now?

jhouserizer: I thought about that, but figured it still makes sense that it is where the particular tool (config-tool, in this case) is.     Similar for cluster-tool.

PR #747, opened 2020-06-12
Migration Doc update

mobasherul:


docs/src/main/asciidoc/dynamic-config/MigrationFrom10xTo107.adoc
```diff
@@ -209,9 +209,11 @@ Note down the hosts the nodes are running on.
 . Use the <<config-converter-tool, config converter tool>> to convert tc-config.xml files to config directory format.
 . Copy the config directories generated from the step above to the hosts from the first step.
 . Start the nodes using <<startup-script, startup script>> with option `-r`, supplying the config directory path.
-
-+
-IMPORTANT: 10.7 server cannot be started with a 10.5 server data yet. Work for the same is underway in TDB-4975.
-
+. After starting servers run the following command to update the environment to be compatible with the new dynamic configuration.
```
saurabhagas, line 212: The mention of "starting servers" step has been mentioned in the previous step already. Could be omitted from here.

docs/src/main/asciidoc/dynamic-config/MigrationFrom10xTo107.adoc
```diff
@@ -209,9 +209,11 @@ Note down the hosts the nodes are running on.
...
+-s option must contain one hostPort per stripe while running the command to ensure that missing entities have been created on all the stripes in the cluster.
+[source,bash]
+----
+> create-permanent-entities.(bat|sh) -s <hostname[:port]>,...
```
saurabhagas, line 216: Don't quite like the name as it leaks internal details, and "permanent entities" is something users need not know or care about. If the name was agreed upon, it's fine, otherwise we should reconsider it.

docs/src/main/asciidoc/dynamic-config/MigrationFrom10xTo107.adoc
```diff
@@ -209,9 +209,11 @@ Note down the hosts the nodes are running on.
 . Use the <<config-converter-tool, config converter tool>> to convert tc-config.xml files to config directory format.
 . Copy the config directories generated from the step above to the hosts from the first step.
 . Start the nodes using <<startup-script, startup script>> with option `-r`, supplying the config directory path.
-
-+
-IMPORTANT: 10.7 server cannot be started with a 10.5 server data yet. Work for the same is underway in TDB-4975.
-
+. After starting servers run the following command to update the environment to be compatible with the new dynamic configuration.
+-s option must contain one hostPort per stripe while running the command to ensure that missing entities have been created on all the stripes in the cluster.
```
saurabhagas, line 213: The consumers of this text are people who want to migrate their data from an older 10.x to a new 10.7 cluster. All we need to say is "If you want to use the old data with 10.7 server, you need to follow the following steps". Rest everything is internal detail.

docs/src/main/asciidoc/dynamic-config/MigrationFrom10xTo107.adoc
```diff
@@ -209,9 +209,11 @@ Note down the hosts the nodes are running on.
...
+[source,bash]
+----
+> create-permanent-entities.(bat|sh) -s <hostname[:port]>,...
+----
```
saurabhagas, line 217: This block messed up the numbering in this document. (Tip: always look at the HTML output of the adoc. IDEe have asciidoc plugins, so it's right there for you to see). I believe adding a new line would fix the numbering.

docs/src/main/asciidoc/dynamic-config/MigrationFrom10xTo107.adoc
mobasherul: Its agreed upon

saurabhagas:
For @jhouserizer to review now. If it looks good, a PDF of this doc can be updated on the XPC.

docs/src/main/asciidoc/dynamic-config/MigrationFrom10xTo107.adoc
```diff
@@ -77,6 +77,15 @@ When the output format is config properties (i.e. `-t properties`), the generate
 information from all the tc-config XML files, and can also be used to start up the nodes.
 However, an additional cluster activation step may be required after the migrated cluster is started using this config properties file.
 
+[[create-permanent-entity-script]]
+== Permanent Entity Creation Script
+If you want to use old data with 10.7 server, you need to run this script. Here -s option must contain one hostport per stripe to ensure that missing entities are created across all the stripes.
```
jhouserizer, line 82: do we need to add a few words about when the script needs to be ran, and that it needs to be ran on all servers, or at least the active servers, or etc.?

mobasherul: I have mentioned about it -
If you want to use old data with 10.7 server, you need to run this script. Here -s option must contain one hostport per stripe to ensure that missing entities are created across all the stripes.

mobasherul: I have added points about client connection only after running the script.

saurabhagas:
I'm merging this PR since it has been open for quite some time now and the changes look fine to me. Further changes, if any, can be taken up in a future PR.

PR #748, opened 2020-06-12
Add AutoActivatingAngelaRule

ramsai1729:


dynamic-config/testing/support/src/main/java/org/terracotta/dynamic_config/test_support/ClusterDefinition.java
```diff
@@ -32,5 +32,5 @@
 
   boolean autoStart() default true;
 
-  boolean autoActivate() default false;
+  int[] autoActivateNodes() default {-1};
```
ramsai1729, line 35: open for suggestions

dynamic-config/testing/support/src/main/java/org/terracotta/dynamic_config/test_support/AutoActivatingAngelaRule.java
```diff
@@ -0,0 +1,220 @@
...
+      );
+
+
+    } catch (IOException e) {
+      throw new UncheckedIOException(e);
+    }
+  }
+
+  @SuppressFBWarnings("NP_NULL_ON_SOME_PATH_FROM_RETURN_VALUE")
+  private void uploadFolder(RemoteFolder remoteFolder, String parentName, File folder) throws IOException {
```
ramsai1729, line 123: planning to add an API in angela which excludes given files, this can be removed after that change

mathieucarbou:
@myronkscott @chrisdennis @mobasherul : FYI closing this PR. If we want to speed up the activation process we know can use the hidden`-auto-activate` option with a crafted config file and start subsequent nodes with the same config file and `-auto-activate` so that they start, sync, and restart. 
Note: This does not really speed up because making a node "join" a cluster (whether activated or not) will trigger a restart of this node anyway.

PR #759, opened 2020-07-07
[WIP] TDB-5083: Long options

saurabhagas:


mathieucarbou:
@saurabhagas : can you add the output of the help for the commands in the ticket description for @jhouserizer ?

dynamic-config/cli/config-tool/src/main/java/org/terracotta/dynamic_config/cli/config_tool/command/ActivateCommand.java
```diff
@@ -36,30 +36,34 @@
...
-  @Parameter(names = {"-W"}, description = "Maximum time to wait for the nodes to restart. Default: 60s", converter = TimeUnitConverter.class)
+  @Parameter(names = {"-W", "--restart-wait-time"}, description = "Maximum time to wait for the nodes to restart. Default: 60s", converter = TimeUnitConverter.class)
   private Measure<TimeUnit> restartWaitTime = Measure.of(60, TimeUnit.SECONDS);
 
-  @Parameter(names = {"-D"}, description = "Delay before the server restarts itself. Default: 2s", converter = TimeUnitConverter.class)
+  @Parameter(names = {"-D", "--restart-delay"}, description = "Delay before the server restarts itself. Default: 2s", converter = TimeUnitConverter.class)
   private Measure<TimeUnit> restartDelay = Measure.of(2, TimeUnit.SECONDS);
 
-  @Parameter(names = {"-R"}, description = "Restrict the activation process to the node only")
+  @Parameter(names = {"-R", "--restrict-activation"}, description = "Restrict the activation process to the node only")
```
jhouserizer, line 66: It isn't clear to me what this does.  If I have attached a 3x3 cluster, and restrict activation to the node only, what does that mean/do?

dynamic-config/cli/config-tool/src/main/java/org/terracotta/dynamic_config/cli/config_tool/command/ActivateCommand.java
```diff
@@ -36,30 +36,34 @@
...
+import static org.terracotta.dynamic_config.api.model.SettingName.CONFIG_FILE;
+import static org.terracotta.dynamic_config.api.model.SettingName.LICENSE_FILE;
+import static org.terracotta.dynamic_config.api.model.SettingName.NODE_HOSTNAME;
 
 @Parameters(commandNames = "activate", commandDescription = "Activate a cluster")
 @Usage("activate (-s <hostname[:port]> | -f <config-file>) [-n <cluster-name>] [-R] [-l <license-file>] [-W <restart-wait-time>] [-D <restart-delay>]")
 public class ActivateCommand extends RemoteCommand {
 
-  @Parameter(names = {"-s"}, description = "Node to connect to", converter = InetSocketAddressConverter.class)
+  @Parameter(names = {"-s", "--" + NODE_HOSTNAME}, description = "Node to connect to", converter = InetSocketAddressConverter.class)
```
jhouserizer, line 48: I don't like "host-name" for this parameter name.   Let's consider what the parameter means/is specifying. ... Node to connect to.     so why note something like "--connect-to"      (and follow the same suggestion on all other commands)

dynamic-config/cli/config-tool/src/main/java/org/terracotta/dynamic_config/cli/config_tool/command/ConfigurationCommand.java
```diff
@@ -27,13 +27,14 @@
...
+import static org.terracotta.dynamic_config.api.model.SettingName.NODE_HOSTNAME;
 
 public abstract class ConfigurationCommand extends RemoteCommand {
 
-  @Parameter(names = {"-s"}, description = "Node to connect to", required = true, converter = InetSocketAddressConverter.class)
+  @Parameter(names = {"-s", "--" + NODE_HOSTNAME}, description = "Node to connect to", required = true, converter = InetSocketAddressConverter.class)
   InetSocketAddress node;
 
-  @Parameter(names = {"-c"}, description = "Configuration properties", splitter = MultiConfigCommaSplitter.class, required = true, converter = ConfigurationConverter.class)
+  @Parameter(names = {"-c", "--config-property"}, description = "Configuration properties", splitter = MultiConfigCommaSplitter.class, required = true, converter = ConfigurationConverter.class)
```
jhouserizer, line 37: This ties to the bigger discussion about changing the way we identify nodes/stripes/etc. in config commands.  Some of those variants we were debating would have different option names here.

dynamic-config/cli/config-tool/src/main/java/org/terracotta/dynamic_config/cli/config_tool/command/ExportCommand.java
```diff
@@ -40,23 +40,25 @@
...
-  @Parameter(names = {"-f"}, description = "Output configuration file", converter = PathConverter.class)
+  @Parameter(names = {"-f", "--" + CONFIG_FILE}, description = "Output configuration file", converter = PathConverter.class)
   private Path outputFile;
 
-  @Parameter(names = {"-i"}, description = "Include default values. Default: false", converter = BooleanConverter.class)
+  @Parameter(names = {"-i", "--include-defaults"}, description = "Include default values. Default: false", converter = BooleanConverter.class)
   private boolean includeDefaultValues;
 
-  @Parameter(names = {"-r"}, description = "Export the runtime configuration instead of the configuration saved on disk. Default: false", converter = BooleanConverter.class)
+  @Parameter(names = {"-r", "--runtime-config"}, description = "Read the runtime configuration instead of the configuration saved on disk. Default: false", converter = BooleanConverter.class)
```
jhouserizer, line 58: begs the question "what is runtime config" vs. config saved on disk.... why would they be different.

dynamic-config/cli/config-tool/src/main/java/org/terracotta/dynamic_config/cli/config_tool/command/TopologyCommand.java
```diff
@@ -34,22 +34,23 @@
...
 import static java.lang.System.lineSeparator;
+import static org.terracotta.dynamic_config.api.model.SettingName.NODE_HOSTNAME;
 import static org.terracotta.dynamic_config.cli.config_tool.converter.OperationType.STRIPE;
 
 /**
  * @author Mathieu Carbou
  */
 public abstract class TopologyCommand extends RemoteCommand {
-  @Parameter(names = {"-t"}, description = "Determine if the sources are nodes or stripes. Default: node", converter = OperationType.TypeConverter.class)
+  @Parameter(names = {"-t", "--type"}, description = "Determine if the sources are nodes or stripes. Default: node", converter = OperationType.TypeConverter.class)
```
jhouserizer, line 44: rather than 

attach  --type stripe  --destination-hostname foo1.local  --source-hostname foo2.local

Isn't it more clear to have something more like:

attach --stripe-to-add-to  xxx --node-to-add yyy
attach --cluster-to-add-to xxx --stripe-to-add yyy


We have gotten a lot of feedback that "source" and "destination" are way too ambiguous, depending upon the frame of mind of who is running the command.

dynamic-config/cli/config-tool/src/main/java/org/terracotta/dynamic_config/cli/config_tool/command/ActivateCommand.java
mathieucarbou: @jhouserizer : see the powerpoint presentation I did.
Activation by default activates the whole cluster formed by all the unconfigured nodes.

Restrictive activation is a way to activate only 1 node with a same topology as an already existing cluster to make it join this cluster. This can be used to append nodes to a 1-stripe cluster that was started with an initial topology with more than 1 node, or it can be used to "repair" an incomplete attach command to make the node join the cluster.
Basically, when a topology of an activated cluster has 1 node that is supposed to be there, but not yet there, restrictive activation triggers the activation process to ONLY the targeted node, which will restart, sync, and join ONLY if its topology matches the one of the cluster it wants to join.

This should be explained in the documentation... If not, @tmesic99 @saurabhagas : can you please make sure this is explained ?

dynamic-config/cli/config-tool/src/main/java/org/terracotta/dynamic_config/cli/config_tool/command/ActivateCommand.java
mathieucarbou: IMO, CLI parameters should describe what they are and not what they are used for, which has to be explained in the help. There seem to be a sort of convention around that when looking at all the CLI out there.
If we have a `--connect-to` parameter, the user knows what this parameter will be used for but it does not hint about what to pass: uri with a specific scheme ? ip ? port ? ip:port ? hostname ? list of addresses ? other thing ? I know that --hostname would not entirely answer that either but this is closer I think because in the doc, this is easier to say: `all --hostname parameters should be in the format <hostname[:port]>`

dynamic-config/cli/config-tool/src/main/java/org/terracotta/dynamic_config/cli/config_tool/command/TopologyCommand.java
mathieucarbou: @jhouserizer : agree, I have described something similar in the ticket TDB-5083: you could have a look ?

dynamic-config/cli/cli-support/src/main/java/org/terracotta/dynamic_config/cli/command/CustomJCommander.java
```diff
@@ -128,28 +135,30 @@ private void appendOptions(JCommander jCommander, StringBuilder out, String inde
...
-      if (nameSpace.length() > maxNamespaceLength) {
-        maxNamespaceLength = nameSpace.length();
+    if (toolName == ToolName.CONFIG_TOOL) {
+      out.append(indent).append(lineSeparator()).append("Definitions:").append(lineSeparator());
+      out.append(indent).append("    ").append("namespace").append(lineSeparator());
+      Map<String, String> nameSpaces = new LinkedHashMap<>();
+      nameSpaces.put("stripe.<stripeId>.node.<nodeId>", "to apply a change only on a specific node");
+      nameSpaces.put("stripe.<stripeId>", "to apply a change only on a specific stripe");
+      nameSpaces.put("'' (empty namespace)", "to apply a change only on all nodes of the cluster");
+
```
mathieucarbou, line 145: I disagree with putting some specific behaviour in a common class, especially if it requires putting some enums.

Another way to do that would be, in this class to have this empty method stub:

`protected void appendDefinitions(StringBuilder out, String indent) {}`

And in config tool, we are using it like that:

```
CustomJCommander jCommander = new CustomJCommander("config-tool", commandRepository, mainCommand) {
    @Override
    protected void appendDefinitions(StringBuilder out, String indent) {
        // place the code here
    }
};
jCommander.
```

There are also several other ways to do that. But please remove all the specific code from this common class ;-)This defies the purpose of having such common class there, and makes it harder to change and leads to more code smell like putting enums there.

dynamic-config/cli/config-tool/src/main/java/org/terracotta/dynamic_config/cli/config_tool/ConfigTool.java
```diff
@@ -147,7 +149,7 @@ public static void start(String... args) {
   }
 
   private static CustomJCommander parseArguments(CommandRepository commandRepository, RemoteMainCommand mainCommand, String[] args) {
-    CustomJCommander jCommander = new CustomJCommander("config-tool", commandRepository, mainCommand);
+    CustomJCommander jCommander = new CustomJCommander(CONFIG_TOOL, commandRepository, mainCommand);
```
mathieucarbou, line 152: to remove.
specific command behaviours should be handled in each command, not at a common place that knows all the commands.

dynamic-config/cli/config-tool/src/main/java/org/terracotta/dynamic_config/cli/config_tool/command/ActivateCommand.java
```diff
@@ -36,30 +36,34 @@
 
 import static java.lang.System.lineSeparator;
 import static java.util.Collections.singletonList;
+import static org.terracotta.dynamic_config.api.model.SettingName.CLUSTER_NAME;
+import static org.terracotta.dynamic_config.api.model.SettingName.CONFIG_FILE;
+import static org.terracotta.dynamic_config.api.model.SettingName.LICENSE_FILE;
+import static org.terracotta.dynamic_config.api.model.SettingName.NODE_HOSTNAME;
 
 @Parameters(commandNames = "activate", commandDescription = "Activate a cluster")
 @Usage("activate (-s <hostname[:port]> | -f <config-file>) [-n <cluster-name>] [-R] [-l <license-file>] [-W <restart-wait-time>] [-D <restart-delay>]")
```
mathieucarbou, line 45: usage needs to be updated with long parameters

dynamic-config/cli/cli-support/src/main/java/org/terracotta/dynamic_config/cli/command/CustomJCommander.java
saurabhagas: Sounds good. Will do.

dynamic-config/cli/config-tool/src/main/java/org/terracotta/dynamic_config/cli/config_tool/command/ActivateCommand.java
jhouserizer: Woah, I'm confused now.  That sounds a lot like "attach"ing a new node to a stripe.

dynamic-config/cli/config-tool/src/main/java/org/terracotta/dynamic_config/cli/config_tool/ConfigTool.java
saurabhagas: Indeed

dynamic-config/cli/config-tool/src/main/java/org/terracotta/dynamic_config/cli/config_tool/command/ActivateCommand.java
saurabhagas: Indeed

dynamic-config/cli/config-tool/src/main/java/org/terracotta/dynamic_config/cli/config_tool/command/ActivateCommand.java
jhouserizer: Ah, but "myserver:9410"   is not a hostname!

dynamic-config/cli/config-tool/src/main/java/org/terracotta/dynamic_config/cli/config_tool/command/ActivateCommand.java
mathieucarbou: not really: flow and validation is completely different (slides also show that). this is really a way to "push" a topology to a node and restart it, activated so that it syncs. This can be used to repair. This is not an attachement: an attachment is when the destination cluster does not yet contains the node definition.
restrictive activation can only be used if the destination cluster already contains the node definition, but the node is missing and we want to activate a node to add it.

mathieucarbou:
@saurabhagas : talked to @jhouserizer and @GaryWKeim . I have updated the [comment above](https://github.com/Terracotta-OSS/terracotta-platform/pull/759#pullrequestreview-443972497) for the things to do.

mathieucarbou:
@saurabhagas : following lead sync, these things have changed:
- `export -connect-to <hostname[:port]> [-output-file <config-file>] [-include-defaults] [-runtime] [-mapping]`

saurabhagas:
> @saurabhagas : following lead sync, these things have changed:
> 
> * `export -connect-to <hostname[:port]> [-output-file <config-file>] [-include-defaults] [-runtime] [-mapping]`

Thanks. What is the `-mapping` option you refer to? Nothing corresponds to it currently..

mobasherul:
Closing this in favor of https://github.com/Terracotta-OSS/terracotta-platform/pull/879

PR #760, opened 2020-07-07
update deps

myronkscott:


dynamic-config/server/configuration-provider/src/main/java/org/terracotta/dynamic_config/server/configuration/startup/StartupConfiguration.java
```diff
@@ -305,7 +305,6 @@ public int getClientReconnectWindow() {
         return nodeContext.getCluster().getClientReconnectWindow().getExactQuantity(TimeUnit.SECONDS).intValueExact();
       }
 
-      @Override
```
mathieucarbou, line 308: does it mean the method can be removed if not anymore in the api ?

dynamic-config/server/services/src/main/java/org/terracotta/dynamic_config/server/service/handler/ClientReconnectWindowConfigChangeHandler.java
```diff
@@ -53,30 +36,6 @@ public void validate(NodeContext nodeContext, Configuration change) throws Inval
...
-          .anyMatch(attr -> ATTR_NAME.equals(attr.getName()) && attr.isReadable() && attr.isWritable());
-    } catch (JMException e) {
-      LOGGER.error("MBeanServer::getMBeanInfo resulted in:", e);
-      canCall = false;
-    }
-
-    if (!canCall) {
-      throw new InvalidConfigChangeException("Unsupported change: " + change);
-    }
+//  change handler is not needed.  core will read directly from the configuration every time it is needed
```
myronkscott, line 39: @mathieucarbou is this change correct?  The Node tree elsewhere?

dynamic-config/server/configuration-provider/src/main/java/org/terracotta/dynamic_config/server/configuration/startup/StartupConfiguration.java
myronkscott: this is a part of the config system.  any change should be set in the config system and read by the platform through ServerConfiguration object.

PR #761, opened 2020-07-07
TDB-5067: Added a unit test regarding changing a setting and changing it back again

mathieucarbou:
if changing a setting requires a restart, if we change it back, no restart is required anymore.

A restart is required ONLY when the topology written on disk after a commit differs from the runtime topology because some changes require a restart to be applied.


saurabhagas:
Changes merged in PR#763

PR #766, opened 2020-07-09
TDB-5092: Improve error message when rejecting a setting that cannot be changed once activated

mathieucarbou:


dynamic-config/server/configuration-repository/src/main/java/org/terracotta/dynamic_config/server/configuration/nomad/processor/SettingNomadChangeProcessor.java
```diff
@@ -62,7 +62,7 @@ public void validate(NodeContext baseConfig, SettingNomadChange change) throws N
       ConfigChangeHandler configChangeHandler = getConfigChangeHandlerManager(change);
       configChangeHandler.validate(baseConfig, configuration);
     } catch (InvalidConfigChangeException | RuntimeException e) {
-      throw new NomadException("Error when trying to apply setting change '" + change.getSummary() + "': " + e.getMessage(), e);
+      throw new NomadException("'" + change.getSummary() + "': " + e.getMessage(), e);
```
tmesic99, line 65: Is there actually better context on the line that was removed?  The new error only shows "changeSummary:Message".  But the old error puts that in context with it having occurred "when trying to apply a setting change".

tmesic99: Oh, I guess the 'Summary' probably contains that context already?

dynamic-config/testing/system-tests/src/test/java/org/terracotta/dynamic_config/system_tests/activated/SetCommand1x1IT.java
```diff
@@ -265,12 +265,12 @@ public void metadata_dir_cannot_be_changed() throws Exception {
...
     assertThat(
         configToolInvocation("set", "-s", "localhost:" + getNodePort(), "-c", "stripe.1.metadata-dir=foo"),
-        containsOutput("Unable to apply this change: stripe.1.metadata-dir=foo"));
+        containsOutput("Reason: 'set metadata-dir=foo (stripe ID: 1)': Setting 'metadata-dir' cannot be changed once a node is activated"));
     assertThat(
         configToolInvocation("set", "-s", "localhost:" + getNodePort(), "-c", "stripe.1.node.1.metadata-dir=foo"),
-        containsOutput("Unable to apply this change: stripe.1.node.1.metadata-dir=foo"));
+        containsOutput("Reason: 'set metadata-dir=foo (stripe ID: 1, node: node-1-1)': Setting 'metadata-dir' cannot be changed once a node is activated"));
   }
 }
```
tmesic99, line 276: Shouldn't all these string literals be set as static variables and referenced across the various src/main and src/test classes?  For example, if we ever want to modify our error messages for better clarity, guidance, grammar fixes, etc.., we will have to visit many test cases to ensure the tests continue to function correctly?

dynamic-config/server/configuration-repository/src/main/java/org/terracotta/dynamic_config/server/configuration/nomad/processor/SettingNomadChangeProcessor.java
mathieucarbou: yes, there was a duplication ;-) this message is in a caller class appended to another one.

dynamic-config/testing/system-tests/src/test/java/org/terracotta/dynamic_config/system_tests/activated/SetCommand1x1IT.java
mathieucarbou: Yes we could but this is a big refactoring and this module would need to be unique across many projects. Also, some errors are changing a little depending on the context and which exception is wrapping what.
Overall, until we have a proper class that defines error codes and error messages I do not think factoring it would help because here for example in the system tests module we do not have and must not have access to the "internal" constant. So there will be at least a duplication...

You could still raise a TDB (not an issue but a user story) and put it in the backlog about having across all CLI better error codes and error messages - which requires a review of every potential exceptions thrown. This is big and would be low priority for now.

PR #773, opened 2020-07-12
Public address validations

saurabhagas:
Added checks to disallow duplicate public addresses

### DIFF
--- a/dynamic-config/model/src/main/java/org/terracotta/dynamic_config/api/service/ClusterValidator.java
+++ b/dynamic-config/model/src/main/java/org/terracotta/dynamic_config/api/service/ClusterValidator.java
@@ -19,7 +19,6 @@ import org.terracotta.dynamic_config.api.model.Cluster;
 import org.terracotta.dynamic_config.api.model.Node;
 import org.terracotta.dynamic_config.api.model.Setting;
 
-import java.net.InetSocketAddress;
 import java.util.Collection;
 import java.util.Map;
 import java.util.Objects;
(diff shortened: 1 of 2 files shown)


dynamic-config/api/src/main/java/org/terracotta/dynamic_config/api/model/nomad/SettingNomadChange.java
```diff
@@ -194,10 +194,14 @@ public static SettingNomadChange fromConfiguration(Configuration configuration,
...
   @SuppressWarnings("OptionalGetWithoutIsPresent")
-  private static Applicability toApplicability(Configuration configuration, Cluster cluster) {
+  private static Applicability toApplicability(Configuration configuration, Cluster originalCluster, Cluster updatedCluster) {
     switch (configuration.getScope()) {
       case NODE:
-        return Applicability.node(configuration.getStripeId(), cluster.getNode(configuration.getStripeId(), configuration.getNodeId()).get().getNodeName());
+        if (configuration.getSetting() == Setting.NODE_NAME) {
+          return Applicability.node(configuration.getStripeId(), originalCluster.getNode(configuration.getStripeId(), configuration.getNodeId()).get().getNodeName());
+        } else {
+          return Applicability.node(configuration.getStripeId(), updatedCluster.getNode(configuration.getStripeId(), configuration.getNodeId()).get().getNodeName());
```
mathieucarbou, line 203: you should NEVER pass the client's updated cluster to server. There is a way on server side to apply the change.
Also, this is not the right way to do that: with TDB-5072, we will use UUIDs now so all this code will change so that node name is only on user facing side and gets converted to a uuid ASAP.

dynamic-config/model/src/main/java/org/terracotta/dynamic_config/api/service/ClusterValidator.java
```diff
@@ -55,18 +54,50 @@ public void validate() throws MalformedClusterException {
...
+          throw new MalformedClusterException("Nodes with names: " + entry.getValue().stream().map(Node::getNodeName).collect(Collectors.joining(", ")) +
+              " have the same address: '" + entry.getKey() + "'");
+        });
+  }
+
+  private void checkDuplicatePublicAddresses() {
+    cluster.getStripes()
+        .stream()
+        .flatMap(s -> s.getNodes().stream())
+        .collect(groupingBy(Node::getNodePublicAddress, Collectors.toList()))
```
mathieucarbou, line 93: you need to filter out the nodes node having a nodePublicAddress first. The field is optional also to calling that directly won't work.

dynamic-config/model/src/main/java/org/terracotta/dynamic_config/api/service/ClusterValidator.java
```diff
@@ -55,18 +54,50 @@ public void validate() throws MalformedClusterException {
   }
 
   private void validateAddresses() {
+    checkDuplicateInternalAddresses();
+    validatePublicAddresses();
+    checkDuplicatePublicAddresses();
```
mathieucarbou, line 59: we should add a validation to check that if we have 2 nodes or more, we do not have node 1 for example having a nodePublicAddress that matches the internal address of another node.

In another way, from a UX perspective, we have to prevent a user to set a public address on a node that would be the internal address of another node, because the node would impersonate it.

BUT we have to allow the user to the for the same node the same public address as the internal address. Public and internal can be the same, and this is valid in case all his servers are accessible from outside the FW with a dns name he could use that both for internal and public.

dynamic-config/model/src/main/java/org/terracotta/dynamic_config/api/service/ClusterValidator.java
saurabhagas: This is taken care of in `.filter(e -> e.getKey().isPresent() && e.getValue().size() > 1)` call later.

dynamic-config/model/src/main/java/org/terracotta/dynamic_config/api/service/ClusterValidator.java
saurabhagas: Hmm. Important points. I want to understand it better before hurrying to make a change. Will do the necessary changes after discussing it briefly with you.

saurabhagas:
> @jhouserizer @saurabhagas : I strongly suggest that we add a validation in the set command to validate that the public address sis reachable.
> Because setting a public address that is not reachable to a node at runtime can break the cluster and prevent clients to connect.

Let's talk about it with James today and reach a consensus. For now, I'm just removing the server name change commit.

dynamic-config/model/src/main/java/org/terracotta/dynamic_config/api/service/ClusterValidator.java
mathieucarbou: you need to do the filtering before grouping because grouping in a map requires to use the equals/hashcode of the object - here you are grouping based on the optional object

PR #777, opened 2020-07-15
TDB-5095: Use to the Nomad Entity channel for the Nomad COMMIT phase of a configuration change to avoid a race condition during passive sync

mathieucarbou:


dynamic-config/cli/config-tool/src/main/java/org/terracotta/dynamic_config/cli/config_tool/nomad/NomadManager.java
```diff
@@ -141,8 +143,11 @@ public void runTopologyChange(Cluster destinationCluster, Map<InetSocketAddress,
     return new NomadClient<>(nomadEndpoints, host, user, clock);
   }
 
-  private NomadClient<T> createTopologyChangeNomadClient(Cluster destinationCluster, Map<InetSocketAddress, LogicalServerState> onlineNodes) {
-    LOGGER.trace("createPassiveChangeNomadClient({}, {})", destinationCluster, onlineNodes);
+  /**
+   * create a nomad client that is preparing through diagnostic port and committing through entity channel
+   */
+  private NomadClient<T> createBiChannelNomadClient(Cluster destinationCluster, Map<InetSocketAddress, LogicalServerState> onlineNodes) {
```
mobasherul, line 149: Few questions : What about rollback ? Do we want to do that also over entity channel.
Why the server side NomadEntityImpl have checks for Prepare and Rollback message if that is send over diagnostic channel .Is there any scenario where those messages also gets send over entity channel.

mathieucarbou: I don't see any use case since a passive could be synced again. WDYT @ramsai1729 ?
I think we could do it yes I don't see any reason why we should rollback with diagnostic.

mathieucarbou: @mobasherul : oh, forget what I said. I am mistaken. For the rollback phase that is handled by nomad here, the roollbacl does not need to go through entities because it will rollback a partially prepared change. So diagnostic port is involved all teh way. We cannot have some servers that have committed, come prepare and then rollback, or rollback a committed tx.
So COMMIT is really an END phase: if we go there, we are sure we need to commit.
If we Rollback, we just need to rollbac a prepared change that won't be activated on servers.

mathieucarbou:
@saurabhagas @mobasherul : PR ready to be merged.

PR #781, opened 2020-07-16
Adding tests for network disruption

mobasherul:


dynamic-config/testing/support/src/main/java/org/terracotta/dynamic_config/test_support/DynamicConfigIT.java
```diff
@@ -114,15 +121,19 @@ public DynamicConfigIT() {
...
+    this(testTimeout, Paths.get(System.getProperty("user.dir"), "target", "test-data"), isNetDisruptionEnabled);
+  }
+  
+  public DynamicConfigIT(Duration testTimeout, Path parentTmpDir, boolean isNetDisruptionEnabled) {
     ClusterDefinition clusterDef = getClass().getAnnotation(ClusterDefinition.class);
     this.timeout = testTimeout.toMillis();
     this.rules = RuleChain.emptyRuleChain()
         .around(tmpDir = new TmpDir(parentTmpDir, false))
-        .around(angela = new AngelaRule(createConfigurationContext(clusterDef.stripes(), clusterDef.nodesPerStripe()), clusterDef.autoStart(), clusterDef.autoActivate()) {
+        .around(angela = new AngelaRule(createConfigurationContext(clusterDef.stripes(), clusterDef.nodesPerStripe(), isNetDisruptionEnabled), clusterDef.autoStart(), clusterDef.autoActivate()) {
```
mathieucarbou, line 136: I would prefer to see this parameter in the `ClusterDefinition` annotation which is controlling the test parameters.
Timeout, sadly, is a special case but ideally it would have been put in this annotation too.

dynamic-config/testing/support/src/main/java/org/terracotta/dynamic_config/test_support/DynamicConfigIT.java
```diff
@@ -213,7 +224,7 @@ protected final int getNodePort() {
   protected final InetSocketAddress getNodeAddress(int stripeId, int nodeId) {
     return InetSocketAddress.createUnresolved("localhost", getNodePort(stripeId, nodeId));
   }
-
+
```
mathieucarbou, line 227: space added here

dynamic-config/testing/support/src/main/java/org/terracotta/dynamic_config/test_support/DynamicConfigIT.java
```diff
@@ -511,4 +523,54 @@ protected Duration getConnectionTimeout() {
...
+  private boolean isServerBlocked(TerracottaServer server) {
+    try (Connection connection = tryDiagnosticConnection(server)) {
+      try (Diagnostics diagnostics = fetchDiagnosticEntity(connection)) {
+        return Boolean.parseBoolean(diagnostics.invoke("ConsistencyManager", "isBlocked"));
+      }
+    } catch (Exception e) {
+      e.printStackTrace();
+      return false;
+    }
+  }
```
mathieucarbou, line 556: you could use the new diagnostic layer that is already wired in this class (example at the end):
```
    try (DiagnosticService diagnosticService = DiagnosticServiceFactory.fetch(
        InetSocketAddress.createUnresolved(host, port),
        getClass().getSimpleName(),
        getConnectionTimeout(),
        getConnectionTimeout(),
        null,
        objectMapperFactory)) {
      return diagnosticService.isBlocked();
    }
```

dynamic-config/testing/support/src/main/java/org/terracotta/dynamic_config/test_support/DynamicConfigIT.java
```diff
@@ -511,4 +523,54 @@ protected Duration getConnectionTimeout() {
...
+
+  protected void waitForServerBlocked(TerracottaServer server) {
+    waitUntil(() -> isServerBlocked(server), is(true), ASSERT_TIMEOUT);
```
mathieucarbou, line 559: I think `waitUntil(() -> isServerBlocked(server), is(true));` is enough: there is a method taking a supplier and matcher and re-using the assertion timeout already.

It is also best to use `getAssertTimeout()` instead of `ASSERT_TIMEOUT ` because `getAssertTimeout()` is meant to be overridden in sub classes to set a different timeout

dynamic-config/testing/support/pom.xml
```diff
@@ -45,6 +45,11 @@
       <artifactId>common-test-utilities</artifactId>
       <version>${project.version}</version>
     </dependency>
+    <dependency>
+      <groupId>org.terracotta.internal</groupId>
+      <artifactId>client-runtime</artifactId>
+      <version>${terracotta-core.version}</version>
+    </dependency>
```
mathieucarbou, line 52: you should't need this addition: this is the module that will use testing support that will provide it. Because for EE, this is another dependency.

dynamic-config/testing/support/src/main/java/org/terracotta/dynamic_config/test_support/DynamicConfigIT.java
```diff
@@ -511,4 +523,54 @@ protected Duration getConnectionTimeout() {
...
+
+  private Connection tryDiagnosticConnection(TerracottaServer server) throws Exception {
+    try {
+      ConnectionService connectionService = new DiagnosticConnectionService();
+      URI uri = URI.create("diagnostic://" + server.getHostname() + ":" + server.getTsaPort());
+      return connectionService.connect(uri, new Properties());
+    } catch (Exception e) {
+     throw e;
+    }
+  }
```
mathieucarbou, line 575: these 2 methods above can be removed - please use the new diagnostic layer ;-)

mobasherul:
> @mobasherul @ramsai1729 : let's ensure the windows build all pass before merging to be sure that the disruption tests won't fail all the subsequent windows builds. Thanks!

All builds have passed now . Please merge it.

PR #789, opened 2020-07-21
TDB-5117: Print more specific message if changes require specific node restart

saurabhagas:
Rebased on https://github.com/Terracotta-OSS/terracotta-platform/pull/802


saurabhagas:
> @saurabhagas : you need to add a commit to handle the backend Nomad part also and test by doing some diagnostic call when changing the node log dir to ensure that the diagnostic call is only saying that only 1 node needs to restart, not the whole cluster.
> 
> For that, you need to change this method:
> 
> `DynamicConfigNomadChange#canApplyAtRuntime()`
> 
> to
> 
> `DynamicConfigNomadChange#canApplyAtRuntime(InetSocketAddress currentNode)`
> 
> At every places this method is called, you need to pass the current node address by using something like `nodeContext.getNode().getAddress()`
> 
> And in `SettingNomadChange`, this code needs an update:
> 
> ```
>   public boolean canApplyAtRuntime() {
>     return !getSetting().isRestartRequired();
>   }
> ```
> 
> The new logic would be to return `false` if `Setting.requires(CLUSTER_RESTART)`, or also return false if `Setting.requires(NODE_RESTART)` AND that the the current node passed as a parameter is impacted by this change.
> 
> Simply, stated: we can apply at runtime if the change does not require a cluster restart **and** does not require this node to restart
> 
> And you can then remove the now obsolete method: `Setting#isRestartRequired()`, which was for the old logic.
> 
> Also please do not forget to add a test regarding this diagnostic call to verify the output for a cluster restart VS a node restart.

I'm confused. Let's talk about this.

saurabhagas:
@mathieucarbou changes done.

dynamic-config/api/src/main/java/org/terracotta/dynamic_config/api/model/nomad/Applicability.java
```diff
@@ -26,9 +26,9 @@
 import static org.terracotta.dynamic_config.api.model.Scope.STRIPE;
 
 public class Applicability {
-  private Scope scope;
-  private String nodeName;
-  private Integer stripeId;
+  private final Scope scope;
+  private final String nodeName;
+  private final Integer stripeId;
```
mathieucarbou, line 31: 👍

saurabhagas:
> @saurabhagas : needs a rebase first!

You can "Rebase and merge" :-)

mathieucarbou:
> > @saurabhagas : needs a rebase first!
> 
> You can "Rebase and merge" :-)

no: rebase and merge will remove the merge commit... The right process is to rebase, then merge.

mathieucarbou:
> > > @saurabhagas : needs a rebase first!
> > 
> > 
> > You can "Rebase and merge" :-)
> 
> no: rebase and merge will remove the merge commit... The right process is to rebase, then merge.

Also, rebase & merge is not right because the build would not have ran against the same commits.

PR #793, opened 2020-07-23
[URG] Node addition and removal should call the mBean with the node internal address, not the public address

mathieucarbou:
Note: this is not a fix for TDB-5050.

Whatever solution is chosen to fix TDB-5050, we have to use the node internal addresses, not the public ones.


mathieucarbou:
Rebased on #796

PR #795, opened 2020-07-23
TDB-5122 : Dynamic Scaling UX : Sync license and nomad changes from existing stripe to nodes in newly added stripe

mathieucarbou:


mathieucarbou:
@rahul-mittal @akomakom : I had t ore-open this PR because the checks results are not showing teh correct build output, and they are building on something wrong.
In the check view, we see that a test is failing but this test as been removed from master and is not in the PR code.

PR #804, opened 2020-07-28
Ubiquitous even-sum warning

saurabhagas:


dynamic-config/server/configuration-provider/src/main/java/org/terracotta/dynamic_config/server/configuration/service/DynamicConfigServiceImpl.java
```diff
@@ -95,7 +95,6 @@ public DynamicConfigServiceImpl(NodeContext nodeContext, LicenseService licenseS
     if (hasLicenseFile()) {
       validateAgainstLicense(upcomingNodeContext.getCluster());
     }
-    new ClusterValidator(nodeContext.getCluster()).validate();
```
mathieucarbou, line 98: not sure I would like to remove that.... This is a safeguard that will prevent the server to start if the computed topology is not right at one point. If there could be a bug one day somewhere (cli parsing, config parsing, migration, etc) that could let an invalid topology go in the config system or parsed at startup.

mathieucarbou: Note: but I agree this would warrant a comment before the line so that it does not get removed again...

saurabhagas: Done

saurabhagas: @mathieucarbou I don't see the point in waiting for another build. Please merge if it looks good to you.

PR #807, opened 2020-07-29
Fix configuration logging when a new change is applied

saurabhagas:
- Log new configuration only after the change has been committed
- Beautify the properties so that they're easy to read
- Log "Pending configuration" instead of "New cluster configuration" if a runtime change is applied after a change requiring a cluster restart.

### DIFF
--- a/dynamic-config/server/configuration-provider/src/main/java/org/terracotta/dynamic_config/server/configuration/service/DynamicConfigServiceImpl.java
+++ b/dynamic-config/server/configuration-provider/src/main/java/org/terracotta/dynamic_config/server/configuration/service/DynamicConfigServiceImpl.java
@@ -31,6 +31,7 @@ import org.terracotta.dynamic_config.api.model.nomad.MultiSettingNomadChange;
 import org.terracotta.dynamic_config.api.model.nomad.SettingNomadChange;
 import org.terracotta.dynamic_config.api.service.ClusterValidator;
 import org.terracotta.dynamic_config.api.service.DynamicConfigService;
+import org.terracotta.dynamic_config.api.service.Props;
 import org.terracotta.dynamic_config.api.service.TopologyService;
 import org.terracotta.dynamic_config.server.api.DynamicConfigEventService;
 import org.terracotta.dynamic_config.server.api.DynamicConfigListener;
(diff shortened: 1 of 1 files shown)


dynamic-config/server/configuration-provider/src/main/java/org/terracotta/dynamic_config/server/configuration/service/DynamicConfigServiceImpl.java
```diff
@@ -275,9 +275,18 @@ public void onNomadCommit(CommitMessage message, AcceptRejectResponse response,
...
             Cluster runtimeCluster = nomadChange.apply(runtimeNodeContext.getCluster());
             runtimeNodeContext = runtimeNodeContext.withCluster(runtimeCluster).orElseGet(runtimeNodeContext::alone);
+            LOGGER.info("New cluster configuration: {}{}", lineSeparator(), clusterProperties(runtimeNodeContext));
+          } else {
+            LOGGER.info("Pending cluster configuration: {}{}", lineSeparator(), clusterProperties(upcomingNodeContext));
           }
         }
       }
+
+      if (runtimeNodeContext.equals(upcomingNodeContext)) {
```
mobasherul, line 285: It already logged right line 278 Why logging again ? same for line 280

dynamic-config/server/configuration-provider/src/main/java/org/terracotta/dynamic_config/server/configuration/service/DynamicConfigServiceImpl.java
```diff
@@ -430,6 +439,10 @@ public synchronized boolean validateAgainstLicense(Cluster cluster) throws Inval
     return true;
   }
 
+  private String clusterProperties(NodeContext upcomingNodeContext) {
```
mobasherul, line 442: rename to nodecontext as runtimecontext can also be passed as parameter name is bit confusing

dynamic-config/server/configuration-provider/src/main/java/org/terracotta/dynamic_config/server/configuration/service/DynamicConfigServiceImpl.java
saurabhagas: Pushed the wrong commit. Fixed.

PR #809, opened 2020-07-29
TDB-5124: API for running config-tool script from server side

devlg75:


dynamic-config/server/configuration-provider/src/main/java/org/terracotta/dynamic_config/server/configuration/util/InvokeConfigToolScript.java
```diff
@@ -0,0 +1,67 @@
...
+                    .command(cmd)
+                    .redirectErrorStream(true)
+                    .start();
+            BufferedReader reader = new BufferedReader(new InputStreamReader(p.getInputStream()));
+            String line;
+            while ((line = reader.readLine()) != null) {
+                LOGGER.info(line);
+            }
+            p.waitFor();
+            throw new ScriptRunFailureException("config-tool script run failure at server side for command: " + cmd.toString());
```
ramsai1729, line 56: this may not be applicable to all config-tool commands

devlg75: yeah this exception was specific to detach, i modified that now.

dynamic-config/server/configuration-provider/src/main/java/org/terracotta/dynamic_config/server/configuration/util/InvokeConfigToolScript.java
```diff
@@ -0,0 +1,75 @@
...
+import java.nio.charset.StandardCharsets;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.stream.Collectors;
+
+public class InvokeConfigToolScript {
+    private static final Logger LOGGER = LoggerFactory.getLogger(InvokeConfigToolScript.class);
+
+    public void execCommand(String[] command) throws InterruptedException, IOException, ScriptRunFailureException {
+        String path = System.getProperty("tc.install-root");
```
chrisdennis, line 35: I would eagerly convert all these string paths to actual `Path` objects and do the manipulation in that form. Safer and you get nice manipulation/testing routines naturally.

dynamic-config/server/configuration-provider/src/main/java/org/terracotta/dynamic_config/server/configuration/util/InvokeConfigToolScript.java
```diff
@@ -0,0 +1,75 @@
...
+        if (path.isEmpty()) {
+            throw new AssertionError("cannot run config-tool script automatically as tc.install-root property" +
+                    " is not set");
+        }
+        String kitPath =  new File(path).getParent();
+        kitPath += "/tools/bin/config-tool";
```
chrisdennis, line 41: Once everything is a path you could extract this out to a static constant and make it:
```java
Path configTool = kitPath.resolve(CONFIG_TOOL);
```

dynamic-config/server/configuration-provider/src/main/java/org/terracotta/dynamic_config/server/configuration/util/InvokeConfigToolScript.java
```diff
@@ -0,0 +1,75 @@
...
+        boolean isWindows = System.getProperty("os.name").toLowerCase().startsWith("windows");
```
chrisdennis, line 42: You can roll this in to a static initializer since it's always going to be windows or not windows. Thats never going to change.

dynamic-config/server/configuration-provider/src/main/java/org/terracotta/dynamic_config/server/configuration/util/InvokeConfigToolScript.java
```diff
@@ -0,0 +1,75 @@
...
+        kitPath += isWindows ? ".bat" : ".sh";
+        ArrayList<String> cmd = (ArrayList<String>) Arrays.stream(command).collect(Collectors.toList());
```
chrisdennis, line 44: Probably easier to do something like:
```java
ProcessBuilder builder  = new ProcessBuilder(configTool);
builder.command().add(asList(command));
```

dynamic-config/server/configuration-provider/src/main/java/org/terracotta/dynamic_config/server/configuration/util/InvokeConfigToolScript.java
```diff
@@ -0,0 +1,75 @@
...
+        cmd.add(0, kitPath);
+        Process p = null;
```
chrisdennis, line 46: You can probably convert these resource loops to either:
```java
Resource r = new Resource();
try {
  ...
} finally {
  r.dispose();
}
```
or
```java
try (Resource r = new Resource()) {
  ...
}
```

dynamic-config/server/configuration-provider/src/main/java/org/terracotta/dynamic_config/server/configuration/util/InvokeConfigToolScript.java
```diff
@@ -0,0 +1,75 @@
...
+            p = new ProcessBuilder()
+                    .command(cmd)
+                    .redirectErrorStream(true)
+                    .start();
+            reader = new BufferedReader(new InputStreamReader(p.getInputStream(), StandardCharsets.UTF_8));
+            String line;
+            while ((line = reader.readLine()) != null) {
+                LOGGER.info(line);
+            }
+            int exitResult = p.waitFor();
```
chrisdennis, line 58: Why not just return the exit code and let the caller decide what should be logged? Might also want to allow the caller to feed in a `Consumer<String>` to do the logging. This way the logging target and level can be decided by the caller.

dynamic-config/server/configuration-provider/src/main/java/org/terracotta/dynamic_config/server/configuration/util/InvokeConfigToolScript.java
```diff
@@ -0,0 +1,75 @@
...
+        ArrayList<String> cmd = (ArrayList<String>) Arrays.stream(command).collect(Collectors.toList());
+        cmd.add(0, kitPath);
+        Process p = null;
+        BufferedReader reader = null;
+        try {
+            p = new ProcessBuilder()
+                    .command(cmd)
+                    .redirectErrorStream(true)
+                    .start();
+            reader = new BufferedReader(new InputStreamReader(p.getInputStream(), StandardCharsets.UTF_8));
```
chrisdennis, line 53: You probably want to use `Charset.defaultCharset()` here since both Java process are spawned on the same machine they should have the same default encoding and so this will be more 'correct'.

chrisdennis:
@mathieucarbou why did you mark this as a draft? Is there some reason I can't merge this once I approve it?

mathieucarbou:
> @mathieucarbou why did you mark this as a draft? Is there some reason I can't merge this once I approve it?

As discussed with @devlg75, this is because this is just a utility class sitting there, doing nothing, not used and its location is yet to be determined: as I understood, this code could be moved to EE later. There is no feature or anything "functionally" added there. Nothing is calling this class. There is no use case. So I have proposed to keep this PR as draft until someone really uses that because at that point, whether this code needs to be put there, or in EE, or refactored because it won't fit the need of the caller, etc... will become more evident.

dynamic-config/server/configuration-provider/src/main/java/org/terracotta/dynamic_config/server/configuration/util/InvokeConfigToolScript.java
devlg75: done.

dynamic-config/server/configuration-provider/src/main/java/org/terracotta/dynamic_config/server/configuration/util/InvokeConfigToolScript.java
devlg75: done

dynamic-config/server/configuration-provider/src/main/java/org/terracotta/dynamic_config/server/configuration/util/InvokeConfigToolScript.java
devlg75: done

dynamic-config/server/configuration-provider/src/main/java/org/terracotta/dynamic_config/server/configuration/util/InvokeConfigToolScript.java
devlg75: done

dynamic-config/server/configuration-provider/src/main/java/org/terracotta/dynamic_config/server/configuration/util/InvokeConfigToolScript.java
devlg75: yeah i think that is more correct, done

dynamic-config/server/configuration-provider/src/main/java/org/terracotta/dynamic_config/server/configuration/util/InvokeConfigToolScript.java
devlg75: done.

dynamic-config/server/configuration-provider/src/main/java/org/terracotta/dynamic_config/server/configuration/util/InvokeConfigToolScript.java
devlg75: done

dynamic-config/server/configuration-provider/src/main/java/org/terracotta/dynamic_config/server/configuration/util/InvokeConfigToolScript.java
```diff
@@ -0,0 +1,86 @@
...
+
+import static java.util.Arrays.asList;
+
+public class InvokeConfigToolScript {
+    private static final boolean IS_WINDOWS = Paths.get(System.getProperty("os.name").toLowerCase()).startsWith("windows");
+    private static final Path CONFIG_TOOL = Paths.get("tools/bin/config-tool" + (IS_WINDOWS ? ".bat" : ".sh"));
+
+    enum CommandList {
+        DETACH("detach"),
+        LOCK("lock-context");
```
ramsai1729, line 36: We need to unlock on server side, not lock, right?

ramsai1729: Perhaps we can provide higher level API instead of validating like this.

```
    int detach(InetSocketAddress connectTo, UUID leavingStripeID, UUID lockToken);
    
    int unlock(InetSocketAddress connectTo, UUID lockToken);
```

dynamic-config/server/configuration-provider/src/test/java/org/terracotta/dynamic_config/server/configuration/util/InvokeConfigToolScriptTest.java
```diff
@@ -0,0 +1,83 @@
...
+            invoke.execCommand(cmd, consumer);
+            fail();
+        } catch (Throwable t) {
+            String msg = "Cannot run config-tool script automatically as tc.install-root property is not set";
+            Assert.assertEquals(msg, t.getMessage());
+        }
+    }
+
+    @Test
+    public void testWhenScriptExecutionFails() {
```
ramsai1729, line 56: These tests are not testing much, maybe we can create dummy config-tool scripts and assert logging and exit status etc

dynamic-config/server/configuration-provider/src/main/java/org/terracotta/dynamic_config/server/configuration/util/InvokeConfigToolScript.java
devlg75: yeah will do that in next task while constructing the commands for  detach and unlock.

dynamic-config/server/configuration-provider/src/main/java/org/terracotta/dynamic_config/server/configuration/util/ConfigToolScriptExecutor.java
```diff
@@ -0,0 +1,65 @@
...
+import java.io.InputStreamReader;
+import java.nio.charset.Charset;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.function.Consumer;
+
+import static java.util.Arrays.asList;
+
+public class ConfigToolScriptExecutor {
+  private static final boolean IS_WINDOWS = Paths.get(System.getProperty("os.name").toLowerCase()).startsWith("windows");
```
chrisdennis, line 29: This isn't a path... it's just a string.

dynamic-config/server/configuration-provider/src/main/java/org/terracotta/dynamic_config/server/configuration/util/ConfigToolScriptExecutor.java
```diff
@@ -0,0 +1,65 @@
...
+  private static final Path CONFIG_TOOL = Paths.get("tools/bin/config-tool" + (IS_WINDOWS ? ".bat" : ".sh"));
```
chrisdennis, line 30: My understanding of the API here is that it is better to do this as an argument list:
```java
Paths.get("tools", "bin", "config-tool" + (IS_WINDOWS ? ".bat" : ".sh"))
```

dynamic-config/server/configuration-provider/src/main/java/org/terracotta/dynamic_config/server/configuration/util/ConfigToolScriptExecutor.java
devlg75: yeah, modified.

dynamic-config/server/configuration-provider/src/main/java/org/terracotta/dynamic_config/server/configuration/util/ConfigToolScriptExecutor.java
devlg75: done.

dynamic-config/server/configuration-provider/src/test/java/org/terracotta/dynamic_config/server/configuration/util/ConfigToolScriptExecutorTest.java
```diff
@@ -0,0 +1,89 @@
...
+      configToolScriptExecutor.execCommand(cmd, consumer);
+      fail();
+    } catch (Exception e) {
+      //do nothing
+    }
+  }
+
+  @Test
+  public void testDummyScriptExecutionAndLoggerConsumer() throws Exception {
+    System.setProperty("tc.install-root", "src" + File.separator + "test" + File.separator +
```
ramsai1729, line 79: I think its better to look up script resource instead of directly using it from the src directory itself, you might need to tweak test-resources processing config in the pom to set executable permission or you could copy the resources to tmp directory, change permissions and execute.

devlg75:
need to work on other task related to this PR so i am asking to merge this, 
if any concern will follow up in next PR.

PR #812, opened 2020-07-30
Bumped-up jackson + added validation for audit-log-dir

saurabhagas:


dynamic-config/model/src/main/java/org/terracotta/dynamic_config/api/service/ClusterValidator.java
```diff
@@ -199,4 +207,17 @@ private void validateSecurityDir() {
       }
     });
   }
+
+  private void validateAuditLogDir() {
+    List<String> nodesWithNoAuditLogDirs = cluster.getStripes()
+        .stream()
+        .flatMap(s -> s.getNodes().stream())
+        .filter(node -> !node.getSecurityAuditLogDir().asOptional().isPresent())
```
mathieucarbou, line 215: node.getSecurityAuditLogDir().isConfigured()

When using the new API, you shouldn't need to use asOptional at all except in some really special cases.

saurabhagas: Fixed

PR #815, opened 2020-08-04
[WIP] TDB-5080: Restricting node names to: [a-zA-Z0-9-_.:]+

mathieucarbou:
// Notes:
  // - For migration purpose, we need to support having DNS names in names (users used to specify dns names as names)
  // - For migration purpose, we need to support special characters like :.[] because server names were generated from IPv4 / IPv6 addresses,
  // so we could in theory have a generated server name (by the old XML parser) like: [2001:db8:85a3:8d3:1319:8a2e:370:7348]:9410

This PR still needs discussion or to be refined because it blocks other special characters like 💩


saurabhagas:
Let's discuss this today. The discussion would include things like:
1. The need for this PR, and the downside of not disallowing special characters
2. What characters we allow in names at other places (cache names, dataset names, connection names etc.)
3. ...

mathieucarbou:
Will reopen if needed

PR #816, opened 2020-08-05
[URG] TDB-5164: Broken repair and auto-activation with hidden settings (option 2)

mathieucarbou:
There is a problem with the fact of hiding some DC settings from the user (like the lock mechanism): it breaks the repair and sync capability.

For a node to be able to join an existing cluster after an automatic activation or after a repair, the node must have EXACTLY the same DC configuration as the destination cluster. This also includes settings like lock, uuids in the future, etc. This is done so that the node car join, identify itself accordingly and also start in the same "state" as other nodes.

To fix this, there were 2 possibilities:

1. Remove the notion of "hidden" settings and ALWAYS export ALL DC settings.
2. Add a `-a` flag to the export command which means: Export all configured DC settings including system/hidden ones

To be able to do a restricted activation, repair or any special the user is required to have a full exported configuration with `-a`.

**Notes:**
- I strongly prefer 1) and let the user see EVERYTHING also because all DC configurations, whether they are internal or not, brings information to the user. But this goes against what has been decided already (which I do not agree BTW because it leads to weird situations and other UX problems).
- 2) is also dangerous to do because if the user forget to use `-a` when exporting, he can mess up the repair and the node won't start, and the user won't have any feedback why.

### DIFF
--- a/dynamic-config/cli/config-tool/src/main/java/org/terracotta/dynamic_config/cli/config_tool/command/ExportCommand.java
+++ b/dynamic-config/cli/config-tool/src/main/java/org/terracotta/dynamic_config/cli/config_tool/command/ExportCommand.java
@@ -41,7 +41,7 @@ import java.util.Properties;
 import static java.lang.System.lineSeparator;
 
 @Parameters(commandNames = "export", commandDescription = "Export a cluster configuration")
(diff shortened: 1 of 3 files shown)


mathieucarbou:
talked to @jhouserizer : will open a new PR

PR #823, opened 2020-08-08
🔥 TDB-5176: DC config format versioning and backward compatibility support

mathieucarbou:
# ONLY MERGE IF ALL BUILDS PASS

This PR adds backward compatibility and versioning support:

1. **Support backward compatibility in Jackson modules**: added v1 jackson mixins which supports the old way of mapping the json attributes to the model

2. **Support backward compatibility when outputting configuration**: added the ability to output the config property file in different format version. All settings now declare in which versions they are used.

3. **Support backward compatibility when parsing configuration**: the configuration parser now supports parsing a V1 config and V2 config. It is also able to automatically "feed" the missing settings with some defaults (i.e. useful when starting from CLI or importing a V1 config using a V2 code)

4. **Support backward compatibility for Sanskrit**: this one was tricky. Since the config format changes, and it is serialized in sanskrit too, we need a marker within the append.log file to know the format for each entry, and use the corresponding mapper.

5. **Support backward compatibility for FileConfigStorage**: the config file saved into the config repository also now contains the format version of the config file.

6. **Support backward compatibility for HashComputer**: tricky one also... this refactoring allows to decide how the hash is computed and checked based on a config and a version now. We cannot check the hash of V1 anymore because of the way it was computed before (see more detailed comments in code around the jackson serialization being non deterministic even when using `SORT_PROPERTIES_ALPHABETICALLY` and `ORDER_MAP_ENTRIES_BY_KEYS`). Now, the hash is computed (and checked) for V2 and more based on the string representation of the property config.

7. **Support upgrading a config repository through a Nomad change**: since we need to be able to upgrade to a new format, this Nomad change will do just that. This nomad change contains the logic to upgrade a given config (adding stripe names, uuids, etc.)

Here is attached some examples of V1 and V2 property files and an append.log that shows a migration change.

You can also see the format change in the append.log

- [append.log](https://github.com/Terracotta-OSS/terracotta-platform/files/5064949/append.log) (bug before: formatting disappeared in Fix1 kit, and now `format version` is added)
- [default-node1.1.properties.txt](https://github.com/Terracotta-OSS/terracotta-platform/files/5064951/default-node1.1.properties.txt) (before upgrade)
```properties
#THIS FILE IS INTENDED FOR BOOK-KEEPING PURPOSES ONLY, AND IS NOT SUPPOSED TO BE EDITED. DO NOT ATTEMPT TO MODIFY.
cluster-name=tc-cluster
failover-priority=availability
stripe.1.node.1.data-dirs=main\:%H/terracotta/default-node1/user-data
stripe.1.node.1.hostname=localhost
stripe.1.node.1.log-dir=%H/terracotta/default-node1/logs
stripe.1.node.1.metadata-dir=%H/terracotta/default-node1/metadata
stripe.1.node.1.name=default-node1
this.name=default-node1
this.node-id=1
this.stripe-id=1
```
- [default-node1.2.properties.txt](https://github.com/Terracotta-OSS/terracotta-platform/files/5064953/default-node1.2.properties.txt) (after upgrade): notice the addition of `stripe-name` and `this.version`
```properties
#THIS FILE IS INTENDED FOR BOOK-KEEPING PURPOSES ONLY, AND IS NOT SUPPOSED TO BE EDITED. DO NOT ATTEMPT TO MODIFY.
cluster-name=tc-cluster
failover-priority=availability
stripe.1.node.1.data-dirs=main\:%H/terracotta/default-node1/user-data
stripe.1.node.1.hostname=localhost
stripe.1.node.1.log-dir=%H/terracotta/default-node1/logs
stripe.1.node.1.metadata-dir=%H/terracotta/default-node1/metadata
stripe.1.node.1.name=default-node1
stripe.1.stripe-name=stripe[0]
this.name=default-node1
this.node-id=1
this.stripe-id=1
this.version=2
```


mathieucarbou:
# PLEASE DO NOT MERGE IF WIN BUILDS ARE NOT PASSING

saurabhagas:
Didn't get to review it yet. Please allow some time. If you're confident of the changes and need the merged asap, please get it merged. I'll review it later and provide my comments.

PR #824, opened 2020-08-10
[WIP] TDB-5177: Modify attach and detach command to trigger re-balancing operation

devlg75:


dynamic-config/cli/config-tool/src/main/java/org/terracotta/dynamic_config/cli/config_tool/command/TopologyCommand.java
```diff
@@ -182,6 +192,21 @@ protected final void validateLogOrFail(Supplier<Boolean> expectedCondition, Stri
     }
   }
 
+  protected void acquireLockAndRunCommand() {
```
chrisdennis, line 195: I think you could dig this down at least one layer deeper. Could you directly instantiate and configure the `SetCommand` objects you need and execute those directly? Essentially strip off a few layers that we don't need to work through here.

dynamic-config/cli/config-tool/src/main/java/org/terracotta/dynamic_config/cli/config_tool/command/TopologyCommand.java
```diff
@@ -117,16 +123,20 @@ public final void run() {
...
     if (destinationClusterActivated) {
-      TopologyNomadChange nomadChange = buildNomadChange(result);
-      licenseValidation(destination, nomadChange.getCluster());
-      onNomadChangeReady(nomadChange);
-      logger.info("Sending the topology change");
-      try {
-        runTopologyChange(destinationCluster, destinationOnlineNodes, nomadChange);
-      } catch (RuntimeException e) {
-        onNomadChangeFailure(nomadChange, e);
+      if (operationType == OperationType.STRIPE && !isLockAwareNomadManager()) {
```
chrisdennis, line 126: This method is getting pretty gnarly at this point... lots of nesting and lots of disparate behaviors driven through one call. I do have a wider refactor that I think might address some of this but I'll put that in a different comment.

dynamic-config/cli/config-tool/src/main/java/org/terracotta/dynamic_config/cli/config_tool/command/TopologyCommand.java
devlg75: done

devlg75:
Dear All,
now we have overridden commands behavior  so it's more appropriate to put this change in EE, So i opened 2 PR's Equivalent to this, one is in OSS and other one in EE
https://github.com/Terracotta-OSS/terracotta-platform/pull/846

mathieucarbou:
> Dear All,
> now we have overridden commands behavior so it's more appropriate to put this change in EE, So i opened 2 PR's Equivalent to this, one is in OSS and other one in EE
> #846

Will you close this one ?

dynamic-config/cli/config-tool/src/main/java/org/terracotta/dynamic_config/cli/config_tool/command/TopologyCommand.java
devlg75: now i think it's resolved  after introducing managed service loader mechanism.

devlg75:
> > Dear All,
> > now we have overridden commands behavior so it's more appropriate to put this change in EE, So i opened 2 PR's Equivalent to this, one is in OSS and other one in EE
> > #846
> 
> Will you close this one ?

yeah sure

PR #826, opened 2020-08-11
Validating file permissions

mobasherul:


dynamic-config/model/src/main/java/org/terracotta/dynamic_config/api/model/SettingValidator.java
```diff
@@ -91,8 +95,12 @@
   static final BiConsumer<String, Tuple2<String, String>> PATH_VALIDATOR = (setting, kv) -> {
     DEFAULT_VALIDATOR.accept(setting, kv);
     try {
-      Paths.get(kv.t2);
-    } catch (RuntimeException e) {
+      Path p = Paths.get(kv.t2);
+      Set<PosixFilePermission> set = Files.getPosixFilePermissions(p);
```
mobasherul, line 99: How to handle path such as "%h-logs"? . These Files API can't handle it .

mathieucarbou: Never validate the content here: this code is comono both client side and server side and on client side you cannot resolve.
The place where you can get the confifj and property validate it after a resolve is inside the config handlers.
Look at the config handlers for data dirs for example

dynamic-config/model/src/main/java/org/terracotta/dynamic_config/api/model/SettingValidator.java
```diff
@@ -91,8 +95,12 @@
...
+      if (!set.contains(PosixFilePermission.OWNER_WRITE) || !set.contains(PosixFilePermission.OWNER_READ)) {
+        throw new RuntimeException();
+      }
+    } catch (RuntimeException | IOException e) {
```
mathieucarbou, line 103: Validations regarding resolved placeholders must happen in the config handlers (or elsewhere on server side)

saurabhagas:
This PR seems to be misguided. I _believe_ it was raised to fix the log dir validation. If that's true, I see several problems:
1. The validation needs to happen in the config change handler like @mathieucarbou said.
2. The above validation exists already, so no change there is needed (hopefully).
3. The log dir validation problem is seen currently on Windows only. So the focus should be tackling why that problem exists. Checking `PosixFilePermissions` will not help there.

PR #829, opened 2020-08-11
Bump-up versions

mathieucarbou:


saurabhagas:
@myronkscott do you think the failures have something to do with the changes done in https://github.com/Terracotta-OSS/terracotta-core/pull/1166?

PR #830, opened 2020-08-12
TDB-5115: Validating the existence of log paths: Replacing Files.exist() with File.exist()

mobasherul:


dynamic-config/server/services/src/main/java/org/terracotta/dynamic_config/server/service/handler/NodeLogDirChangeHandler.java
```diff
@@ -43,7 +43,7 @@ public void validate(NodeContext nodeContext, Configuration change) throws Inval
     }
 
     Path substitutedLogPath = substitute(Paths.get(logPath));
-    if (!Files.exists(substitutedLogPath)) {
+    if (!substitutedLogPath.toFile().exists()) {
```
mobasherul, line 46: File.exists(): Returns true if and only if the file or directory denoted by this abstract pathname exists; false otherwise.

Files.exists(): Returns true if the file exists; false if the file does not exist or its existence cannot be determined.

mobasherul: Tested and it works in windows now.

saurabhagas: Assuming that's true, we'd need to revisit all instances of `Files.exists`, and notify everyone to stop using it immediately.

Out of curiosity, has the behavior of `Files.exists` been fixed in newer Java versions?

Can you also add a test, if it's not too painful to write one using angela? A unit test will also do.

mobasherul: Files.notExists() also dont work well -
When the network share file actually exists

Files.exists(path): false
Files.notExists(path): false
path.toFile().exists(): true

mobasherul: For this - "Out of curiosity, has the behavior of Files.exists been fixed in newer Java versions?" I guess the anwer is they won't change it since it clearly states that it returns false if the existence can't be determined which is the case when there is no permission .

saurabhagas: @mathieucarbou please have a look and share your opinion in the light of my comments.

mathieucarbou: I would vote for replacing all 20 occurrences in the project that needs a more reliable check:
image

dynamic-config/testing/support/src/main/java/org/terracotta/dynamic_config/test_support/util/ConfigurationGenerator.java
```diff
@@ -39,7 +39,7 @@
   public interface PortSupplier {
     int getNodePort(int stripeId, int nodeId);
 
-    int getNodeGroupPort(int stripeId, int nodeId);
+    int getNodeGroupPort(int stripeId, int noFdeId);
```
saurabhagas, line 42: stray change

PR #836, opened 2020-08-14
add a timestamp to exported configs (at least in properties ones ... …

jhouserizer:
…json doesn't like comments)


dynamic-config/cli/config-tool/src/main/java/org/terracotta/dynamic_config/cli/config_tool/command/ExportCommand.java
```diff
@@ -119,6 +120,13 @@ private String buildOutput(Cluster cluster, OutputFormat outputFormat) {
         defaults.keySet().removeAll(userDefined.keySet());
         // write them all
         try (StringWriter out = new StringWriter()) {
+          // write a timestamp as a comment in the file header
+          out.write("# Timestamp of configuration export: ");
+          out.write(Instant.now().toString());
```
jhouserizer, line 125: this should output the date and time, in UTC/GMT TZ offset

mathieucarbou:
-t option is hidden : the user wont't and should not export to JSON

PR #837, opened 2020-08-15
🔥 TDB-5184: Fix JSON Path serializer/deserializer

mathieucarbou:
... and support backward compatibility with the old serialized format.

The old one was not able to correctly deserialize on Linux a Path serialized on Windows because `\\` is not treated as a path separator on linux.

**Explanations:**

This bug would cause:
- the config-tool to fail some operations if used on a linux node to access a windows node
- the passive sync process to fail if the active is a windows machine and passive is a linux machine

@jhouserizer @tmesic99 : FYI, I will open another PR to keep the user input (from config file and set command) as-is and treat differently the Path object we have in the DC model as Strings. This PR is to fix the general way of handling Path through JSON. It will also automatically fix the issue we saw but not the one related to the export which is not keeping the user input. There are 2 issues.


mathieucarbou:
@saurabhagas : pr ready to be merged

PR #838, opened 2020-08-16
🔥 TDB-5185: Introduce replacement for Path where we need to keep user input

mathieucarbou:
This PR introduces a `RawPath` object to replace Path in the model since we need to keep user input.

**IMPORTANT NOTES:**

With already existing users, the value in their config file from their config repository will be deserialized correctly and used.

Sadly this value was depending on how it was serialized (i.e. win users will have \\ in their paths in their config files and lin users will have / in their path). So when re-loading, depending on the OS of the node, the considered user input might be a/b or a\b. So this will obviously fail any passive sync in case the user had a mix environment, which, anyway, was not even possible because of the Path deserialization bug fixed in #837.

So we can consider that from now on, after this PR, all users will have their user input kept as-it regardless of the OS and mixed environments will be supported.


mathieucarbou:
Merging: 3/4 builds success, incl. win and lin. The only failure is with `test_detach_when_active_passive_disrupted ` (ignite timeout).

PR #840, opened 2020-08-17
Bumped up versions

saurabhagas:


saurabhagas:
@mathieucarbou bumping up galvan and core versions here so that they can become a part of the next release train. Given that we have a bug in M&M which is going downstream tests to fail, I'm guessing that the fix along with another release train will happen today itself.

PR #841, opened 2020-08-17
Reverting problematic changes in Settings class from commit 51cbcf304

saurabhagas:
The changes in this class cause failures in downstream branches. I'm only reverting the `Settings` (not sure if this is right). If they can be fixed in the proper way at the earliest (as we need to trigger a release train), this PR can be declined.


saurabhagas:
Closing PR in favor of https://github.com/Terracotta-OSS/terracotta-platform/pull/842

PR #844, opened 2020-08-18
[WIP] Attach node to stripe on secured cluster.

prasantakumar82:


prasantakumar82:
> Agree with the impl, but there are many more occurrences of calls to `fetchDiagnosticService()` to replace...

Do we need to follow unsecured path for other use cases? For example with diagnostic command, if user passing "security root directory" and if the security is not enabled in cluster, should the request fail instead of falling back to non-secured connection.

dynamic-config/cli/config-tool/src/main/java/org/terracotta/dynamic_config/cli/config_tool/command/RemoteCommand.java
```diff
@@ -431,14 +431,14 @@ protected final void ensureNodesAreEitherActiveOrPassive(Map<InetSocketAddress,
...
-    try (DiagnosticService diagnosticService = diagnosticServiceProvider.fetchDiagnosticService(expectedOnlineNode)) {
+    try (DiagnosticService diagnosticService = diagnosticServiceProvider.fetchDiagnosticServiceWithFallback(expectedOnlineNode)) {
       return diagnosticService.getProxy(TopologyService.class).isActivated();
     }
   }
 
   protected final void resetAndStop(InetSocketAddress expectedOnlineNode) {
     logger.info("Reset node: {}. Node will stop in 5 seconds", expectedOnlineNode);
-    try (DiagnosticService diagnosticService = diagnosticServiceProvider.fetchDiagnosticService(expectedOnlineNode)) {
+    try (DiagnosticService diagnosticService = diagnosticServiceProvider.fetchDiagnosticServiceWithFallback(expectedOnlineNode)) {
```
mathieucarbou, line 441: not needed

mathieucarbou:
> > Agree with the impl, but there are many more occurrences of calls to `fetchDiagnosticService()` to replace...
> 
> Do we need to follow unsecured path for other use cases? For example with diagnostic command, if user passing "security root directory" and if the security is not enabled in cluster, should the request fail instead of falling back to non-secured connection.

oh yes just the attach case with a mix of secured / non secured nodes is enough.

saurabhagas:
I have a few concerns about this PR. Let's huddle together briefly today and discuss this change.

mathieucarbou:
@prasantakumar82 : converting this PR to draft. This is not finished (see my comment) and requires to be tested in EE with a system test. Thanks!

PR #846, opened 2020-08-19
TDB-5177: introducing framework to run EE attach and detach commands

devlg75:


dynamic-config/server/configuration-provider/src/main/java/org/terracotta/dynamic_config/server/configuration/sync/DynamicConfigNomadSynchronizer.java
```diff
@@ -51,14 +53,19 @@ public DynamicConfigNomadSynchronizer(NodeContext nodeStartupConfiguration, Upgr
 
   public Set<Require> syncNomadChanges(List<NomadChangeInfo> sourceNomadChanges, Cluster sourceTopology) throws NomadException {
     List<NomadChangeInfo> nomadChanges = nomadServer.getAllNomadChanges();
+    NomadChange firstNomadChange = nomadChanges.get(0).getNomadChange();
```
saurabhagas, line 56: This part can be extracted to its own method. It will save you from having to create a `firstFinalNomadChange` variable.

saurabhagas (changes requested):

dynamic-config/server/configuration-provider/src/main/java/org/terracotta/dynamic_config/server/configuration/sync/DynamicConfigNomadSynchronizer.java
devlg75: agreed, done.

saurabhagas (approved):

PR #847, opened 2020-08-19
TDB-5071 - Generating Friendly Node Names

mobasherul:


dynamic-config/model/src/main/resources/dictionaries/sports.txt
```diff
@@ -0,0 +1,8 @@
+Cricket
+Football
+Badminton
+Tennis
+Hockey
+BasketBall
+VolleyBall
+BaseBall
```
mathieucarbou, line 8: this list has too few items IMO... dangerous...

dynamic-config/model/src/main/java/org/terracotta/dynamic_config/api/service/NameGenerator.java
```diff
@@ -0,0 +1,69 @@
...
+import java.util.concurrent.ThreadLocalRandom;
+
+public class NameGenerator {
+  private static int count = 0;
+  private static final String[] dictionary = {"animals.txt", "colors.txt", "space.txt", "continents.txt", "sports.txt"};
+  private static final String[] categories = {"Animal", "Color", "Space", "Continent", "Sport"};
+
+  public static String getStripeName() {
+    if (count == categories.length) {
+      count = 0;
```
mathieucarbou, line 36: The counter is going back to zero, no counter suffix is added, so this code only supports 5 stripes before failing since the stripe names will repeat.
Should we add another counter that will always be incremented and appended to the generated stripe name ?

dynamic-config/model/src/main/java/org/terracotta/dynamic_config/api/service/NameGenerator.java
```diff
@@ -0,0 +1,69 @@
...
+    }
+    StringBuilder builder = new StringBuilder(categories[count++]);
+    builder.append("-stripe");
```
mathieucarbou, line 39: @jhouserizer : appending this suffix is inconsistent with the new config format. The suer will then have to use: `config-tool set -c stripe:Sport-stripe ...`  ? What a redundancy here ;-)
I suggest we remove any suffix when generating names.

dynamic-config/model/src/main/java/org/terracotta/dynamic_config/api/service/NameGenerator.java
```diff
@@ -0,0 +1,69 @@
...
+
+  public static String getNodeName() {
+    if (count == categories.length) {
+      count = 0;
+    }
+    String fileName = dictionary[count];
+    InputStream inputStream = Setting.class.getClassLoader().getResourceAsStream("dictionaries/" + fileName);
+    try {
+      List<String> allLines = readFromInputStream(inputStream);
+      int ind = ThreadLocalRandom.current().nextInt(0, allLines.size());
```
saurabhagas, line 51: This has no guarantee that same name won't be chosen twice

dynamic-config/model/src/main/java/org/terracotta/dynamic_config/api/service/NameGenerator.java
```diff
@@ -0,0 +1,69 @@
...
+  }
+
+  public static String getNodeName() {
+    if (count == categories.length) {
+      count = 0;
+    }
+    String fileName = dictionary[count];
+    InputStream inputStream = Setting.class.getClassLoader().getResourceAsStream("dictionaries/" + fileName);
+    try {
+      List<String> allLines = readFromInputStream(inputStream);
```
saurabhagas, line 50: Seems overkill to read all the dictionaries every time this method is method.

dynamic-config/model/src/main/resources/dictionaries/space.txt
```diff
@@ -0,0 +1,13 @@
+Mercury
+Venus
+Earth
+Mars
+Jupiter
+Saturn
+Neptune
+Titan
+Europa
+Pandia
```
saurabhagas, line 10: Other constellations could be added too.

dynamic-config/model/src/main/resources/dictionaries/animals.txt
```diff
@@ -0,0 +1,21 @@
...
+Cat
+Cattle
+Dog
+Donkey
+Goat
+Guinea pig
+Horse
+Pig
+Rabbit
+Fancy rat varieties
```
saurabhagas, line 13: I'd vote for one-word names everywhere. This is unnecessarily complex. If we need more animal names, there will be more out there. Zoology has advanced significantly :-D

saurabhagas (changes requested):

dynamic-config/model/src/main/java/org/terracotta/dynamic_config/api/service/NameGenerator.java
mathieucarbou: That is OK. We just do not need that it gets chosen twice when names are generated. This only happens:
- when CLI si parsed for stripe name and node name
- when config is parsed for stripe name
- when xml is migrated for stripe name and node name
these are the only use cases

dynamic-config/model/src/main/java/org/terracotta/dynamic_config/api/service/NameGenerator.java
mathieucarbou: this is acceptable IMO considering the very few times it is called (https://github.com/Terracotta-OSS/terracotta-platform/pull/847#discussion_r472993689)

mobasherul: It won't read all the dictionaries . It will read only one of the file from dictionaries

dynamic-config/model/src/main/java/org/terracotta/dynamic_config/api/service/NameGenerator.java
jhouserizer: I talked with Haque. There was a misunderstanding about where stripe names and node come from, so this will be resolved by some changes he's planning.  (i.e. "space" and "continent" are not to be stripe names, the longer list of things within them are stripe names).

dynamic-config/model/src/main/java/org/terracotta/dynamic_config/api/service/NameGenerator.java
```diff
@@ -0,0 +1,69 @@
...
+      count = 0;
+    }
+    StringBuilder builder = new StringBuilder(categories[count++]);
+    builder.append("-stripe");
+    return builder.toString();
+  }
+
+  public static String getNodeName() {
+    if (count == categories.length) {
+      count = 0;
```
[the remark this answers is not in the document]
jhouserizer: again, this should be moot, with the changes Haque has now planned.

dynamic-config/model/src/main/resources/dictionaries/sports.txt
jhouserizer: again, not relevant issue with the changes planned.

dynamic-config/model/src/main/java/org/terracotta/dynamic_config/api/service/NameGenerator.java
jhouserizer: I have mixed feelings on this, but mostly disagree.   In the command line you showed, the word "stripe" is certainly redundant.  But elsewhere, in the properties file, in TMC screens, etc., the "-stripe" will add clarity.

mathieucarbou: yes, that's right!

dynamic-config/model/src/main/java/org/terracotta/dynamic_config/api/service/NameGenerator.java
saurabhagas: Ah right. But we'd need to handle the cases when we run out of names to give.

Also, does it give any value to store these names in files? If we're going to expose these files to users for them to add their own dictionaries, then it's a different story.

PR #852, opened 2020-08-20
UID -> UUID

mathieucarbou:


dynamic-config/model/src/main/java/org/terracotta/dynamic_config/api/model/UID.java
```diff
@@ -44,28 +59,40 @@ public boolean equals(Object o) {
 
   @Override
   public int hashCode() {
-    return Objects.hash(value);
+    return value.hashCode();
   }
 
   @Override
   public String toString() {
+    return encodeB64(toBytes(new long[]{value.getMostSignificantBits(), value.getLeastSignificantBits()}));
```
saurabhagas, line 67: I'd be curious to understand the need to base-64 encoding, and shortening the UUID. Can't we simply rely on UUID itself?

PR #848, opened 2020-08-19
Add support for force unlocking the config in repair command

ramsai1729:
This PR adds:

* support for force unlocking the config
* commands for locking/unlocking the config (useful for implementing force unlock as well), these are hidden commands.
* lock/unlock API in RemoteCommand, these apis can be used on client side and commands can be used on server side.

PR #856, opened 2020-08-25
[DO NOT MERGE] Handling invalid value against lock-context set command

prasantakumar82:


mathieucarbou:
@prasantakumar82 : discussed with @tmesic99 yesterday. This TDB has been closed. It will be obsoleted by #848 from @ramsai1729 . The user won't be able to set or unset a lock context with set unset commands.

Can you please close this PR ?

prasantakumar82:
@ramsai1729 can you add test similar to the one I added..
I have raised this PR so I can share the test with you.

@mathieucarbou I will delete the branch.

ramsai1729:
we already have tests to ensure set/unset doesn't work for lock-context, maybe a unit test to verify LockContext.from(...) method can be added.

PR #860, opened 2020-08-25
[URG] RELEASES + TDB-5071 (Fix for friendly name generation)

mathieucarbou:
Keep the uuid assigned names.
Otherwise node attachment or stripe attachment could fail with duplicate names.
And we do not know the full set of nodes in advance that will be activated


mathieucarbou:
I will run all the tests locally while the PR is building...

PR #862, opened 2020-08-27
Fixing TEST logback level (no runtime code change)

mathieucarbou:
WE HAVE TO SEE WARNINGS AND ERRORS!
Because of that, the stack traces dumped by core in case of a failure in the connection thread (i.e. codec) was hidden!


mathieucarbou:
merging directly - no harm

PR #865, opened 2020-08-28
Node can be attached to secured cluster only if it's started with security pre-configured

prasantakumar82:


saurabhagas (changes requested):
Please see Mathieu's comment.

saurabhagas:
> @prasantakumar82 @saurabhagas : is it the right way and what we have discussed ? I thought the idea now would be to override in EE the `ServiceProvider ` that was done in PR #863 to be able to replace the oss diagnostic service by the EE one (TDB-5209)
> 
> I am not really confortable in adding such checks in OSS... You'll tell me, there might be a good reason perhaps ;-)
> 
> I will let you check and confirm or discard this review...
> 
> Thanks!

Agree. This PR needs to be declined and an EE PR needs to be raised.

PR #866, opened 2020-08-31
Fixing ConfigSyncIT test failure.

mobasherul:


dynamic-config/testing/system-tests/src/test/java/org/terracotta/dynamic_config/system_tests/activated/ConfigSyncIT.java
```diff
@@ -143,8 +142,11 @@ public void testPassiveZapsAppendLogHistoryMismatch() throws Exception {
         is(throwing(instanceOf(RuntimeException.class))));
 
     //TODO TDB-4842: The stop and corresponding start is needed to prevent IOException on Windows
-    stopNode(1, passiveNodeId);
+
+    // wait for some time to ensure the Nomad Message gets applied to passive
+    Thread.sleep(5000);
```
mobasherul, line 147: getNode().getLoggerOverrides can't be done as the logger override is node's property . We are just writing about this nomad change in passive's append.log. That's why I am adding some sleep to ensure it gives time to replicate the NomadCommit message to passive.

PR #870, opened 2020-09-03
Misc. changes

saurabhagas:
See commit list.

Have a few more changes planned for this PR. Feel free to merge it if the changes are okay and builds are green, however.

### DIFF
--- a/dynamic-config/model/src/main/java/org/terracotta/dynamic_config/api/model/Requirement.java
+++ b/dynamic-config/model/src/main/java/org/terracotta/dynamic_config/api/model/Requirement.java
@@ -31,9 +31,9 @@ public enum Requirement {
   NODE_RESTART,
 
   /**
-   * Setting change needs only active servers to be online plus eventually some passive servers, but not all
+   * Setting change needs only active servers to be online plus the targeted objects (which can be nodes, stripes or the cluster)
    */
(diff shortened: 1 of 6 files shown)


dynamic-config/model/src/main/java/org/terracotta/dynamic_config/api/model/Requirement.java
```diff
@@ -31,9 +31,9 @@
   NODE_RESTART,
 
   /**
-   * Setting change needs only active servers to be online plus eventually some passive servers, but not all
+   * Setting change needs only active servers to be online plus the targeted objects which can be nodes, stripes or the cluster
    */
-  ACTIVES_ONLINE,
+  TARGET_ONLINE,
```
mathieucarbou, line 36: With an -S- ?

mathieucarbou (approved):

mathieucarbou:
Some tests need to be updated

dynamic-config/model/src/main/java/org/terracotta/dynamic_config/api/model/Requirement.java
saurabhagas: Done

saurabhagas:
> Some tests need to be updated

Done

PR #872, opened 2020-09-03
[WIP] TDB-5230: Transform ClusterValidator class into an overridable service both on client-side and server-side

mathieucarbou:
This PR is required for the upcoming validation fixes in OSS / EE (TDB-5194, TDB-5195, TDB-5207)


mathieucarbou:
Please hold your review: after thinking about it a Little more and talking to Myron I will change the service

PR #873, opened 2020-09-04
Method to check if security is configured.

prasantakumar82:
This small change is required to know if srd is passed or not.

### DIFF
--- a/diagnostic/client/src/main/java/org/terracotta/diagnostic/client/connection/DiagnosticServiceProvider.java
+++ b/diagnostic/client/src/main/java/org/terracotta/diagnostic/client/connection/DiagnosticServiceProvider.java
@@ -55,4 +55,8 @@ public class DiagnosticServiceProvider {
       throw new DiagnosticServiceProviderException(e);
     }
   }
+
+  public boolean isSecurityConfigured() {
+    return securityRootDirectory != null && !securityRootDirectory.isEmpty();
+  }
 }


diagnostic/client/src/main/java/org/terracotta/diagnostic/client/connection/DiagnosticServiceProvider.java
```diff
@@ -55,4 +55,8 @@ public DiagnosticService fetchDiagnosticService(InetSocketAddress address, Durat
       throw new DiagnosticServiceProviderException(e);
     }
   }
+
+  public boolean isSecurityConfigured() {
+    return securityRootDirectory != null && !securityRootDirectory.isEmpty();
```
saurabhagas, line 60: To check security, it's enough to check is the security root directory is null - you don't need to do an `isEmpty()` check. With that, you wouldn't need this change.

prasantakumar82: This method is used in enterprise code.

saurabhagas: Like I said, it's enough to check `securityRootDirectory != null` to check if security is or isn't configured. With that, the check becomes trivial, and the method you added will not be needed.

mathieucarbou:
not needed anymore

PR #877, opened 2020-09-09
[DO NOT MERGE] Replaced whitelist with "permit list"

saurabhagas:
The branch got deleted. Raising this PR again so that it can be used in the future.


mathieucarbou:
@jhouserizer : closing this PR because a replacement cannot be done anymore... If we have to change the wording, the changes required will be highly more complex now because of the required backward compatibility both in CLI, config file, config repository, client-server com layer.

PR #881, opened 2020-09-15
Code cleanup: removing unused hidden parameter: --entity-connection-timeout

mathieucarbou:


dynamic-config/cli/cli-support/src/main/java/org/terracotta/dynamic_config/cli/command/RemoteMainCommand.java
```diff
@@ -32,9 +32,6 @@
   @Parameter(names = {"-er", "--entity-request-timeout"}, hidden = true, description = "Entity operation timeout. Default: 120s", converter = TimeUnitConverter.class)
   private Measure<TimeUnit> entityOperationTimeout;
 
-  @Parameter(names = {"-et", "--entity-connection-timeout"}, hidden = true, description = "Entity Connection timeout. Default: 30s", converter = TimeUnitConverter.class)
```
mathieucarbou, line 35: this hidden parameter is never used

dynamic-config/cli/cli-support/src/main/java/org/terracotta/dynamic_config/cli/command/RemoteMainCommand.java
```diff
@@ -79,10 +72,5 @@ public void validate() {
           requestTimeout.getExactQuantity().multiply(BigInteger.valueOf(12)),
           requestTimeout.getUnit());
     }
-    if (entityConnectionTimeout == null) {
-      entityConnectionTimeout = Measure.of(
-          connectionTimeout.getExactQuantity().multiply(BigInteger.valueOf(3)),
-          connectionTimeout.getUnit());
```
mathieucarbou, line 85: and it was defaulted to the connectionTimeout

tmesic99 (approved):

dynamic-config/cli/cli-support/src/main/java/org/terracotta/dynamic_config/cli/command/RemoteMainCommand.java
chrisdennis: I assume we're declaring this safe (i.e. backwards compatible) because this is a hidden switch?

dynamic-config/cli/config-tool/src/main/java/org/terracotta/dynamic_config/cli/config_tool/command/OssServiceProvider.java
```diff
@@ -73,7 +73,7 @@ protected RestartService createRestartService(RemoteMainCommand mainCommand) {
   protected NomadEntityProvider createNomadEntityProvider(RemoteMainCommand mainCommand) {
     return new NomadEntityProvider(
         "CONFIG-TOOL",
-        getEntityConnectionTimeout(mainCommand),
+        getConnectionTimeout(mainCommand),
```
chrisdennis, line 76: Haven't you just changed the default used here from 30s to 10s?

mathieucarbou: yes! which is acceptable, especially that when the entity connection is made, there has been previously several connections made already so the hosts are likely to be up and reachable within 10 sec.

dynamic-config/cli/cli-support/src/main/java/org/terracotta/dynamic_config/cli/command/RemoteMainCommand.java
mathieucarbou: > I assume we're declaring this safe (i.e. backwards compatible) because this is a hidden switch?

yes for them we can they are not documented.

mathieucarbou:
merging. Linux Java8 test failure is caused by: java.net.BindException: Address already in use

PR #891, opened 2020-09-28
bump core version

myronkscott:


myronkscott:
fixes an issue in core where the oldest transaction id is not properly used on the server.

PR #893, opened 2020-09-29
TDB-5261-even-node-warnings-in-logs

tmesic99:
-fixing even-node/voter warning in server log to get logged once on startup and for any configuration change (previous behavior was buggy with every cluster validation call resulting in a log entry)
- fixed bug in Attach/Detach/Import commands not recognizing when consistency mode had non-zero voters configured

### DIFF
--- a/dynamic-config/cli/config-tool/src/main/java/org/terracotta/dynamic_config/cli/config_tool/command/AttachCommand.java
+++ b/dynamic-config/cli/config-tool/src/main/java/org/terracotta/dynamic_config/cli/config_tool/command/AttachCommand.java
@@ -41,6 +41,7 @@ import java.util.LinkedHashMap;
 import java.util.Map;
 
 import static java.lang.System.lineSeparator;
+import static org.terracotta.dynamic_config.api.model.FailoverPriority.Type.CONSISTENCY;
 import static org.terracotta.dynamic_config.api.model.FailoverPriority.consistency;
 import static org.terracotta.dynamic_config.cli.config_tool.converter.OperationType.NODE;
 import static org.terracotta.dynamic_config.cli.config_tool.converter.OperationType.STRIPE;
@@ -118,7 +119,7 @@ public class AttachCommand extends TopologyCommand {
 
       Stripe destinationStripe = destinationCluster.getStripeByNode(destination.getNodeUID()).get();
       FailoverPriority failoverPriority = destinationCluster.getFailoverPriority();
-      if (failoverPriority.equals(consistency())) {
+      if (failoverPriority.getType() == CONSISTENCY) {
         int voterCount = failoverPriority.getVoters();
         int nodeCount = destinationStripe.getNodes().size();
         int sum = voterCount + nodeCount;
--- a/dynamic-config/cli/config-tool/src/main/java/org/terracotta/dynamic_config/cli/config_tool/command/DetachCommand.java
+++ b/dynamic-config/cli/config-tool/src/main/java/org/terracotta/dynamic_config/cli/config_tool/command/DetachCommand.java
@@ -40,6 +40,7 @@ import java.util.ArrayList;
 import java.util.Collection;
 
 import static java.lang.System.lineSeparator;
+import static org.terracotta.dynamic_config.api.model.FailoverPriority.Type.CONSISTENCY;
 import static org.terracotta.dynamic_config.api.model.FailoverPriority.consistency;
 import static org.terracotta.dynamic_config.cli.config_tool.converter.OperationType.NODE;
 import static org.terracotta.dynamic_config.cli.config_tool.converter.OperationType.STRIPE;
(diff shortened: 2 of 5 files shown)


dynamic-config/cli/config-tool/src/main/java/org/terracotta/dynamic_config/cli/config_tool/command/AttachCommand.java
```diff
@@ -118,7 +119,7 @@ public void validate() {
 
       Stripe destinationStripe = destinationCluster.getStripeByNode(destination.getNodeUID()).get();
       FailoverPriority failoverPriority = destinationCluster.getFailoverPriority();
-      if (failoverPriority.equals(consistency())) {
+      if (failoverPriority != null && failoverPriority.getType().equals(CONSISTENCY)) {
```
mathieucarbou, line 122: enums in java have the same reference so then can be compared with `==` like primitives: `failoverPriority.getType() == CONSISTENCY`

dynamic-config/cli/config-tool/src/main/java/org/terracotta/dynamic_config/cli/config_tool/command/DetachCommand.java
```diff
@@ -92,7 +93,7 @@ public void validate() {
       }
 
       FailoverPriority failoverPriority = destinationCluster.getFailoverPriority();
-      if (failoverPriority.equals(consistency())) {
+      if (failoverPriority != null && failoverPriority.getType().equals(CONSISTENCY)) {
```
mathieucarbou, line 96: - failoverPriority is not optional so no need for the null check
- `==` instead of `equals`

dynamic-config/cli/config-tool/src/main/java/org/terracotta/dynamic_config/cli/config_tool/command/ImportCommand.java
```diff
@@ -58,7 +59,7 @@
   public void validate() {
     cluster = loadCluster();
     FailoverPriority failoverPriority = cluster.getFailoverPriority();
-    if (failoverPriority.equals(consistency())) {
+    if (failoverPriority != null && failoverPriority.getType().equals(CONSISTENCY)) {
```
mathieucarbou, line 62: - failoverPriority is not optional so no need for the null check
- `==` instead of `equals`

dynamic-config/server/configuration-provider/src/main/java/org/terracotta/dynamic_config/server/configuration/service/DynamicConfigServiceImpl.java
```diff
@@ -21,12 +21,7 @@
 import com.fasterxml.jackson.databind.ObjectMapper;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-import org.terracotta.dynamic_config.api.model.Cluster;
-import org.terracotta.dynamic_config.api.model.License;
-import org.terracotta.dynamic_config.api.model.Node;
-import org.terracotta.dynamic_config.api.model.NodeContext;
-import org.terracotta.dynamic_config.api.model.Stripe;
-import org.terracotta.dynamic_config.api.model.UID;
+import org.terracotta.dynamic_config.api.model.*;
```
mathieucarbou, line 24: you should configure your IDEA to not replace a serie of imports by star by putting a big number here:
image

dynamic-config/server/configuration-provider/src/main/java/org/terracotta/dynamic_config/server/configuration/service/DynamicConfigServiceImpl.java
```diff
@@ -343,6 +342,8 @@ public synchronized void activate(Cluster maybeUpdatedCluster, String licenseCon
 
     clusterActivated = true;
     LOGGER.info("Node activation successful");
+
+    warnIfProblematicConsistency(runtimeNodeContext);
```
mathieucarbou, line 346: This is safer and more true to use `upcomingNodeContext` here.

dynamic-config/server/configuration-provider/src/main/java/org/terracotta/dynamic_config/server/configuration/service/DynamicConfigServiceImpl.java
```diff
@@ -492,4 +493,25 @@ public void run() {
       }
     }.start();
   }
+
+  private void warnIfProblematicConsistency(NodeContext nodeContext) {
+    FailoverPriority failover = nodeContext.getCluster().getFailoverPriority();
+    if (failover != null && failover.getType().equals(CONSISTENCY)) {
```
mathieucarbou, line 499: - failover will always be there
- use `==` to compare enums

mathieucarbou (changes requested):

mathieucarbou (changes requested):
@tmesic99 : please also rename your commit message to a more descriptive sentence like: `TDB-5261: Fix display frequency and location of even node warning messsage`

PR #896, opened 2020-10-02
TDB-5247: Refactor the old UpgradableNomadServer and NomadChangeInfo to be able to expose the change result hash

mathieucarbou:
- Introduced a DynamicConfigNomadServer interface to hold all the DC related enhancements
- Remove NomadChangeInfo from Nomad module: this class is DC-centric and will hold soon the result hash
- Remove checkpoint information: checkpoints are not possible anymore through UUID since they can differ from node to node

There is no functional code change / added / removed here: just some refactoring to allow for the result hash to be exposed correctly and to decouple the NomadChangeInfo from the core Nomad module.

### DIFF
--- a/common/nomad/src/main/java/org/terracotta/nomad/messages/DiscoverResponse.java
+++ b/common/nomad/src/main/java/org/terracotta/nomad/messages/DiscoverResponse.java
@@ -15,11 +15,9 @@
  */
 package org.terracotta.nomad.messages;
 
-import org.terracotta.nomad.server.NomadChangeInfo;
 import org.terracotta.nomad.server.NomadServerMode;
 
 import java.time.Instant;
-import java.util.List;
 
 import static java.util.Objects.requireNonNull;
 
@@ -32,7 +30,6 @@ public class DiscoverResponse<T> {
   private final long currentVersion;
   private final long highestVersion;
   private final ChangeDetails<T> latestChange;
-  private final List<NomadChangeInfo> checkpoints;
 
   public DiscoverResponse(NomadServerMode mode,
                           long mutativeMessageCount,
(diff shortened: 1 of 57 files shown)


common/nomad/src/main/java/org/terracotta/nomad/messages/DiscoverResponse.java
```diff
@@ -51,17 +47,12 @@ public DiscoverResponse(NomadServerMode mode,
     this.currentVersion = currentVersion;
     this.highestVersion = highestVersion;
     this.latestChange = latestChange;
-    this.checkpoints = checkpoints;
```
mathieucarbou, line 54: - checkpoints removed (not possible anymore based on UUID)
- NomadChangeInfo class removed from Nomad module

common/nomad/src/main/java/org/terracotta/nomad/server/NomadServer.java
```diff
@@ -22,7 +22,13 @@
 import org.terracotta.nomad.messages.RollbackMessage;
 import org.terracotta.nomad.messages.TakeoverMessage;
 
+import java.util.Optional;
+
 public interface NomadServer<T> extends AutoCloseable {
```
mathieucarbou, line 27: hasIncompleteChange, getCurrentCommittedConfig, reset moved to NomadServer inetrface

common/nomad/src/main/java/org/terracotta/nomad/server/NomadServerImpl.java
```diff
@@ -44,7 +40,7 @@
 import static org.terracotta.nomad.server.ChangeRequestState.PREPARED;
 import static org.terracotta.nomad.server.ChangeRequestState.ROLLED_BACK;
 
-public class NomadServerImpl<T> implements UpgradableNomadServer<T> {
+public class NomadServerImpl<T> implements NomadServer<T> {
```
mathieucarbou, line 43: methods moved to DynamicConfigNomadServerImpl

PR #898, opened 2020-10-03
[WIP] TDB-5247 (#2) - expose change result hash

mathieucarbou:


common/nomad/src/main/java/org/terracotta/nomad/server/NomadServerImpl.java
```diff
@@ -110,24 +110,25 @@ public boolean hasIncompleteChange() {
...
     if (latestChangeUuid != null) {
-      ChangeRequest<T> changeRequest = state.getChangeRequest(latestChangeUuid);
-      ChangeRequestState changeState = changeRequest.getState();
-      long changeVersion = changeRequest.getVersion();
-      NomadChange change = changeRequest.getChange();
-      T changeResult = changeRequest.getChangeResult();
-      String changeCreationHost = changeRequest.getCreationHost();
-      String changeCreationUser = changeRequest.getCreationUser();
-      Instant changeCreationTimestamp = changeRequest.getCreationTimestamp();
+      ChangeState<T> changeState = state.getChangeState(latestChangeUuid);
```
mobasherul, line 113: why creating new class ChangeState can't it be solved using just adding a field in ChangeRequest class or is it something for backward compatibility.

mathieucarbou: Because the change request only contains the requested change by the user. At this point, in nomad API, there is nothing to enable to compute a hash. Hash computation happens in sanskrit (persistence layer) and needs to deal with version compatibility. So we can only have a hash in the object that is read from persistence, not saved initially.

PR #919, opened 2020-12-10
Reduce coupling between JCommander commands and the actions they trigger

chrisdennis:


dynamic-config/cli/config-tool/src/main/java/org/terracotta/dynamic_config/cli/config_tool/command/OssCommandProvider.java
```diff
@@ -15,67 +15,67 @@
...
-    commands.add(new RepairJCommanderCommand());
-    commands.add(new DeprecatedRepairJCommanderCommand());
-    commands.add(new LogJCommanderCommand());
-    commands.add(new DeprecatedLogJCommanderCommand());
-    commands.add(new LockConfigJCommanderCommand());
-    commands.add(new DeprecatedLockConfigJCommanderCommand());
-    commands.add(new UnlockConfigJCommanderCommand());
-    commands.add(new DeprecatedUnlockConfigJCommanderCommand());
+  public Set<Command> getCommands() {
+    Set<Command> commands = new HashSet<>();
```
mathieucarbou, line 52: I was thinking yesterday about the extensibility of this thing and I do not really like that. Having a bag of items makes defining a different flavour harder because we would need to copy/paste all of this again, or manipulate the collection and try to find the right item to remove...

I am wondering if we should instead extract methods (like it is done in `OssServiceProvider`: 1 method for each service), or return a Map where the key would be the action for example ? or command ?

mobasherul: Instead of functions for each command map would be better if we are taking this route.

chrisdennis: I think the JCommander aligned approach here would be to make them a list?

dynamic-config/cli/dynamic-config-cli-api/src/main/java/org/terracotta/dynamic_config/cli/api/command/LocalConfig.java
```diff
@@ -24,7 +24,7 @@
 
 import static org.slf4j.Logger.ROOT_LOGGER_NAME;
 
-public class LocalConfig extends Command {
+public class LocalConfig implements Runnable {
```
mathieucarbou, line 27: Could you take this opportunity to remove `implements Runnable` from `LocalConfig ` ?

LocalConfig and RemoteConfig are now just configuration classes passed to the service provider to create the services.

The problem is that they are also used to execute something (setup of the logging system). Which is wrong.

We could then introduce a `MainAction` that would be used in the `LocalMainCommand` and a `MainRemoteAction` that would be used in `RemoteMainCommand`, to match what is done elsewhere.

dynamic-config/cli/cli-support/src/main/java/org/terracotta/dynamic_config/cli/command/LocalMainCommand.java
```diff
@@ -17,26 +17,32 @@
...
-public class LocalMainJCommanderCommand extends JCommanderCommand {
+@Parameters(commandNames = LocalMainCommand.NAME)
+public class LocalMainCommand extends Command {
   public static final String NAME = "main";
 
   @Parameter(names = {"-verbose", "-v", "--verbose"}, description = "Verbose mode. Default: false")
   public boolean verbose = false;
 
-  private final LocalConfig underlying = new LocalConfig();
+  private final LocalConfig config;
```
mathieucarbou, line 29: `private final MainAction action;` (see my other comment)

dynamic-config/cli/config-tool/src/main/java/org/terracotta/dynamic_config/cli/config_tool/ConfigTool.java
```diff
@@ -80,7 +80,7 @@ public static void start(String... args) {
         return true;
       } else {
         LOGGER.debug("Injecting services in specified command");
-        Injector.inject(command.getCommand(), services);
+        Injector.inject(command, services);
```
mathieucarbou, line 83: 👍

dynamic-config/cli/cli-support/src/main/java/org/terracotta/dynamic_config/cli/command/LocalMainCommand.java
```diff
@@ -17,26 +17,32 @@
...
+  public LocalMainCommand(LocalConfig action) {
+    this.config = action;
+  }
 
   @Override
   public void run() {
-    underlying.setVerbose(verbose);
-    underlying.run();
+    getConfiguration().setVerbose(verbose);
+    getConfiguration().run();
```
mathieucarbou, line 42: related to my other comment... Or we could just remove the logging code from LocalConfig and just put it there...

dynamic-config/cli/dynamic-config-cli-api/src/main/java/org/terracotta/dynamic_config/cli/api/command/Injector.java
```diff
@@ -36,12 +36,17 @@
...
         .filter(field -> field.isAnnotationPresent(Inject.class))
         .forEach(field -> {
-          Object found = services.stream()
-              .filter(service -> field.getType().isInstance(service))
-              .findAny()
-              .orElseThrow(() -> new IllegalStateException("No service found to inject into " + field));
           try {
-            field.set(target, found);
+            Object targetValue = field.get(target);
+            if (targetValue == null) {
```
mathieucarbou, line 41: if we check for null, why annotating action fields with `@Inject` ? These fields will never be null so won't be injected ?

```
  @Inject
  public final ImportAction action;

  public ImportCommand() {
    this(new ImportAction());
  }
```

We are controlling the commands through service override, so why not just remove `@Inject` in then and make the fields private ?

mathieucarbou (changes requested):
@chrisdennis also, I saw at least one unused import (in RemoteMainCommand I think).
You can do a global format / reorganize imports in theory in this project: all the team is suing the same settings so that should work.

dynamic-config/cli/cli-support/src/main/java/org/terracotta/dynamic_config/cli/command/LocalMainCommand.java
chrisdennis: Done

dynamic-config/cli/dynamic-config-cli-api/src/main/java/org/terracotta/dynamic_config/cli/api/command/Injector.java
chrisdennis: Clarified.

dynamic-config/cli/cli-support/src/main/java/org/terracotta/dynamic_config/cli/command/LocalMainCommand.java
```diff
@@ -13,22 +13,28 @@
...
-public class LocalConfig extends Command {
-  private boolean verbose = false;
+@Parameters(commandNames = LocalMainCommand.NAME)
+public class LocalMainCommand extends Command {
+  public static final String NAME = "main";
+
+  @Parameter(names = {"-verbose", "-v", "--verbose"}, description = "Verbose mode. Default: false")
+  public boolean verbose = false;
+
+  public LocalMainCommand() {
```
mathieucarbou, line 36: can be removed

mathieucarbou (changes requested):

mathieucarbou (approved):

mathieucarbou (approved):

PR #926, opened 2021-01-19
Enhanced error message when attempting secured/unsecured connections

tmesic99:
For TDB-5163:
All Config Tool client->server commands utilize DiagnosticServiceProvider.fetchDiagnosticService
which in turn calls DiagnosticServiceFactory.fetch
which in turn calls ConnectionFactory.connect
which throws the 'handshake failed' exception.  
DiagnosticServiceProvider.fetchDiagnosticService is the first opportunity for the client to catch the exception - hence the change.  Let me know if there is a better place to inject this exception logic.

### DIFF
--- a/diagnostic/client/src/main/java/org/terracotta/diagnostic/client/connection/DiagnosticServiceProvider.java
+++ b/diagnostic/client/src/main/java/org/terracotta/diagnostic/client/connection/DiagnosticServiceProvider.java
@@ -52,6 +52,23 @@ public class DiagnosticServiceProvider {
     try {
       return DiagnosticServiceFactory.fetch(address, connectionName, connectTimeout, diagnosticInvokeTimeout, securityRootDirectory, objectMapperFactory);
     } catch (ConnectionException e) {
+      if (e.getMessage().contains("Client was able to establish connection with server but handshake with server failed")) {
+        String warning;
+        if (this.securityRootDirectory == null)
+        {
+          warning = "The client used a non-secure connection to connect with server '" + address.toString() + "'. " +
+                  "The connection was successful but the handshake failed. " +
+                  "This failure can occur if the server was configured to use secured connections. ";
+        }
+        else
+        {
+          warning = "The client used a secure connection to connect with server '" + address.toString() + "'. " +
+                  "The connection was successful but the handshake failed. " +
+                  "This failure can occur if the server was configured to not use secured connections. ";
+        }
+        warning += "Check the client and server logs for more information.";
+        throw new DiagnosticServiceProviderException(new Throwable(warning));
+      }
       throw new DiagnosticServiceProviderException(e);
     }
   }


diagnostic/client/src/main/java/org/terracotta/diagnostic/client/connection/DiagnosticServiceProvider.java
```diff
@@ -52,6 +52,23 @@ public DiagnosticService fetchDiagnosticService(InetSocketAddress address, Durat
     try {
       return DiagnosticServiceFactory.fetch(address, connectionName, connectTimeout, diagnosticInvokeTimeout, securityRootDirectory, objectMapperFactory);
     } catch (ConnectionException e) {
+      if (e.getMessage().contains("Client was able to establish connection with server but handshake with server failed")) {
+        String warning;
+        if (this.securityRootDirectory == null)
+        {
```
mathieucarbou, line 58: needs a reformat

diagnostic/client/src/main/java/org/terracotta/diagnostic/client/connection/DiagnosticServiceProvider.java
```diff
@@ -52,6 +52,23 @@ public DiagnosticService fetchDiagnosticService(InetSocketAddress address, Durat
     try {
       return DiagnosticServiceFactory.fetch(address, connectionName, connectTimeout, diagnosticInvokeTimeout, securityRootDirectory, objectMapperFactory);
     } catch (ConnectionException e) {
+      if (e.getMessage().contains("Client was able to establish connection with server but handshake with server failed")) {
```
mathieucarbou, line 55: This is one way to solve the issue, which is also used at another place.

But I am wondering if it is the right way to fix that, because the exception message could be changed in EE without us knowing.

In EE, we are using a `EnterpriseDiagnosticProvider` which overrides the OSS one. The original idea was to have this sort of code in EE in EnterpriseDiagnosticProvider.

Also, the advantage of using the EnterpriseDiagnosticProvider is that we have access to the SecurityException class that is thrown and contain the message you are checking.

mathieucarbou: @mobasherul : WDYT ?

mathieucarbou (changes requested):

diagnostic/client/src/main/java/org/terracotta/diagnostic/client/connection/DiagnosticServiceProvider.java
tmesic99: @mathieucarbou: I moved that same code block to the EnterpriseDiagnosticService and it behaves essentially the same way (but it's still using the approach to match the exception message - not ideal, I totally agree).
But note that the 'handshake failure' gets thrown initially by core as a RuntimeException and gets rethrown as just an Exception.  As far as I can see, there are no SecurityException being used anywhere (which is why I did the string message match).  Maybe I'm missing that.

mathieucarbou: ok. Let's look at that tomorrow morning with @mobasherul. There is an `ExceptionUtils` class tha twe can use for that I think that already contains some logic.

tmesic99:
Closing PR.  Implementation will be done in EE.

PR #928, opened 2021-01-22
Added Setting#vetoRuntimeChange() to allow a setting to veto a change that would be applied at runtime

mathieucarbou:


dynamic-config/cli/dynamic-config-cli-api/src/main/java/org/terracotta/dynamic_config/cli/api/command/ConfigurationMutationAction.java
```diff
@@ -138,6 +139,14 @@ public void run() {
             "=======================================================================================" + lineSeparator() +
             "IMPORTANT: A restart of nodes: " + toString(nodesRequiringRestart) + " is required to apply the changes" + lineSeparator() +
             "=======================================================================================" + lineSeparator());
+      } else {
+        Collection<String> updatedNodes = onlineNodes.keySet().stream().filter(this::mustBeRestarted).map(Endpoint::toString).collect(Collectors.toSet());
```
mobasherul, line 143: since we have to explicitly check all online nodes whether they need to be restarted because of this veto feature we can remove that block above line142 - 
 else if(!nodesRequiringRestart.isEmpty()) .WYT ?

mathieucarbou: even if this is more "general" the idea behind that is:

- if a setting is marked as requiring a restart (i.e. log dir), then we will go in the code block above, which will display a more detailed message with only the nodes targeted by the change.

- if we have a setting (like backup dir) which is not marked as requiring a restart in the definition, but the code has vetoed the runtime change (i.e. unset), then it won't be in the list of node requiring restart so the only way we can display a message is to loop over the online nodes to see which ones require a restart.

This has 2 drawbacks:
1) we have to connect to all the online nodes - so slow down the CLI a little
2) if a change was made before which require a restart (i.e. node log dir), then even if the last change does not require a restart, the warning will be displayed because a previous change was made and required a restart.

So this is mostly a fallback... In case the CLI is not able to determine if some restart are needed (with the provided changes), then it will look at all the online nodes.

Of course we can discuss and change this logic if we find something better :-) I didn't have any better idea yesterday than doing that.

@tmesic99 : FYI ^^

tmesic99: I think the overall approach is good: check for cluster restart, then NODE_RESTARTs and then veto'd node restarts.  Connecting to all online nodes to make the final determination shouldn't be that bad of a performance hit.

But should the logic of the if-elseif-else block be refactored to handle a set operation involving multiple settings that are applicable to both the NODE_RESTART else-if block and the 'updatedNodes' else block?  For example, if you set both the log-dir and backup-dir in the same set command, the NODE_RESTART else-if block will get hit (because of the presence of the log-dir setting) but the backup-dir 'updatedNodes else block will not run.  No?

mathieucarbou: You are right Tom!
In this case we should display one message containing the union of the nodes requiring a restart, whether it is because of NODE_RESTART or because of a veto. I will change!

tmesic99 (approved):

PR #931, opened 2021-02-01
TDB-5263: Warn user to ask for a restart on node-backup-dir update

mathieucarbou:


dynamic-config/cli/dynamic-config-cli-api/src/main/java/org/terracotta/dynamic_config/cli/api/command/ConfigurationMutationAction.java
```diff
@@ -66,20 +66,19 @@ public void run() {
...
+      targets.stream()
+          .filter(o -> o.getScope() == NODE)
+          .map(PropertyHolder::getName)
+          .filter(originalCluster::containsNode)
+          .forEach(name -> {
+            targetedNodes.add(name);
+            if (c.getSetting().requires(NODE_RESTART)) {
+              nodesRequiringRestart.add(name);
+            }
+          });
```
mathieucarbou, line 81: there was an issue here

dynamic-config/cli/dynamic-config-cli-api/src/main/java/org/terracotta/dynamic_config/cli/api/command/GetAction.java
```diff
@@ -53,9 +53,7 @@ public void run() {
           .filter(e -> configuration.matchConfigPropertyKey(e.getKey()))
           .map(e -> e.getKey() + "=" + e.getValue())
           .collect(Collectors.toList());
-      if (out.isEmpty()) {
-        output.out(configuration + "=");
```
mathieucarbou, line 57: this was a bug because it would output wrong setting like `backuo-dir=` (without `stripe.1.node.1.` prefix)

dynamic-config/model/src/main/java/org/terracotta/dynamic_config/api/model/Setting.java
```diff
@@ -447,17 +449,21 @@
...
         // the user wants to activate the backup feature).
         // - if the user wants to unset the directory or change it, we require the node to be restarted because
         // it could happen that a backup is already in progress
-        boolean isSetOrUpdate = configuration.hasValue();
-        boolean alreadyConfigured = currentNode.getNode().getBackupDir().isConfigured();
-        return alreadyConfigured || !isSetOrUpdate;
+        final String current = currentNode.getNode().getBackupDir().map(RawPath::getValue).orElse(null);
+        vetoed = !configuration.getValue().isPresent() // unset
+            || current != null && !Objects.equals(configuration.getValue().get(), current); // update
+      } else {
```
mathieucarbou, line 462: there was an issue in the logic above - we must not veto if the value is the same

dynamic-config/server/configuration-provider/src/main/java/org/terracotta/dynamic_config/server/configuration/service/DynamicConfigServiceImpl.java
```diff
@@ -253,7 +253,7 @@ public void onNomadCommit(CommitMessage message, AcceptRejectResponse response,
           upcomingNodeContext = upcomingNodeContext.withCluster(upcomingCluster).orElseGet(upcomingNodeContext::alone);
           // if the change can be applied at runtime, it was previously done in the config change handler.
           // so update also the runtime topology there
-          if (nomadChange.canUpdateRuntimeTopology(upcomingNodeContext)) {
+          if (nomadChange.canUpdateRuntimeTopology(runtimeNodeContext)) {
```
mathieucarbou, line 256: there was a big here

tmesic99 (approved):

dynamic-config/cli/dynamic-config-cli-api/src/main/java/org/terracotta/dynamic_config/cli/api/command/GetAction.java
tmesic99: An observation I'm seeing in some tests I'm creating:  If a user runs 'get' for several  settings, then only the settings with configured values or default values are returned in the output.  And we assume the user should interpret the 'missing' ones as settings which do not possess any values (configured or default).  
And if ALL of settings supplied to 'get' do not have any configured values, then the command will not return anything.  Now, they can either interpret that in the same manner as previous (that the setting do not have any values configured) or that something has gone wrong (since the output doesn't explicitly state that 'no configured values were found').   Are we 'ok' with the tool not explicitly stating that no configured value or default value was found for the specified setting?

mathieucarbou: So for the answer, I don't know...

But I have removed that because the display was wrong. In case a setting was not set, it was displayed as the user input without considering the scope of the setting, which could lead to some issues if the output is piped to something processing the output. If the user does a `get backup-dir`, then `backup-dir=` is displayed, which is wrong because backup-dir is a node scope settings.

If a setting were to be set on some nodes but not others (i.e. tc-property), then get node- properties would get all the props for all the nodes on the cluster, scoped and only display those set.

I think the right behaviour might be to extract from the list of settings in the get command those without anything set.

 I.e.
 
**1) if the user does `get -c failover-priority -c node-backup-dir` and only failover-priority is set, then we display:**

STDOUT:  
```
failover-priority=....
```

STDERR: 
```
backup-dir is not set on any node
```

**2) if the user does `get -c failover-priority -c node-tc-properties` and failover-priority is set, and some props are set on only 1 node:**

STDOUT: 
```
failover-priority=....
stripe.1.node.2.tc-properties=foo:bar
```
(nothing regarding node 1)

STDERR: 
```
<nothing>
```

**2) if the user does `get -c failover-priority -c node-tc-properties` and some props are set on ALL THE nodes:**

STDOUT: 
```
failover-priority=....
stripe.1.node.1.tc-properties=foo:bar
stripe.1.node.2.tc-properties=foo:bar
```

STDERR: 
```
<nothing>
```

This is what I would propose.

WDYT @jhouserizer  ?

PR #935, opened 2021-02-10
Adding AttachCommand tests with voter during failover.

mobasherul:


dynamic-config/testing/system-tests/src/test/java/org/terracotta/dynamic_config/system_tests/activated/AttachCommandWithMultipleVoter1x3IT.java
```diff
@@ -0,0 +1,111 @@
...
+      String propertySettingString = "stripe.1.node." + activeId + ".tc-properties.failoverAddition=killAddition-commit";
+      startNode(1, 3);
+      waitForDiagnostic(1, 3);
+      assertThat(getUpcomingCluster("localhost", getNodePort(1, 3)).getNodeCount(), is(equalTo(1)));
+
+      //setup for failover in commit phase on active
+      invokeConfigTool("set", "-s", "localhost:" + getNodePort(1, 1), "-c", propertySettingString);
+
+      invokeConfigTool("attach", "-f", "-d", "localhost:" + getNodePort(1, activeId), "-s", "localhost:" + getNodePort(1, 3));
+      waitForPassive(1, 3);
```
mathieucarbou, line 89: with the failover, if I remember, what we saw:
- the passive was not able to become active
- the passive node that is attached was "waiting" for the active to come to be able to sync, so it was not able to vote
- the 3rd server (old active) is down.

So I think the conclusion was:
- to add enough voters (3 ?) to be able to vote so that the active can be come active without any passive server
- then the attached node could sync and become passive
- the server that is down could then be restarted

So the `waitForPassive(1, 3);` should be placed after the voter have voted, and before `waitForPassive(1, 3);` we should wait for the active server to be there.

@mobasherul @chrisdennis : please correct me if I am wrong ;-)

mathieucarbou (changes requested):

dynamic-config/testing/system-tests/src/test/java/org/terracotta/dynamic_config/system_tests/activated/AttachCommandWithMultipleVoter1x3IT.java
mobasherul: Its not required now since we have two voters so when active goes down then also old passive will become active and  new attached node will be able to sync and become passive.

mobasherul: 3 servers and 2 voters will need 3 votes to become active that's why 2 voters are enough. It is because of this we don't even need the votes of the new passive as well as the killed active since voters will serve the purpose.

mathieucarbou (approved):

PR #936, opened 2021-02-12
New Angela version: using "angela.java.resolver=user" now

mathieucarbou:


dynamic-config/testing/support/src/main/java/org/terracotta/dynamic_config/test_support/DynamicConfigIT.java
```diff
@@ -300,9 +300,7 @@ protected ConfigurationContext createConfigurationContext(int stripes, int nodes
         .tsa(tsa -> tsa
             .clusterName(CLUSTER_NAME)
             .license(getLicenceUrl() == null ? null : new License(getLicenceUrl()))
-            .terracottaCommandLineEnvironment(TerracottaCommandLineEnvironment.DEFAULT
-                .withJavaOpts("-Xms32m -Xmx256m")
-                .withJavaHome(System.getProperty("java.home")))
```
mathieucarbou, line 305: we do not need to "customize" some process launching anymore because by default, the java.home taken will be the running one when using `<angela.java.resolver>user</angela.java.resolver>`

dynamic-config/testing/system-tests/pom.xml
```diff
@@ -145,8 +145,7 @@
             <angela.tsa.fullLogging>true</angela.tsa.fullLogging>
             <angela.distribution>${project.version}</angela.distribution>
             <angela.kitInstallationDir>${project.build.directory}/angela/platform-kit-${project.version}</angela.kitInstallationDir>
-            <angela.java.vendor/> <!--empty to allow any vendor-->
-            <angela.java.version>${java.test.version}</angela.java.version>
+            <angela.java.resolver>user</angela.java.resolver>
```
mathieucarbou, line 148: angela does not need anymore to lookup the JVM path into the maven toolchain: it will reuse the java.home system property passed by maven (and toolchain plugin)

pom.xml
```diff
@@ -39,7 +39,7 @@
     <terracotta-core.version>5.7.1-pre9</terracotta-core.version>
     <tripwire.version>1.0.0</tripwire.version>
     <galvan.version>1.5.0</galvan.version>
-    <angela.version>3.1.13</angela.version>
+    <angela.version>3.1.14</angela.version>
```
mathieucarbou, line 42: builds will fail because release is in progress

akomakom (approved):

PR #939, opened 2021-02-17
Several fixes for unset command and added a couple of system test for activated and in-configuration mode

mathieucarbou:


mathieucarbou:
Angela `3.1.16` is releasing...

mathieucarbou:
@mobasherul @tmesic99 @chrisdennis : this PR contains several things: 

1. fixes the `unset` command: some use cases were not working correctly
2. Added some new system tests for `unset`
2. Method renames to cleanup test code
3. Wrapped all config tool invocation with: `assertThat(..., is(successfull()))`
4. Updated the assertions for config tool according to the latest angela changes:

**Angela changes:**

I found an issue in Angela when trying to test my fixes: Angela was forcing all command invocations to return an exit code 0. If not, it was throwing with the process output within the exception message. This was problematic because we were not able to re-use the normal matchers when doing assertions on the config tool result: instead we had to wrap the config tool invocation within a closure and catch an exception.

So I have removed this Angela limitation. 

Now, all config tool invocation won't "throw" and the result object will contain the exist code and output, whether the command failed or not, which allows the test to completely control the assertions, and with the same result object.

mobasherul (approved):

PR #949, opened 2021-02-23
Fix: Changing a tc-property only requires a node restart

mathieucarbou:


tmesic99 (approved):

mathieucarbou:
will integrate the code in #948

PR #957, opened 2021-03-02
set JAVA_HOME from java.home system property

mathieucarbou:


galvan-platform-support/src/main/java/org/terracotta/testing/config/ConfigFileStartupBuilder.java
```diff
@@ -116,6 +116,7 @@ private void executeCommand(List<String> command, Path workingDir) {
       AnyProcessBuilder<? extends AnyProcess> builder = AnyProcess.newBuilder()
           .command(command.toArray(new String[0]))
           .workingDir(workingDir.toFile())
+          .env("JAVA_HOME", System.getProperty("java.home"))
```
mobasherul, line 119: Is it enough ? I can see that in galvan while spawning process for running server we use this - 
private String getJavaHome() {
    String javaHome = System.getenv("JAVA_HOME");
    if (null == javaHome) {
      javaHome = System.getProperty("java.home");
      Assert.assertNotNull(javaHome);
      this.harnessLogger.output("WARNING:  JAVA_HOME not set!  Defaulting to \"" + javaHome + "\"");
    }

    return javaHome;
  }

mathieucarbou: @mobasherul : this is what I told @myronkscott yesterday.
I have fixed what we can in tc-platform but galvan will need to be fixed.

mobasherul: ok got it.

mobasherul (approved):

PR #962, opened 2021-03-05
Inline config-tool calls in DC system tests

mathieucarbou:
**OutputService refactoring:**

- cleanup hierarchy
- introduced `OutputService#close()`
- introduced `OutputService#then(OutputService)`
- introduced `ConsoleOutputService`
- introduced `InMemoryOutputService`

**Fix:** 
- unclosed file output stream when using `FileOutputService`

**ConfigTool and ConfigConverterTool now support the injection of an OutputService**

**DC system tests are now using inline config-tool invocations**


mathieucarbou:
All tests passed locally 👍

mathieucarbou:
FYI, the build is between 6-12min faster depending on the OS and Java version. 
The build nows lasts 1h20 - 1h45 depending on the OS and Java version.

myronkscott (approved):

PR #970, opened 2021-03-18
[WIP] Data Dir changes

mobasherul:


dynamic-config/model/src/main/java/org/terracotta/dynamic_config/api/model/Setting.java
```diff
@@ -407,9 +407,9 @@
       asList(
           when(CONFIGURING).allow(IMPORT).atLevel(NODE),
           when(CONFIGURING).allow(GET, SET, UNSET).atAnyLevels(),
-          when(ACTIVATED).allow(GET).atAnyLevels()
+          when(ACTIVATED).allow(GET, SET).atAnyLevels()
       ),
-      of(PRESENCE),
+      of(NODE_RESTART, PRESENCE),
```
tmesic99, line 412: Add a task to our Documentation story (TDB-4806) to update the Config-Tool.adoc to reflect these updates.

mathieucarbou: Wait... @mobasherul : `NODE_RESTART ` should not be added. When a requirement is added there, it is global to any change. For data-dir, it is OK to add a data-dir without restart. Only updating requires a restart (and this has to be handled with the new veto method). The code should be similar to backup-dir, which is a setting we can set and update, but not unset.

data-root-resource/src/main/java/org/terracotta/config/data_roots/DataRootsDynamicConfigExtension.java
```diff
@@ -42,10 +51,55 @@ public void configure(Registrar registrar, PlatformConfiguration platformConfigu
...
+    });
     DataDirsConfigImpl dataDirectoriesConfig = new DataDirsConfigImpl(parameterSubstitutor, pathResolver, nodeMetadataDir, dataDirs);
     configChangeHandlerManager.set(Setting.DATA_DIRS, new DataDirConfigChangeHandler(dataDirectoriesConfig, parameterSubstitutor, pathResolver));
 
     registrar.registerExtendedConfiguration(dataDirectoriesConfig);
   }
+
+  private void copyIfMarkFilePresent(Path newPath, IParameterSubstitutor parameterSubstitutor, PathResolver pathResolver) {
+    Path substitutedPath = parameterSubstitutor.substitute(pathResolver.resolve(newPath));
+    if (Files.isReadable(substitutedPath.resolve(markFile))) {
```
mathieucarbou, line 67: if we have a marker file but it is not readable, we have to fail the server startup because if we start without the data it will mess up things ;-)

data-root-resource/src/main/java/org/terracotta/config/data_roots/DataRootsDynamicConfigExtension.java
```diff
@@ -42,10 +51,55 @@ public void configure(Registrar registrar, PlatformConfiguration platformConfigu
...
+      try {
+        String line = Files.readAllLines(substitutedPath.resolve(markFile)).get(0);
```
mathieucarbou, line 69: the logic is wrong: there can be other files created between the time we have created the marker and the time the user restarts. Also, the user may have messed up things. So to prevent any problem, here is the logic we should apply:

* when the marker file is created, put inside the old location + the list of all files inside the old data dir (including sub folders)
* when starting:
  * first load the marker file, load the old location and verify that the old location contains __at least__ all the listed files in the marker file
  * if no, or if the old location does not exist => we fail the startup: it means we miss a file that has been deleted (by the user ?) and this is wrong
  * if the old location exist AND all the listed files in the marker file are there
    * we copy all the files from old to the new location: __we have to replace existing files__ because if the server crashes and restarts, we need to be able to resume. __Note__: we do not copy the list in the marker file: we copy the exact content from the old location (which might contain more than the marker file, __but not less__)
    * once the whole copy is done, we can remove the marker file (first) : if the server crashes before the removal, the operation will happen again and will work, if the server crashes after the removal of this file, that is ok we have everything we need
    * then we can delete the old location completely.

dynamic-config/server/services/src/main/java/org/terracotta/dynamic_config/server/service/handler/MetaDataDirConfigChangeHandler.java
```diff
@@ -0,0 +1,83 @@
...
+  private final IParameterSubstitutor parameterSubstitutor;
+  private final PathResolver pathResolver;
+
+  public MetaDataDirConfigChangeHandler(IParameterSubstitutor parameterSubstitutor, PathResolver pathResolver) {
+    this.parameterSubstitutor = parameterSubstitutor;
+    this.pathResolver = pathResolver;
+  }
+
+  @Override
+  public void validate(NodeContext baseConfig, Configuration change) throws InvalidConfigChangeException {
```
mathieucarbou, line 41: same comment for data dirs: 
- we need to check the rawpath, path, then substituted real path
- we need to write in the marker file the old location plus the list the whole folder content
- etc.

perhaps it could be worth factoring out some  utility class that would be responsible for the move, something like:

`new MoveOperation(Path destination).prepare(Path source)`
`new MoveOperation(Path destination).move()`

- MoveOperation is constructed with a destination folder
- it can prepare the marker file by specifying a source location
- it can run the move operation by checking the marker file which contains the source and checklist

mathieucarbou (changes requested):

dynamic-config/server/server-api/src/main/java/org/terracotta/dynamic_config/server/api/MoveOperation.java
```diff
@@ -0,0 +1,105 @@
...
+  private static final Logger LOGGER = LoggerFactory.getLogger(MoveOperation.class);
+
+  public MoveOperation(Path destination) {
+    this.destination = destination;
+  }
+
+  public void prepare(Path sourcePath) throws IOException {
+    Path mark = Files.createFile(destination.resolve(markFile));
+    List<String> list = new ArrayList<>();
+    Files.walk(sourcePath).forEach(path -> list.add(path.toString()));
```
mathieucarbou, line 46: file streams need to be closed:
```
try (Stream<Path> paths = Files.walk(...)) {
...
}
```

dynamic-config/server/server-api/src/main/java/org/terracotta/dynamic_config/server/api/MoveOperation.java
```diff
@@ -0,0 +1,105 @@
...
+  private final Path destination;
+  protected final String markFile = "mark.txt";
+  private static final Logger LOGGER = LoggerFactory.getLogger(MoveOperation.class);
+
+  public MoveOperation(Path destination) {
+    this.destination = destination;
+  }
+
+  public void prepare(Path sourcePath) throws IOException {
+    Path mark = Files.createFile(destination.resolve(markFile));
```
mathieucarbou, line 44: we shouldn't fail if file exists. If file exists, we should truncate and re-use it. we need to make sure the user can "replay" a change again if it fails, partially prepared, etc.

dynamic-config/server/server-api/src/main/java/org/terracotta/dynamic_config/server/api/MoveOperation.java
```diff
@@ -0,0 +1,105 @@
...
+
+  private final Path destination;
+  protected final String markFile = "mark.txt";
+  private static final Logger LOGGER = LoggerFactory.getLogger(MoveOperation.class);
+
+  public MoveOperation(Path destination) {
+    this.destination = destination;
+  }
+
+  public void prepare(Path sourcePath) throws IOException {
```
mathieucarbou, line 43: I think we should do something like: `sourcePath = sourcePath.toRealPath()` right ?
Because pathResolver is just rebasing a path based on a working directory if the path is relative.
So I am wondering if the output of walk could produce in some cases a file with a list of path inside that are not absolute... This could happen if the base directory passed to the path resolver is not absolute, like: `./kit/server`.

dynamic-config/server/server-api/src/main/java/org/terracotta/dynamic_config/server/api/MoveOperation.java
```diff
@@ -0,0 +1,105 @@
...
+          try {
+            org.terracotta.utilities.io.Files.deleteTree(sourcePath);
+          } catch (IOException e) {
+            LOGGER.warn("Unable to delete obsolete data-dir path: {}", sourcePath);
+          }
+        } catch (IOException e) {
+          throw new UncheckedIOException(e);
+        }
+      } else {
+        throw new RuntimeException("Server can't start up because of permission issues which can lead to data loss");
```
mathieucarbou, line 100: we should add more explanation: 
- this is when moving a data directory
- which data directory ?

dynamic-config/server/server-api/src/main/java/org/terracotta/dynamic_config/server/api/MoveOperation.java
```diff
@@ -0,0 +1,105 @@
...
+  }
+
+  public void move() {
+    Path mark = destination.resolve(markFile);
+    if (mark.toFile().exists()) {
+      if (Files.isReadable(mark)) {
+        try {
+          Files.readAllLines(mark).forEach(path -> {
+            if (!Files.isReadable(Paths.get(path))) {
+              throw new UncheckedIOException(new IOException(path + " must be present"));
```
mathieucarbou, line 57: add better error message like: `Moving data directory from: {} to: {} failed because file: {} is not readable`

dynamic-config/server/server-api/src/main/java/org/terracotta/dynamic_config/server/api/MoveOperation.java
```diff
@@ -0,0 +1,105 @@
...
+          Files.walkFileTree(sourcePath, new SimpleFileVisitor<Path>() {
+            @Override
+            public FileVisitResult visitFile(Path file,
+                                             BasicFileAttributes attributes) {
+
+              try {
+                Path targetFile = destination.resolve(sourcePath.relativize(file));
+                org.terracotta.utilities.io.Files.copy(file, targetFile, StandardCopyOption.REPLACE_EXISTING);
+              } catch (IOException ex) {
+                throw new UncheckedIOException(ex);
```
mathieucarbou, line 71: add better error message like: `Moving data directory from: {} to: {} failed because file: {} cannot be copied. Error: {}`

dynamic-config/server/server-api/src/main/java/org/terracotta/dynamic_config/server/api/MoveOperation.java
```diff
@@ -0,0 +1,105 @@
...
+    Path mark = destination.resolve(markFile);
+    if (mark.toFile().exists()) {
+      if (Files.isReadable(mark)) {
+        try {
+          Files.readAllLines(mark).forEach(path -> {
+            if (!Files.isReadable(Paths.get(path))) {
+              throw new UncheckedIOException(new IOException(path + " must be present"));
+            }
+          });
+          String line = Files.readAllLines(mark).get(0);
```
mathieucarbou, line 60: just do only one `Files.readAllLines(mark)` and keep this list

mathieucarbou (changes requested):

dynamic-config/server/server-api/src/main/java/org/terracotta/dynamic_config/server/api/MoveOperation.java
mobasherul: Do we want to add info about specific data-drectory as this could happen due to metadata-dir as well

mobasherul: How about -> Server cannot startup because movement for one of the " +
            "existing data-dir or metadata-dir to new destination: " + destination +
            "cannot be completed because of permission issues");

dynamic-config/server/server-api/src/main/java/org/terracotta/dynamic_config/server/api/MoveOperation.java
mobasherul: I am not sure if it can cause problem because whatever be the basepath for resolver, both the existing datadir path and newdatadir path are resolved from that same base path.

mathieucarbou: yes you are right... forget about my comment ;-)

dynamic-config/server/server-api/src/main/java/org/terracotta/dynamic_config/server/api/MoveOperation.java
mathieucarbou: perfect!

dynamic-config/server/server-api/src/main/java/org/terracotta/dynamic_config/server/api/MoveOperation.java
mathieucarbou: yes right! forget my comment

mathieucarbou (approved):

PR #978, opened 2021-03-24
Remove usage of Stream.parallel() that could exhaust fork join pool in case of a JVM running tests in parallel and if one of the test times out

mathieucarbou:


dynamic-config/model/src/main/java/org/terracotta/dynamic_config/api/service/ClusterValidator.java
```diff
@@ -110,7 +110,6 @@ private void validateNames() {
           // invalid chars
           {
             Character invalid = IntStream.range(0, name.length())
-                .parallel()
```
mathieucarbou, line 113: was using `.parallell()` here but this validation is not done frequently enough that it warrants to be sped up

common/test-utilities/src/main/java/org/terracotta/testing/ThreadDump.java
```diff
@@ -96,20 +96,16 @@ public void writeTo(Path out) {
...
-  public static void dumpAll(Path outputDir, boolean parallel, Duration timeout) {
+  public static void dumpAll(Path outputDir, Duration timeout) {
     try {
       Files.createDirectories(outputDir);
     } catch (IOException e) {
       throw new UncheckedIOException(e);
     }
-    Stream<ThreadDump> stream = dumpAll(timeout);
-    if (parallel) {
-      stream = stream.parallel();
```
mathieucarbou, line 110: jstack processes were launched (and waited for) in the fork join pool - I didn't use an executor because before 1 jvm was not processing tests in parallel, and this code was executed at the end of each test in case of timeout, but whatever, this is common code and this .patallel usage is bad.

myronkscott (approved):

PR #981, opened 2021-03-25
Jmap

mathieucarbou:
- Refactored the thread dump code
- Added memory dump feature (**not enabled by default**)
- Removed the unnecessary extension of timeout rule, to use a test watcher instead to capture dumps when a test fails, regardless of a timeout or not. Test watcher are more flexible because we can access the test information + error.


common/test-utilities/src/main/java/org/terracotta/testing/JavaTool.java
```diff
@@ -0,0 +1,164 @@
...
+import static java.util.Objects.requireNonNull;
+import static org.terracotta.testing.JavaBinary.JMAP;
+import static org.terracotta.testing.JavaBinary.JPS;
+import static org.terracotta.testing.JavaBinary.JSTACK;
+import static org.terracotta.testing.JavaBinary.exec;
+
+/**
+ * @author Mathieu Carbou
+ */
+public class JavaTool {
```
mathieucarbou, line 41: Renamed from `ThreadDump`

dynamic-config/testing/support/src/main/java/org/terracotta/dynamic_config/test_support/DynamicConfigIT.java
```diff
@@ -254,6 +254,21 @@ protected void succeeded(Description description) {
...
+              // thread dumps
+              Path threadDumpOutput = target.resolve("thread-dumps").resolve(description.toString());
+              LOGGER.info("Taking thread dumps after timeout of test: {} into: {}", description, threadDumpOutput);
+              JavaTool.threadDumps(threadDumpOutput, Duration.ofSeconds(15));
+            }
+            {
+              // memory dumps on test failures
+              //Path memoryDumpOutput = target.resolve("memory-dumps").resolve(description.toString());
+              //LOGGER.info("Taking memory dumps after timeout of test: {} into: {}", description, memoryDumpOutput);
+              //JavaTool.threadDumps(memoryDumpOutput, Duration.ofSeconds(20));
```
mathieucarbou, line 270: to enable memory dumps

mathieucarbou, line 270: @akomakom : is it possible to update the pattern of what gets collected in the zip file to be:
```
**/thread-dumps/**
**/memory-dumps/**
```
thanks!

tmesic99 (approved):

PR #987, opened 2021-04-19
Fixing Data-Dir and MetaData-Dir update.

mobasherul:


mobasherul:
This  is required for cases where suppose two nodes are running on same machine do their data-dir have paths like-
<home>/user-data/main/node1 and other one has <home>/user-data/main/node-2.
And we want to change the data-dir path for only one of the node.

dynamic-config/server/server-api/src/main/java/org/terracotta/dynamic_config/server/api/MoveOperation.java
```diff
@@ -78,7 +78,7 @@ public FileVisitResult visitFile(Path file,
                                              BasicFileAttributes attributes) {
 
               try {
-                Path targetFile = destination.resolve(sourcePath.relativize(file));
+                Path targetFile = destination.resolve(sourcePath.getParent().relativize(file));
```
chrisdennis, line 81: Isn't this encoding a detail of how a dependent module uses this method. I.e. you're baking details of how `data-root-resource` treats node directories in to an api module that doesn't know anything about this. Could we inject this knowledge somehow? In fact shouldn't the whole `MoveOperation` be operating against a source and target that have had the node-name convention pre-applied, perhaps by feeding them the result of a `getDataDirectoriesForServer(...)` call?

mathieucarbou: @chrisdennis : 
- the problem is that the the move operation needs to happen eagerly at server startup before any plugin / service provide gets initialized.

But I _think_ that the move operation could be constructed after the call to `DataDirsConfigImpl dataDirectoriesConfig = new DataDirsConfigImpl(...)`, and this way we could do handle the data dirs + metadata dir the same way in 2 lines:

```
    DataDirs dirs = dataDirectoriesConfig.getDataDirectoriesForServer(platformConfiguration); 
    Stream.concat(dirs.getDataDirectoryNames().stream().map(Optional::of), Stream.of(dirs.getPlatformDataDirectoryIdentifier()))
        .filter(Optional::isPresent)
        .map(name -> dirs.getDataDirectory(name.get()))
        .forEach(path -> new MoveOperation(path).move());
```

mobasherul: I have done this before calling getDataDirectoriesForServer() as it creates internal lock files which can cause problem during movement to new data-dir.

tmesic99 (approved):

PR #989, opened 2021-04-20
sync management server model with DC

GaryWKeim:
Fixes a bug with DataDir stats.
Fast access to any running node's diagnostic service.. for dynamic config monitoring service.
M&M TopologyService enhancements/fixes/optimizations (notably black-listing certain clients/entities).

### DIFF
--- a/diagnostic/client/src/main/java/org/terracotta/diagnostic/client/connection/ConcurrentDiagnosticServiceProvider.java
+++ b/diagnostic/client/src/main/java/org/terracotta/diagnostic/client/connection/ConcurrentDiagnosticServiceProvider.java
@@ -50,13 +50,22 @@ public class ConcurrentDiagnosticServiceProvider<K> implements MultiDiagnosticSe
 
   @Override
   public DiagnosticServices<K> fetchDiagnosticServices(Map<K, InetSocketAddress> addresses) {
+    return _fetchOnlineDiagnosticService(addresses, false);
+  }
+
+  @Override
+  public DiagnosticServices<K> fetchAnyOnlineDiagnosticService(Map<K, InetSocketAddress> addresses) {
+    return _fetchOnlineDiagnosticService(addresses, true);
+  }
+
+  private DiagnosticServices<K> _fetchOnlineDiagnosticService(Map<K, InetSocketAddress> addresses, boolean firstAvailable) {
     if (addresses.isEmpty()) {
       return new DiagnosticServices<>(emptyMap(), emptyMap());
     }
 
     ExecutorService executor = Executors.newFixedThreadPool(
-        concurrencySizing.getThreadCount(addresses.size()),
-        r -> new Thread(r, "diagnostics-connect"));
+      concurrencySizing.getThreadCount(addresses.size()),
+      r -> new Thread(r, "diagnostics-connect"));
 
     try {
       CompletionService<Tuple3<K, DiagnosticService, DiagnosticServiceProviderException>> completionService = new ExecutorCompletionService<>(executor);
@@ -86,7 +95,11 @@ public class ConcurrentDiagnosticServiceProvider<K> implements MultiDiagnosticSe
           Tuple3<K, DiagnosticService, DiagnosticServiceProviderException> tuple = completed.get();
           if (tuple.t3 == null) {
             online.put(tuple.t1, tuple.t2);
-          } else {
+            if (firstAvailable) {
+              executor.shutdownNow();
+              break;
+            }
+          } else if (!firstAvailable) {
             offline.put(tuple.t1, tuple.t3);
           }
         }
--- a/diagnostic/client/src/main/java/org/terracotta/diagnostic/client/connection/MultiDiagnosticServiceProvider.java
+++ b/diagnostic/client/src/main/java/org/terracotta/diagnostic/client/connection/MultiDiagnosticServiceProvider.java
@@ -56,11 +56,20 @@ public interface MultiDiagnosticServiceProvider<K> {
   }
 
   /**
-   * Concurrently fetch the diagnostic service of all the provides nodes.
+   * Concurrently fetch the diagnostic service of all the provided nodes.
    * The method will not fail, except if some connection timeout is reached.
    * If some nodes are offline, they will be reported in {@link DiagnosticServices#getOfflineEndpoints()}.
    * <p>
    * The returned {@link DiagnosticServices} will have a list of online nodes and a list of offline nodes.
    */
   DiagnosticServices<K> fetchDiagnosticServices(Map<K, InetSocketAddress> addresses);
+
+  /**
+   * Concurrently fetch any single online diagnostic service of all the provided nodes.
+   * The method will not fail, except if some connection timeout is reached.
+   * Offline nodes are ignored.
+   * <p>
+   * If successful, the returned {@link DiagnosticServices} will have a single online node.
+   */
+  DiagnosticServices<K> fetchAnyOnlineDiagnosticService(Map<K, InetSocketAddress> addresses);
 }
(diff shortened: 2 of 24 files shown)


diagnostic/client/src/main/java/org/terracotta/diagnostic/client/connection/ConcurrentDiagnosticServiceProvider.java
```diff
@@ -106,6 +106,64 @@ public ConcurrentDiagnosticServiceProvider(DiagnosticServiceProvider diagnosticS
...
+
+      try {
+        // capture the task output
+        int count = addresses.size();
+        while (count-- > 0) {
+          // we do not need to handle any timeout here during a take or get because they are handled in the submitted tasks
+          Future<Tuple3<K, DiagnosticService, DiagnosticServiceProviderException>> completed = completionService.take();
+          Tuple3<K, DiagnosticService, DiagnosticServiceProviderException> tuple = completed.get();
+          if (tuple.t3 == null) {
+            online.put(tuple.t1, tuple.t2);
```
tmesic99, line 146: From what I can see, the only difference between this new method and the existing fetchDiagnosticServices() are lines 147 and 148.  If this is in fact the only difference, how about making this method private (change its name) and add a boolean parameter ('useAny') to its signature.  Then you can call this private method from the existing fetchDiagnosticServices() passing in 'false'.   And create a new public method  fetchAnyOnlineDiagnosticService() that calls the private method with 'true' - and wrap lines 147-148 accordingly.

mathieucarbou: Agree, we could have a utility method that has a boolean, whether we want to wait for all the results or if we only want to wait for the first available result.

diagnostic/client/src/main/java/org/terracotta/diagnostic/client/connection/MultiDiagnosticServiceProvider.java
```diff
@@ -63,4 +63,13 @@
...
   DiagnosticServices<K> fetchDiagnosticServices(Map<K, InetSocketAddress> addresses);
+
+  /**
+   * Concurrently fetch any single online diagnostic service of all the provides nodes.
+   * The method will not fail, except if some connection timeout is reached.
+   * Offline nodes are ignored.
+   * <p>
+   * If successful, the returned {@link DiagnosticServices} will have a single online node.
+   */
+  DiagnosticServices<K> fetchAnyOnlineDiagnosticService(Map<K, InetSocketAddress> addresses);
```
mathieucarbou, line 74: This API would be highly useful for the cluster tool too 👍

diagnostic/model/src/main/java/org/terracotta/diagnostic/model/LogicalServerState.java
```diff
@@ -101,11 +104,7 @@ public static LogicalServerState from(String state, boolean reconnectWindow, boo
       return LogicalServerState.PASSIVE_SUSPENDED;
     } else if (parsedLogicalServerState == LogicalServerState.STARTING && isBlocked) {
       return LogicalServerState.START_SUSPENDED;
-    } else if (parsedLogicalServerState == LogicalServerState.DIAGNOSTIC) {
-      return LogicalServerState.DIAGNOSTIC;
-    } else {
-      return parsedLogicalServerState;
-    }
+    } else return parsedLogicalServerState;
```
mathieucarbou, line 107: thanks for the fix ;-)

management/monitoring-service/src/main/java/org/terracotta/management/service/monitoring/TopologyService.java
```diff
@@ -568,4 +598,23 @@ private static ClientIdentifier toClientIdentifier(PlatformConnectedClient conne
         connection.uuid);
   }
 
+  private final static Set<String> CLIENT_NAME_BLACKLIST = new HashSet<>(Arrays.asList("", "CONFIG-TOOL"));
```
mathieucarbou, line 601: `CLUSTER-TOOL` could be added. There is also another tool that has been added recently to create the permanent entities I think ? This one should have a name too

mathieucarbou (approved):
just some comments :-) good to go for me!

management/monitoring-service/src/main/java/org/terracotta/management/service/monitoring/TopologyService.java
```diff
@@ -143,10 +152,18 @@ public synchronized void serverDidJoinStripe(PlatformServer platformServer) {
...
         .computeUpTime();
 
     if (stripe.addServer(server)) {
-      firingService.fireNotification(new ContextualNotification(server.getContext(), SERVER_JOINED.name()));
+      Map<String, String> attrs = new HashMap<>();
+      attrs.put("startTime", String.valueOf(server.getStartTime()));
+      attrs.put("activateTime", String.valueOf(server.getActivateTime()));
+      attrs.put("version", server.getVersion());
+      attrs.put("buildId", server.getBuildId());
+      attrs.put("state", "STARTING");
```
chrisdennis, line 164: Shouldn't this be `server.getState().toString()`?

management/monitoring-service/src/main/java/org/terracotta/management/service/monitoring/TopologyService.java
```diff
@@ -160,7 +177,14 @@ public synchronized void serverDidLeaveStripe(PlatformServer platformServer) {
 
       serverEntities.remove(platformServer.getServerName());
 
-      firingService.fireNotification(new ContextualNotification(context, SERVER_LEFT.name()));
+      Map<String, String> attrs = new HashMap<>();
+      attrs.put("startTime", "0");
+      attrs.put("activateTime", "0");
+      attrs.put("version", null);
+      attrs.put("buildId", null);
+      attrs.put("state", "UNREACHABLE");
```
chrisdennis, line 185: Can this be `server.getState().toString()` I would hope/assume that status would have been updated, but it's not immediately obvious that is true.

management/monitoring-service/src/main/java/org/terracotta/management/service/monitoring/TopologyService.java
```diff
@@ -123,10 +126,16 @@ public synchronized void serverDidBecomeActive(PlatformServer self) {
 
       topologyEventListeners.forEach(listener -> listener.onBecomeActive(platformConfiguration.getServerName()));
 
-      firingService.fireNotification(new ContextualNotification(server.getContext(), SERVER_JOINED.name()));
+      Map<String, String> attrs = new HashMap<>();
+      attrs.put("startTime", String.valueOf(server.getStartTime()));
+      attrs.put("activateTime", String.valueOf(server.getActivateTime()));
+      attrs.put("version", server.getVersion());
+      attrs.put("buildId", server.getBuildId());
+      attrs.put("state", "ACTIVE");
```
chrisdennis, line 134: `server.getState().toString()` again.

management/cluster-topology/src/main/java/org/terracotta/management/model/cluster/Server.java
```diff
@@ -18,6 +18,7 @@
 import org.terracotta.management.model.context.Context;
 
 import java.time.Clock;
+import java.util.Comparator;
```
chrisdennis, line 21: Unused.

diagnostic/client/src/main/java/org/terracotta/diagnostic/client/connection/ConcurrentDiagnosticServiceProvider.java
chrisdennis: I think this could be done using a stream of `CompletionStage` instances with the executor shutdown triggered via an on-close. The "first online" is then a reduction using an `applyEither(...)` and the other is a grouping collection.

chrisdennis (changes requested):

diagnostic/client/src/main/java/org/terracotta/diagnostic/client/connection/ConcurrentDiagnosticServiceProvider.java
chrisdennis: Sorry, just to be clear... I wasn't serious about the reduce thing... (also it turns out applyEither has an annoying contract that makes this yet more complex).

chrisdennis (approved):

PR #990, opened 2021-04-21
Fixing Data-Dir and MetaData-Dir update.

mobasherul:


data-root-resource/src/main/java/org/terracotta/config/data_roots/DataDirConfigChangeHandler.java
```diff
@@ -104,6 +105,10 @@ public void validate(NodeContext baseConfig, Configuration changes) throws Inval
           throw new InvalidConfigChangeException("Directory: " + substitutedDataDirPath + " doesn't have write permissions" +
               " for the user: " + parameterSubstitutor.substitute("%n") + " running the server process");
         }
+
+        if (!isEmpty(substitutedDataDirPath)) {
```
mobasherul, line 109: This is required now if somebody is running all servers in same host and do something like set data-dir=main:<path> then it should fail otherwise subsequent nodes will overwrite the mark files we are writing in new path.

mathieucarbou: @mobasherul @tmesic99 @chrisdennis : regarding any validation / changes in the validate method: pay attention that a nomad change needs to be "replay-able" in case of partial prepare for example. If we require the `substitutedDataDirPath` to be empty, which is logical, how to handle the case where we have a partial prepare that is caused by a node crash at the time we do the `prepare()` ? 
In such case, the destination folder is created, the marker file is being written, but the node crashes, and the nomad transaction won't be repairable because this `validate()` method will always fail.

mathieucarbou (changes requested):

data-root-resource/src/main/java/org/terracotta/config/data_roots/Utils.java
```diff
@@ -0,0 +1,40 @@
...
+import org.terracotta.dynamic_config.server.api.InvalidConfigChangeException;
+
+import java.io.IOException;
+import java.nio.file.DirectoryStream;
+import java.nio.file.Files;
+import java.nio.file.Path;
+
+public class Utils {
+  public static boolean overLaps(Path existing, Path newPath) {
+    return existing.startsWith(newPath) || newPath.startsWith(existing);
```
chrisdennis, line 27: I think both of these needs to be converted to `toRealPath()` before this method will work correctly? In fact you may even need to be more exhaustive by actually resolving paths against the real filesystem otherwise you might get defeated by things like symlinks.

data-root-resource/src/main/java/org/terracotta/config/data_roots/Utils.java
```diff
@@ -0,0 +1,40 @@
...
+  }
+
+  public static boolean isEmpty(Path path) throws InvalidConfigChangeException {
+    if (Files.isDirectory(path)) {
+      try (DirectoryStream<Path> directory = Files.newDirectoryStream(path)) {
+        return !directory.iterator().hasNext();
```
chrisdennis, line 33:
```java
return !directory.findAny().isPresent()
```
seems more Stream-like.

data-root-resource/src/main/java/org/terracotta/config/data_roots/Utils.java
mobasherul: Existing path resolver would already have resolved the paths according to base dir before passing the paths here but I am not sure whether that will work with symlinks.

chrisdennis:
```java
  public Path resolve(Path path) {
    if (path == null) {
      return null;
    }
    Path transformed = transform.apply(path);
    if (transformed.isAbsolute()) {
      return path; // keep original path and placeholders
    }
    return baseDir.resolve(path); // keep original path and placeholders
  }
```
Unless `baseDir` is absolute this doesn't necessarily return an absolute path.

Also... and absolute and a real path aren't the same thing (especially on case insensitive filesystems).

data-root-resource/src/main/java/org/terracotta/config/data_roots/Utils.java
mobasherul: Its not a stream.

mathieucarbou (approved):

PR #995, opened 2021-04-29
updating sample config file to new format

tmesic99:


kit/src/main/kit/server/conf/tc-cluster.cfg
```diff
@@ -65,51 +65,50 @@ failover-priority=availability
 # A comma-separated list of offheap (in-memory) resource pools for data storage
 offheap-resources=main:512MB
 
-### Define stripes, and nodes within the stripes
+# Define stripes in the cluster
+stripe-names=tc-stripe1
+# If specifying multiple stripes, use comma-separated-list
+#stripe-names=tc-stripe1,tc-stripe2
```
chrisdennis, line 71: This sample config is only used by Ehcache now... probably doesn't need to (or maybe shouldn't) mention multiple stripes.

chrisdennis (dismissed):

kit/src/main/kit/server/conf/tc-cluster.cfg
mathieucarbou: Ehcache uses his own one: https://github.com/ehcache/ehcache3/tree/master/clustered/clustered-dist/src/assemble/server/conf

This one is for the vanilla kit.

mathieucarbou (approved):

kit/src/main/kit/server/conf/tc-cluster.cfg
chrisdennis: Okay... but this vanilla kit is I think only used by Ehcache (I assume this property file gets over-written?).

mathieucarbou: > Okay... but this vanilla kit is I think only used by Ehcache (I assume this property file gets over-written?).

I don't know. What I can tell is that this vanilla kit was introduced so that we have a kit in tc-platform to test with, and also because there has been a long period of time where DC was in development and not integrated neither in Ehcache or EE. So we were taking this vanilla kit for our local testing.

Technically speaking we do not need. But from a UX perspective for devs, if someone uses the platform kit for some local testing, this is a nice to have. And FYI we still currently use this vanilla kit sometimes when we are not able to use a plain EE kit because of incompatibilities.

PR #996, opened 2021-05-04
Rework the CLI/UX regarding the different options for supported config file formats

mathieucarbou:
- Also updated the command helps
- Also fixed a bug related to unsupported extensions with uppercase characters


dynamic-config/cli/config-tool/src/main/java/org/terracotta/dynamic_config/cli/config_tool/parsing/GetCommand.java
```diff
@@ -18,22 +18,22 @@
...
+import org.terracotta.dynamic_config.cli.converter.FormatConverter;
 import org.terracotta.dynamic_config.cli.converter.InetSocketAddressConverter;
 import org.terracotta.dynamic_config.cli.converter.MultiConfigCommaSplitter;
 
 import java.net.InetSocketAddress;
 import java.util.List;
 
 @Parameters(commandDescription = "Read configuration properties")
-@Usage("-connect-to <hostname[:port]> [-runtime] [-outputformat name|index] -setting <[namespace:]setting> -setting <[namespace:]setting> ...")
+@Usage("-connect-to <hostname[:port]> [-runtime] [-outputformat cfg|properties] -setting <[namespace:]setting> -setting <[namespace:]setting> ...")
```
tmesic99, line 36: The format types 'name' and 'index' was meant to align with the two different namespace formats: 'name-based' and 'index-based'.  The 'cfg' and 'properties' types are meant to align with supported exported file extension names.  While the code change will work, using the export file extension names as the type identifiers for the 'get' command seems a bit strange and not as intuitive to the user.  But hey, once they're documented, the user can understand their exact usage anyway.

tmesic99, line 36: If we move from name/index to cfg/properties for the get command, then there are many test cases which need to be fixed.  Just search for all usages of '-oututformat' and they should reveal themselves.  Same on the enterprise side..

mathieucarbou: Let's discuss that: it is fine if we leave it like before, it was an attempt to simplify the concepts of output format displayed to the user because there is actually only 2: the properties format (using indexes) and the new config format.

If we look at what was there before, the output of the get was matching the lines in the properties file format. They were on the properties file format then.

I only saw to places to update on the doc if we decide to go with that but I can look again.

tmesic99 (approved):

PR #997, opened 2021-05-06
Notify that a client/entity registry is AVAILABLE after it has changed in any way.

GaryWKeim:
This defect shows itself as the no statistics appearing in the Server Resource Usage panel for newly attached stripes.

### DIFF
--- a/management/monitoring-service/src/main/java/org/terracotta/management/service/monitoring/TopologyService.java
+++ b/management/monitoring-service/src/main/java/org/terracotta/management/service/monitoring/TopologyService.java
@@ -437,9 +437,8 @@ class TopologyService implements PlatformListener {
     LOGGER.trace("[{}] willSetClientManagementRegistry({}, {})", consumerId, clientDescriptor, newRegistry);
 
     whenFetchClient(consumerId, clientDescriptor).executeOrDelay("client-registry", client -> {
-      boolean hadRegistry = client.getManagementRegistry().isPresent();
-      client.setManagementRegistry(newRegistry);
-      if (!hadRegistry) {
+      if (!newRegistry.equals(client.getManagementRegistry().orElse(null))) {
+        client.setManagementRegistry(newRegistry);
         firingService.fireNotification(new ContextualNotification(client.getContext(), Notification.CLIENT_REGISTRY_AVAILABLE.name()));
       }
     });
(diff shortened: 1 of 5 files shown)


management/monitoring-service/src/main/java/org/terracotta/management/service/monitoring/TopologyService.java
```diff
@@ -437,9 +437,8 @@ void willSetClientManagementRegistry(long consumerId, ClientDescriptor clientDes
     LOGGER.trace("[{}] willSetClientManagementRegistry({}, {})", consumerId, clientDescriptor, newRegistry);
 
     whenFetchClient(consumerId, clientDescriptor).executeOrDelay("client-registry", client -> {
-      boolean hadRegistry = client.getManagementRegistry().isPresent();
-      client.setManagementRegistry(newRegistry);
-      if (!hadRegistry) {
+      if (!newRegistry.equals(client.getManagementRegistry().orElse(null))) {
+        client.setManagementRegistry(newRegistry);
```
mathieucarbou, line 441: whatever you decide to change you always need to call `client.setManagementRegistry(newRegistry);` (not just inside the if) because otherwise the model in the monitoring service won't be updated with the new management registry sent from the client.

Here, the code handles the case where monitoring service can receive a new client registry BEFORE platform informs us of the fetch...

management/monitoring-service/src/main/java/org/terracotta/management/service/monitoring/TopologyService.java
```diff
@@ -502,11 +501,8 @@ void willSetEntityManagementRegistry(long consumerId, String serverName, Managem
     LOGGER.trace("[{}] willSetEntityManagementRegistry({}, {})", consumerId, serverName, names);
 
     whenServerEntity(consumerId, serverName).executeOrDelay("entity-registry", serverEntity -> {
-      List<String> names2 = newRegistry.getCapabilities().stream().map(Capability::getName).collect(Collectors.toList());
-
-      boolean hadRegistry = serverEntity.getManagementRegistry().isPresent();
-      serverEntity.setManagementRegistry(newRegistry);
-      if (!hadRegistry) {
+      if (!newRegistry.equals(serverEntity.getManagementRegistry().orElse(null))) {
+        serverEntity.setManagementRegistry(newRegistry);
```
mathieucarbou, line 505: same comment here but for the entities ;-)

mathieucarbou (changes requested):

management/monitoring-service/src/main/java/org/terracotta/management/service/monitoring/TopologyService.java
mathieucarbou: Oh sorry my mistake... I am re-reading my comment and it does not make any sense lol. I must be tried.
So yes the code is good sorry!

I remember though we had to add this check on the backend (which was not there before) because too many events were sent, and we needed to differentiate in the TMS between when a registry is FIRST made available (== we can start collecting stats on client and do management stuff) VS when the registry is updated.

There is a LOT of updates in a registry. So firing an event each time there is an update can possibly make the TMC run too many refreshes for nothing and TMS do management calls for nothing.

Maybe a solution could be to have:
- `CLIENT_REGISTRY_AVAILABLE`: when the registry is first made available
- `CLIENT_REGISTRY_UPDATED`: when it is updated (**new event**)

Because be careful, in TMS, for most of the cases right now are triggering actions as soon as something becomes "manageable" (has management registry), and until now we were not interested in these updates.

mathieucarbou (approved):

PR #999, opened 2021-05-15
Cleanup: Huge XML cleanup

mathieucarbou:
- Moved all XML parsers to the converter (which was depending on them)
- Stripped the XML parsers to their bare minimal requirement (which is outputting jaxb objects)
- Removed any dependency towards tc-config-parser plus the descriptors
- Removed any dependency towards tcconfig-schemas in plugins / services
- Introduced a new interface `XmlParser` in DC api to replace the previous ones, and which is more focused on parsing jaxb objects


mathieucarbou:
@chrisdennis : if this refactoring works for you, we can apply the same recipe for EE. The idea is to move the parser and jaxb objects to the converter EE module, and after that we can remove the dependency the converter had with the plugins / services.

The idea is to move out the XML code from the plugins / services to the converter tool, and then cleanup the dependencies of the converter tool.

Also, no need anymore to depend on tc-parser and the 2 interfaces. We can replace that easily with another one.

chrisdennis (changes requested):
I'm not doubting that this works... but I'm not a fan of the approach here. I'd much rather treat the old XML parsing code as legacy that we support. That means minimal changes to avoid accidentally changing the semantics of the parsing behavior. To that end I'd much prefer moving the XML parsing classes to `<service-provider-module>/xml` child modules with distinct coordinates. The classes can all move without being modified which minimized the chances of any behavioral changes, the upgrade tool depends on those modules (that can transitively depend on their parents if necessary) and the new kit gets naturally cleaned of all xml traces.

mathieucarbou:
> I'd much rather treat the old XML parsing code as legacy that we support. 

=> the removed code was dead and was never called. We do not use the parse methods and old interfaces anymore. Whether this is in this PR or before in DC...

mathieucarbou:
will try to look at this one in some spare time...

PR #1004, opened 2021-05-17
Fix: Problems in the activation flow

mathieucarbou:
- To remove the un-necessary boolean state "activated"
- To introduce the concept of enabling/disabling Nomad
- To correctly report when a node is activated
- To avoid starting a node that was incorrectly activated with an alternate topology if not started in repair mode

### DIFF
--- a/dynamic-config/server/configuration-provider/src/main/java/org/terracotta/dynamic_config/server/configuration/startup/ConfigurationGeneratorVisitor.java
+++ b/dynamic-config/server/configuration-provider/src/main/java/org/terracotta/dynamic_config/server/configuration/startup/ConfigurationGeneratorVisitor.java
@@ -85,14 +85,6 @@ public class ConfigurationGeneratorVisitor {
     this.server = server;
   }
 
-  public boolean isUnConfiguredMode() {
-    return unConfiguredMode;
-  }
-
-  public boolean isRepairMode() {
-    return repairMode;
-  }
-
   public StartupConfiguration generateConfiguration() {
     requireNonNull(nomadServerManager);
     requireNonNull(nodeContext);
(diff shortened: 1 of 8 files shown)


dynamic-config/server/configuration-provider/src/main/java/org/terracotta/dynamic_config/server/configuration/startup/ConfigurationGeneratorVisitor.java
```diff
@@ -166,7 +158,11 @@ void startUsingConfigRepo(Path nodeConfigurationDir, String nodeName, boolean re
     DynamicConfigService dynamicConfigService = nomadServerManager.getDynamicConfigService();
     TopologyService topologyService = nomadServerManager.getTopologyService();
     if (!repairMode) {
-      dynamicConfigService.activate(topologyService.getUpcomingNodeContext().getCluster(), dynamicConfigService.getLicenseContent().orElse(null));
+      // if not in repair mode, ensure we do not start a node with an empty or prepared activate change that has not been completed
+      if (!nomadServerManager.getConfiguration().isPresent()) {
```
mobasherul, line 162: So if the server didn't had the activation change and u started the server which can accept nomad change what exactly was happening , I guess the subsequent nomad change would have failed in discovery right @mathieucarbou ?

mathieucarbou: So I have discovered that while doing another PR (make failover optional). This would fail in the cluster validation (in the new PR). But the problem is much more that this "path" was not properly handled and we didn't display the proper error message to the user...

mathieucarbou: But the main change of this PR is really the removal of the boolean `isActivated` to replace this weird boolean by checking the Nomad state instead. This is more reliable.

mobasherul (approved):

mathieucarbou:
merging - builds passed before rebase, 2 passed already now.

PR #1005, opened 2021-05-17
Do not create config directories when in diagnostic mode

mathieucarbou:
added javadoc and updated a test

### DIFF
--- a/dynamic-config/server/configuration-provider/src/main/java/org/terracotta/dynamic_config/server/configuration/service/NomadServerManager.java
+++ b/dynamic-config/server/configuration-provider/src/main/java/org/terracotta/dynamic_config/server/configuration/service/NomadServerManager.java
@@ -18,7 +18,6 @@ package org.terracotta.dynamic_config.server.configuration.service;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.terracotta.dynamic_config.api.model.NodeContext;
-import org.terracotta.dynamic_config.api.model.UID;
 import org.terracotta.dynamic_config.api.model.nomad.ClusterActivationNomadChange;
 import org.terracotta.dynamic_config.api.model.nomad.FormatUpgradeNomadChange;
 import org.terracotta.dynamic_config.api.model.nomad.LockConfigNomadChange;
@@ -32,6 +31,7 @@ import org.terracotta.dynamic_config.api.service.DynamicConfigService;
 import org.terracotta.dynamic_config.api.service.IParameterSubstitutor;
 import org.terracotta.dynamic_config.api.service.TopologyService;
 import org.terracotta.dynamic_config.server.api.ConfigChangeHandlerManager;
+import org.terracotta.dynamic_config.server.api.DelegatingDynamicConfigNomadServer;
 import org.terracotta.dynamic_config.server.api.DynamicConfigEventFiring;
 import org.terracotta.dynamic_config.server.api.DynamicConfigEventService;
 import org.terracotta.dynamic_config.server.api.DynamicConfigNomadServer;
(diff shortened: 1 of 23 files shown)


mathieucarbou:
@mobasherul @tmesic99 : 

Refactored the nomad manager initialization process to avoid creating an empty structure when starting a node unconfigured.

The initialization process now only initializes the Nomad system when a node is activated.
The Nomad server is consequently optionally there, and when started in diagnostic mode, a node is then not creating its config folder structure immediately.

dynamic-config/server/configuration-provider/src/main/java/org/terracotta/dynamic_config/server/configuration/service/NomadServerManager.java
```diff
@@ -211,26 +220,36 @@ public synchronized boolean isNomadEnabled() {
...
+    if (nomadServer == null) {
+      this.configurationManager.createDirectories();
+      try {
+        this.nomadServer = nomadServerFactory.createServer(configurationManager, nodeContext.getNode().getName(), getEventFiringService());
+      } catch (SanskritException | NomadException | ConfigStorageException e) {
+        throw new UncheckedNomadException("Exception initializing Nomad Server: " + e.getMessage(), e);
+      }
+    }
+
+    router.clear();
```
mobasherul, line 239: why to clear the nomad change processors in between enabling and disabling ?

dynamic-config/server/configuration-provider/src/main/java/org/terracotta/dynamic_config/server/configuration/service/NomadServerManager.java
```diff
@@ -152,42 +147,56 @@ public NomadPermissionChangeProcessor getNomadPermissionChangeProcessor() {
...
+
+  public void configure(Path configPath, NodeContext nodeContext) throws UncheckedNomadException {
+    requireNonNull(configPath);
+    requireNonNull(nodeContext);
+
+    this.configurationManager = new NomadConfigurationManager(configPath, parameterSubstitutor);
+
+    // the eventFiringService is used by callers to fire events
+    // events are received by the DC service, processed, and fired back to the listeners registered in the event service
+    DynamicConfigEventServiceImpl eventService = new DynamicConfigEventServiceImpl();
```
mobasherul, line 195: Shouldn' these services(DynamicConfigEventService, DynamicConfigEventFiringService) also get intitilaized during activation only since it will listen to nomad changes and nomad server that will be available during activation.

dynamic-config/server/configuration-provider/src/main/java/org/terracotta/dynamic_config/server/configuration/service/NomadServerManager.java
mathieucarbou: because the route is not re-initialized: it is final and reused if someone does a enable-diable-enable.

dynamic-config/server/configuration-provider/src/main/java/org/terracotta/dynamic_config/server/configuration/service/NomadServerManager.java
mathieucarbou: yes! I will look into that ! that's true.

mathieucarbou: I did a change to initialize (final) these services to avoid the dupliction but I prefer to keep the `register` calls in the init methods (relaod/configure) because this is where objects are created and these need to be called once only.

dynamic-config/server/configuration-provider/src/main/java/org/terracotta/dynamic_config/server/configuration/service/NomadServerManager.java
mobasherul: may be we can check if its already populated with entries since clearing it and then popluating with same entries is weird ?

mathieucarbou: no because the entries depend on the value of the current topology when the router is initialized... It would be correct if those values would not change but the router being a global service linked to a method initializing it with values referencing a nodeContext at a specific point in time, I think this is unsafe to not clear and recreate the entries with the right topology in place when the method is called.

tmesic99 (approved):

PR #1006, opened 2021-05-17
TDB-5371: Do not load DC extensions (== offheap, data dirs, lease configs) when starting in configuration / repair mode

mathieucarbou:


myronkscott (approved):

data-root-resource/src/main/java/org/terracotta/config/data_roots/EmptyDataDirs.java
```diff
@@ -0,0 +1,50 @@
...
+import java.util.Set;
+
+import static java.util.Collections.emptySet;
+
+/**
+ * @author Mathieu Carbou
+ */
+@CommonComponent
+public class EmptyDataDirs implements DataDirs {
+  @Override
```
myronkscott, line 32: Not sure you need, want, this

mathieucarbou: So yeah I can remove it know but first I need to update the EE PR. 
It was requierd if we wanted to keep the exact same logic as now (to avoid a NPE).
But since `metadata-dir` is required, some situation cannot occur anymore. So I will update the EE code first and then update this PR :-)

PR #1007, opened 2021-05-18
BUGFIX: install the license AFTER the config directories have been cr…

mathieucarbou:
…eated


mathieucarbou:
@mobasherul @tmesic99 : if you can please merge this PR when you can. I have left a bug in a previous one that I saw when testing in EE. Thanks!

mobasherul (approved):

PR #1008, opened 2021-05-20
bump core for bug fixes

myronkscott:


myronkscott:
Fixes logging issues when no file is specified
Fixes retirement tracking
cleanup client handshake threading
cleanup healthchecker config
cleanup server bootstrap and shutdown for both inline and script start

mathieucarbou (approved):

PR #1009, opened 2021-05-25
Unhide -auto-activate

mathieucarbou:
Recall that this option was initially introduced to be able to start a node with a cluster topology (from CLI or config file), directly pre-activated without any interaction using config-tool. This can help in some cases where remote com. is difficult, or docker image where we want the node to start and join an existing cluster, etc.


mathieucarbou:
replaced by #1010

PR #1011, opened 2021-06-03
Allow some extensions like security to be loaded even in diagnostic mode

mathieucarbou:


mathieucarbou:
@chrisdennis @myronkscott : this is required for security purposes because security is using the dynamic config extension mechanism which is used to load both extended serviced AND service configurations...

dynamic-config/server/server-api/src/main/java/org/terracotta/dynamic_config/server/api/DynamicConfigExtension.java
```diff
@@ -28,6 +28,12 @@
  */
 public interface DynamicConfigExtension {
 
+  /**
+   * @return true if this extension can only be loaded when a node is starting from an already created configuration folder.
+   * If the node is starting in diagnostic mode (for repair or initially) then the extension won't be loaded except if this method returns false.
+   */
+  default boolean onlyWhenNodeConfigured() { return true; }
```
chrisdennis, line 35: This is always inverted when used. Should we think about inverting it? `isAvailableInDiagnosticMode()`?

chrisdennis (approved):

PR #1012, opened 2021-06-04
Display, in startup console, in warning, the CLI parameters that do not match the persisted config anymore

mathieucarbou:
Here are some examples of output that we can see in the console log at startup:

```
2021-06-04 16:23:23,871 [Thread-299] INFO org.terracotta.console - 

=================================================================================================================
Node is starting from an existing configuration directory.
The following command-line (CLI) parameters differ from the loaded configuration. 
They have been ignored and can be removed from the CLI.
 - Setting: 'name': CLI=node-1-1 vs CONFIG=foo
=================================================================================================================
```

```
2021-06-04 16:22:44,530 [Thread-84] INFO org.terracotta.console - 

=================================================================================================================
Node is starting from an existing configuration directory.
The following command-line (CLI) parameters differ from the loaded configuration. 
They have been ignored and can be removed from the CLI.                                       
 - Setting: 'log-dir': CLI=logs vs CONFIG=logs/stripe1
 - Setting: 'metadata-dir': CLI=metadata vs CONFIG=metadata/stripe1
 - Setting: 'data-dirs': CLI=main:data-dir vs CONFIG=main:user-data/main/stripe1
 - Setting: 'offheap-resources': CLI=foo:1GB,main:512MB vs CONFIG=main:512MB,second:1GB
 - Setting: 'client-reconnect-window': CLI=10s vs CONFIG=<unspecified>
 - Setting: 'cluster-name': CLI=tc-cluster vs CONFIG=my-cluster
=================================================================================================================
```

### DIFF
--- a/dynamic-config/model/src/main/java/org/terracotta/dynamic_config/api/model/NodeContext.java
+++ b/dynamic-config/model/src/main/java/org/terracotta/dynamic_config/api/model/NodeContext.java
@@ -62,6 +62,23 @@ public class NodeContext implements Cloneable {
     return stripe.getUID();
   }
 
+  public Optional<String> getProperty(Setting setting) {
+    return setting.getProperty(getObject(setting.getScope()));
+  }
+
+  public PropertyHolder getObject(Scope scope) {
+    switch (scope) {
+      case CLUSTER:
+        return getCluster();
+      case STRIPE:
+        return getStripe();
+      case NODE:
+        return getNode();
+      default:
+        throw new AssertionError(scope);
+    }
+  }
+
   @Override
   @SuppressWarnings("MethodDoesntCallSuperMethod")
   @SuppressFBWarnings("CN_IDIOM_NO_SUPER_CALL")
--- a/dynamic-config/server/configuration-provider/src/main/java/org/terracotta/dynamic_config/server/configuration/startup/ConfigRepoCommandLineProcessor.java
+++ b/dynamic-config/server/configuration-provider/src/main/java/org/terracotta/dynamic_config/server/configuration/startup/ConfigRepoCommandLineProcessor.java
@@ -25,7 +25,11 @@ import org.terracotta.server.Server;
 import java.nio.file.Path;
 import java.util.LinkedHashMap;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Optional;
+
+import static java.lang.System.lineSeparator;
+import static org.terracotta.common.struct.Tuple3.tuple3;
 
 public class ConfigRepoCommandLineProcessor implements CommandLineProcessor {
   private final Options options;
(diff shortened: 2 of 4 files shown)


dynamic-config/server/configuration-provider/src/main/java/org/terracotta/dynamic_config/server/configuration/startup/ConfigRepoCommandLineProcessor.java
```diff
@@ -63,6 +71,34 @@ public void process() {
...
+              + "=================================================================================================================" + lineSeparator()
+              + "Node is starting from an existing configuration directory.                                                       " + lineSeparator()
+              + "The following command-line (CLI) parameters differ from the loaded configuration.                                " + lineSeparator()
+              + "They have been ignored and can be removed from the CLI.                                                          " + lineSeparator()
+              + " - " + String.join("\n - ", mismatches) + lineSeparator()
+              + "=================================================================================================================" + lineSeparator()
+          );
+        }
+      }
+
```
chrisdennis, line 101: If you insist on using streams I think this is a lot clearer:
```java
cliOptions.keySet().stream()
        .map(setting -> tuple2(setting,
                tuple2OfOptionals(setting.getProperty(getContext(alternate, setting.getScope())),
                        setting.getProperty(getContext(loadedNodeContext, setting.getScope())))))
        .filter(setting -> {
          Tuple2<String, String> values = setting.getT2();
          return !Objects.equals(values.getT1(), values.getT2());
        })
        .map(mismatch -> " - Setting: '" + mismatch.getT1() + "': CLI=" + mismatch.getT2().findT1().orElse("<unavailable>") + " vs CONFIG=" + mismatch.getT2().findT2().orElse("<unavailable>") + lineSeparator())
        .reduce(String::concat).ifPresent(mismatches -> server.console(lineSeparator() + lineSeparator()
                + "=================================================================================================================" + lineSeparator()
                + "Node is starting from an existing configuration directory.                                                       " + lineSeparator()
                + "The following command-line (CLI) parameters differ from the loaded configuration.                                " + lineSeparator()
                + "They have been ignored and can be removed from the CLI.                                                          " + lineSeparator()
                + mismatches
                + "=================================================================================================================" + lineSeparator()
        ));
```
but to be honest I'm not sure streams are buying anything here. I think the following is clearer still:
```java
List<String> mismatches = new ArrayList<>();
for (Setting setting : cliOptions.keySet()) {
  Optional<String> fromCli = setting.getProperty(getContext(alternate, setting.getScope()));
  Optional<String> fromRepo = setting.getProperty(getContext(loadedNodeContext, setting.getScope()));
  if (!fromCli.equals(fromRepo)) {
    mismatches.add(" - Setting: '" + setting + "': CLI=" + fromCli.orElse("<unavailable>") + " vs CONFIG=" + fromRepo.orElse("<unavailable>") + lineSeparator());
  }
}
if (!mismatches.isEmpty()) {
  server.console(lineSeparator() + lineSeparator()
          + "=================================================================================================================" + lineSeparator()
          + "Node is starting from an existing configuration directory.                                                       " + lineSeparator()
          + "The following command-line (CLI) parameters differ from the loaded configuration.                                " + lineSeparator()
          + "They have been ignored and can be removed from the CLI.                                                          " + lineSeparator()
          + String.join(lineSeparator(), mismatches) + lineSeparator()
          + "=================================================================================================================" + lineSeparator());
}
```
I've extracted a utility method here... but this should probably go on the `Scope` enum:
```java
private static PropertyHolder getContext(NodeContext alternate, Scope scope) {
  switch (scope) {
    case NODE:
      return alternate.getNode();
    case STRIPE:
      return alternate.getStripe();
    case CLUSTER:
      return alternate.getCluster();
  }
  throw new AssertionError();
}
```

chrisdennis (changes requested):

dynamic-config/server/configuration-provider/src/main/java/org/terracotta/dynamic_config/server/configuration/startup/ConfigRepoCommandLineProcessor.java
mathieucarbou: I definitely started my stream with the wrong set of data ;-)
I took your ideas and changes things a bit, kept `unspecified` because this is really what it means - not unavailable because a setting can be available (resolved) but not specified (configured).

dynamic-config/model/src/main/java/org/terracotta/dynamic_config/api/model/NodeContext.java
```diff
@@ -62,6 +62,23 @@ public UID getStripeUID() {
     return stripe.getUID();
   }
 
+  public Optional<String> getProperty(Setting setting) {
+    return setting.getProperty(getObject(setting.getScope()));
+  }
+
+  public PropertyHolder getObject(Scope scope) {
```
chrisdennis, line 69: Why not convert this switch over an enum to a dynamic dispatch on the enum?

chrisdennis (approved):

dynamic-config/model/src/main/java/org/terracotta/dynamic_config/api/model/NodeContext.java
mathieucarbou: Because I don't like that the enum class depends on NodeContext, Node, Cluster, etc...

PR #1023, opened 2021-06-29
Add custom manifest entries in jars

mathieucarbou:
### This PR contains only:
- [x] Added custom manifest entries in jars


tmesic99 (approved):

chrisdennis (commented):
I don't understand why all of these Java classes are here... why can't all of this be done using the manifest?

mathieucarbou:
> I don't understand why all of these Java classes are here... why can't all of this be done using the manifest?

There are 2 layers:

1. The implementation of the `ExtensionInfo` interface => service used by platform  to discover and log the installed extensions
2. The retrieval of kit information (kit version, build info, etc + installed components) through diagnostic port

The second layer is using the first one, but by asking the extension info for some keys (name, description, version, etc). Thee keys are not linked to any implementation details. This is the current chosen implementation to re-route the key values to the existing manifest entries in platform jars. So the implementation is specific to platform. Ehcache has another one (otehr manifest entries)

I will just update the code to make the skeleton in a standalone common module so that Ehcache and DC can depend on that and in order to remove the code duplication that was initially there.

pom.xml
```diff
@@ -291,6 +291,10 @@
           <configuration>
             <archive>
               <manifestFile>${project.build.outputDirectory}/META-INF/MANIFEST.MF</manifestFile>
+              <manifestEntries>
+                <Ext-TC-Artifact>${project.artifactId}</Ext-TC-Artifact>
+                <Ext-TC-Version>${project.version}</Ext-TC-Version>
```
mathieucarbou, line 296: @myronkscott @chrisdennis : if you agree with those names, I am proposing to expose for each jar 2 custom manifest entries: `Ext-TC-Artifact` and `Ext-TC-Version` to output the artifact ID (not name) and the version. These are the only 2 fields that can be consistently sourced out from a Maven or Gradle project because these fields are required. `project.name` and `project.description` are not required and worse, can be inherited (in case of description) which leads to false information.

pom.xml
```diff
@@ -291,9 +291,31 @@
           <configuration>
             <archive>
               <manifestFile>${project.build.outputDirectory}/META-INF/MANIFEST.MF</manifestFile>
+              <manifestEntries>
+                <Implementation-Version>${project.version}</Implementation-Version>
+                <Artifact>${project.artifactId}</Artifact>
+                <Implementation-SCM-Revision>${buildNumber}</Implementation-SCM-Revision>
+                <Implementation-SCM-Branch>${scmBranch}</Implementation-SCM-Branch>
```
mathieucarbou, line 298: I am proposing these new manifest entries added always the same way:
```
Implementation-Version: 5.8-SNAPSHOT
Artifact: lease-entity-common
Implementation-SCM-Revision: 8e4ecd1ebe0c920a186c777d78c7a2b647fa4ba7
Implementation-SCM-Branch: TDB-5308
```

pom.xml
```diff
@@ -291,9 +291,32 @@
...
+                <Build-Timestamp>${maven.build.timestamp}</Build-Timestamp>
+              </manifestEntries>
```
mathieucarbou, line 300: I am proposing these new manifest entries added always the same way:
```
Implementation-Version: 5.8-SNAPSHOT
Artifact: lease-entity-common
Implementation-SCM-Revision: 8e4ecd1ebe0c920a186c777d78c7a2b647fa4ba7
Implementation-SCM-Branch: TDB-5308
Build-Timestamp: 2021-06-30T20:37:10Z
```

diagnostic/client/src/main/java/org/terracotta/diagnostic/client/DiagnosticServiceImpl.java
```diff
@@ -168,6 +175,23 @@ public LogicalServerState getLogicalServerState() throws DiagnosticOperationTime
...
+    Matcher sha = Pattern.compile(".*([0-9a-fA-F]{40}).*").matcher(b);
+    String revision = sha.matches() ? sha.group(1) : "<unknown>";
+    DateTimeFormatter dtf = DateTimeFormatter.ofPattern("yyyy-MM-dd 'at' HH:mm:ss z"); // from core
+    Instant timestamp = null;
+    try {
+      timestamp = dtf.parse(b.substring(0, 26), Instant::from);
+    } catch (DateTimeException e) {
+      LOGGER.debug(b, e);
+    }
+    return new KitInformation(version, revision, timestamp);
```
mathieucarbou, line 192: @myronkscott : right now, from the Mbean methods, we are able to extract the version, revision and timestamp and branch. But once you can expose in a more structured way these information we will be able to get rid of these hacky string parsing.

diagnostic/client/src/main/java/org/terracotta/diagnostic/client/DiagnosticServiceImpl.java
```diff
@@ -168,6 +175,30 @@ public LogicalServerState getLogicalServerState() throws DiagnosticOperationTime
...
+
+    DateTimeFormatter dtf = DateTimeFormatter.ofPattern("yyyy-MM-dd 'at' HH:mm:ss z"); // from core
+    Instant timestamp = null;
+    try {
+      timestamp = dtf.parse(b.substring(0, 26), Instant::from);
+    } catch (DateTimeException e) {
+      LOGGER.debug(b, e);
+    }
+
+    return new KitInformation(version, revision, branch, timestamp);
```
mathieucarbou, line 199: @myronkscott : right now, from the Mbean methods, we are able to extract the version, revision and timestamp and branch. But once you can expose in a more structured way these information we will be able to get rid of these hacky string parsing.

dynamic-config/testing/system-tests/src/test/java/org/terracotta/dynamic_config/system_tests/diagnostic/DiagnosticIT.java
```diff
@@ -66,11 +66,9 @@ public void test_get_kit_information() throws Exception {
         null,
         objectMapperFactory)
     ) {
-      System.out.println(diagnosticService);
-      //System.out.println(diagnosticService.);
+      System.out.println(diagnosticService.getKitInformation());
```
mathieucarbou, line 69: @tmesic99 : this is how to get the kit information object from diagnostic port, which contains the version, revision, branch and timestamp information

pom.xml
chrisdennis: Do we need more than: `Implementation-Title` and `Implementation-Version`?:
```
Implementation-Title: lease-entity-common
Implementation-Version: 5.8-SNAPSHOT 8e4ecd1ebe0c920a186c777d78c7a2b647fa4ba7
```

mathieucarbou: So since we concluded that these things would be used only by core to output more info about the components and for us for troubleshooting, I find useful to also have the revision (at least). I have put the branch and timestamp because they cost nothing to add, but we could remove them.
I would really like to keep the revision though because this is more useful to quickly grab and look at the right code.

For the naming of the keys, I will change yes 👍

mathieucarbou: I don't see why we could have the revision inside the version. These are 2 separate things and it just makes the code that will use that ugly... Also the usage of both is not the same. We care about revision to look at the right codebase. Whereas the version is something that is worth outputting in the logs IMO

chrisdennis: Meh... we already merge them in the output of the server version. That server version string has had SVN versions embedded in it for the customer to see for years. I don't see the harm.

mathieucarbou: I disagree: the output of the server version is : `Terracotta x.y.z`. this is the build ID (another method) that contains the revision, timestamp and branch. The combination of both is outputed in the logs. But the server version does not contain the revision.
And all these fields are not structured / typed. A version is a version and a revision a revision: IMO a field that contains both is neither a version neither a revision. This is a a mix bag of information.

The product info object is a bit more structured but we do not have access to it.

IMO we should prone something that is consistent with what we will get through diagnostic port. And we could get the version , the revision, the timestamp, etc all in separate fields.

That is why I am more in favor of having manifest entries that really represent what they mean.

Otherwise, what's the point of all this effort. Let's just have an entry called "mud" with the value being a concatenation of artifact, version, revision. But then with such description, nobody could really use it except reading it. 

If we want to be able to read that manifest information and be able to group the information, then let's have the right fields for that with the keys representing what they mean.

mathieucarbou: @tmesic99 @chrisdennis : I have extracted the 2 commits for TDB-5308 in another PR. I don't want to delay TDB-5308 because of these discussions that are out of scope for TDB-5308.

Here is the other PR to merge Tom: https://github.com/Terracotta-OSS/terracotta-platform/pull/1024 :-)

pom.xml
```diff
@@ -291,9 +291,30 @@
           <configuration>
             <archive>
               <manifestFile>${project.build.outputDirectory}/META-INF/MANIFEST.MF</manifestFile>
+              <manifestEntries>
+                <Implementation-Title>${project.artifactId}</Implementation-Title>
+                <Implementation-Version>${project.version}</Implementation-Version>
+                <Implementation-Revision>${buildNumber}</Implementation-Revision>
+              </manifestEntries>
```
mathieucarbou, line 298: Updated to only keep the artifact name, version and commit hash

mathieucarbou: Note: `Implementation-Vendor-Id` set to `${project.groupId}` could be added to : this would help organize / regroupe the jars based on a common group ID. This is what jackson is using in their manifest files.

mathieucarbou: Example:

```
Implementation-Title: lease-entity-server
Implementation-Version: 5.8-SNAPSHOT
Implementation-Vendor-Id: org.terracotta
Implementation-Revision: fb4f88ff7592ad4c54456426c34c9b6d01e7dc75
```

mathieucarbou:
Replacing by #1027 (wrong branch)

PR #1024, opened 2021-07-01
TDB-5308

mathieucarbou:
This PR contains only:
- [x] Cleanup any code related to the existing core `ExtensionInfo` interface
- [x] Rename `LogicalServerStateProvider` => `DiagnosticExtensions` to support extending the diagnostic layer easily
- [x] Add a method that exposes in 1 call the kit information taken from the current mBeans version and buildID.


dynamic-config/testing/system-tests/src/test/java/org/terracotta/dynamic_config/system_tests/diagnostic/DiagnosticIT.java
```diff
@@ -66,11 +66,9 @@ public void test_get_kit_information() throws Exception {
         null,
         objectMapperFactory)
     ) {
-      System.out.println(diagnosticService);
-      //System.out.println(diagnosticService.);
+      System.out.println(diagnosticService.getKitInformation());
       // output:
-/*
-*/
+      // KitInformation{version='5.8.2-pre6', revision='4450fe6fc2c174abd3528b8636b3296a6a79df00', branch='UNKNOWN', timestamp=2021-06-29T20:54:46Z}
```
mathieucarbou, line 71: @tmesic99 : this is how to get the kit information object from diagnostic port, which contains the version, revision, branch and timestamp information. What we need to display for TMC + compat' purposes for tooling.

diagnostic/client/src/main/java/org/terracotta/diagnostic/client/DiagnosticServiceImpl.java
```diff
@@ -168,6 +175,30 @@ public LogicalServerState getLogicalServerState() throws DiagnosticOperationTime
...
+
+    DateTimeFormatter dtf = DateTimeFormatter.ofPattern("yyyy-MM-dd 'at' HH:mm:ss z"); // from core
+    Instant timestamp = null;
+    try {
+      timestamp = dtf.parse(b.substring(0, 26), Instant::from);
+    } catch (DateTimeException e) {
+      LOGGER.debug(b, e);
+    }
+
+    return new KitInformation(version, revision, branch, timestamp);
```
mathieucarbou, line 199: @myronkscott : right now, from the Mbean methods, we are able to extract the version, revision and timestamp and branch. But once you can expose in a more structured way these information we will be able to get rid of these hacky string parsing.

PR #1027, opened 2021-07-02
Add custom manifest entries in jars

mathieucarbou:


chrisdennis (approved):
As with ehcache/ehcache3#2917 this look fine, but @tmesic99 might want to look it over I guess.

PR #1028, opened 2021-07-02
Add  DiagnosticService#getKitInformation() to expose in 1 call the equivalent of ServerInfo#getVersion() + ServerInfo#getBuildId()

mathieucarbou:


dynamic-config/testing/system-tests/src/test/java/org/terracotta/dynamic_config/system_tests/diagnostic/DiagnosticIT.java
```diff
@@ -55,4 +55,31 @@ public void test_access_logical_server_state() throws Exception {
...
+      /*
+version=5.8.2-pre6
+revision=4450fe6fc2c174abd3528b8636b3296a6a79df00
+branch=UNKNOWN
+timestamp=2021-06-29T20\:54\:46Z
+       */
+      System.out.println(diagnosticService.getKitInformation().getVersion()); // i.e.: 5.8.2-pre6
+      System.out.println(diagnosticService.getKitInformation().getBranch()); // i.e.: master
+      System.out.println(diagnosticService.getKitInformation().getRevision()); // i.e.: 4450fe6fc2c174abd3528b8636b3296a6a79df00
+      System.out.println(diagnosticService.getKitInformation().getTimestamp()); // i.e.: 2021-06-29T20:54:46Z
```
mathieucarbou: @tmesic99 : FYI, from the diagnostic interface, it was possible already to get the **version** and **build ID** (through exposed JMX beans). The problem is that it was split between 2 calls (getServer and getBuildID) and besides that it required parsing. So I added `diagnosticService.getKitInformation()` so that we can get the equivalent information from 1 call only. **This is the exact same information, but structured and returned in 1 call only.**

For now, this will help us continue our user stories.

But be aware that @myronkscott is currently working on a better integrated api to get the kit version (and client versions) from the connection api or something like that. So we can use this call for now to be able to implement what we need regarding the use of kit version.

tmesic99 (approved):

PR #1034, opened 2021-07-27
New Apis for offheap usage listeners.

mobasherul:


offheap-resource/src/main/java/org/terracotta/offheapresource/OffHeapResourceImpl.java
```diff
@@ -234,12 +269,22 @@ public boolean setCapacity(long size) throws IllegalArgumentException {
...
+        stateUpdated(currentState, newState);
         return true;
       }
     }
   }
-  
+
+  @Override
+  public void addUsageListener(UUID listenerUUID, float threshold, Consumer<OffHeapUsageEvent> consumer) {
+    listenerMap.put(listenerUUID, new OffHeapUsageListener(threshold, consumer));
```
chrisdennis, line 280: Surely you mean `putIfAbsent(...)` here?

offheap-resource/src/main/java/org/terracotta/offheapresource/OffHeapResourceImpl.java
```diff
@@ -234,12 +269,22 @@ public boolean setCapacity(long size) throws IllegalArgumentException {
...
+  }
+
+  @Override
+  public void removeUsageListener(UUID listenerUUID) {
+    listenerMap.remove(listenerUUID);
```
chrisdennis, line 285: Should this have a different behavior if the mapping is present vs if it is absent?

offheap-resource/src/main/java/org/terracotta/offheapresource/OffHeapUsageEventImpl.java
```diff
@@ -0,0 +1,57 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.terracotta.offheapresource;
+
+public class OffHeapUsageEventImpl implements OffHeapUsageEvent {
+  private final float occupancy;
```
chrisdennis, line 19: Isn't `occupancy` just a simple bit of FPU calc from the other values? Can't you just define a default implementation?

offheap-resource/src/main/java/org/terracotta/offheapresource/OffHeapResourceImpl.java
```diff
@@ -234,12 +269,22 @@ public boolean setCapacity(long size) throws IllegalArgumentException {
...
-        stateUpdated(newState);
+        stateUpdated(currentState, newState);
         return true;
       }
     }
   }
-  
+
+  @Override
+  public void addUsageListener(UUID listenerUUID, float threshold, Consumer<OffHeapUsageEvent> consumer) {
```
chrisdennis, line 279: Shouldn't this take a rising/falling type?

chrisdennis (changes requested):

offheap-resource/src/main/java/org/terracotta/offheapresource/OffHeapResourceImpl.java
```diff
@@ -145,15 +150,45 @@ public boolean reserve(long size) throws IllegalArgumentException {
...
     long used = newState.getUsed();
+    long prevUsed = prevState.getUsed();
+    long prevCapacity = prevState.getCapacity();
+
+    if (used > prevUsed || capacity < prevCapacity) {
+      // check for rising event.
+      float occupancy = (used * 1.0f) / capacity;
+      OffHeapUsageEvent offHeapUsageEvent = null;
+      for (OffHeapUsageListener offHeapUsageListener : listenerMap.values()) {
+        if (!offHeapUsageListener.isFired() && offHeapUsageListener.getThreshold() < occupancy) {
```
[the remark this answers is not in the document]
mobasherul: I did that for scenarios in which threshold is 0.7f rising and it send the event for the first time when it crosses 0.7f but when it will not send the event for subsequent reserve for offheap as it already has sent the event .

offheap-resource/src/main/java/org/terracotta/offheapresource/OffHeapResourceImpl.java
mobasherul: It doesn't require it. It will send the rising/ falling event and on client side I will check what is the actual listner type registered.

offheap-resource/src/main/java/org/terracotta/offheapresource/OffHeapResourceImpl.java
mobasherul: If it is absent it will be caught on client side early and exception will be thrown.

offheap-resource/src/main/java/org/terracotta/offheapresource/OffHeapResourceImpl.java
mobasherul: I want the listener to be re registered in case of application reconnect since I will be sending those listeners with same UUID during reconnect. I am doing this to ensure that if some events have been missed during reconnect I am registering new listener and these listeners will send the events if further reserve/free call happens as the flag has not been set for these listeners.

chrisdennis: Okay, but how do you know the difference between a re-registration and a UUID collision? The caller might know if he expects a "re-registration" but this code doesn't. I guess I don't understand how the offheap resource should have to care about any of that. If you want to know if there is a listener registered there then at least have it return the old listeners do that the caller can check what happened and behave accordingly.

offheap-resource/src/main/java/org/terracotta/offheapresource/OffHeapResourceImpl.java
chrisdennis: Right... but this API needs to look sane on it's own. Otherwise you're practicing building brittle APIs. Just define a sane behavior for failure to remove and implement it.

offheap-resource/src/main/java/org/terracotta/offheapresource/OffHeapResourceImpl.java
chrisdennis: Right, but you're defining an API on an actual API here right? Why not do it right here as well?

chrisdennis (approved):
There are some sharp edges here still, but it seems like it should function okay.

offheap-resource/src/main/java/org/terracotta/offheapresource/OffHeapResourcesProvider.java
```diff
@@ -169,14 +169,14 @@ private boolean addToResources(OffHeapResourceIdentifier identifier, long capaci
       OffHeapResourceImpl offHeapResource = new OffHeapResourceImpl(
           identifier.getName(),
           capacityInBytes,
-          (res, update) -> {
+          (event) -> {
             for (EntityManagementRegistry registry : registries) {
               Map<String, String> attrs = new HashMap<>();
-              attrs.put("oldThreshold", String.valueOf(update.old));
```
chrisdennis, line 175: Is it safe to just change this API? Why can't we preserve the attributes and not have to also change the consumer side?

chrisdennis (approved):

offheap-resource/src/main/java/org/terracotta/offheapresource/OffHeapResourceImpl.java
mobasherul: Because the listener registered here have to send both rising and falling event for specific threshold  irrespective of whether the listener is rising/falling one.

offheap-resource/src/main/java/org/terracotta/offheapresource/OffHeapResourcesProvider.java
mobasherul: Earlier it was used to verify it is rising/falling which is not required anymore and the stuff displayed on tmc does not get affected as well so I removed it.

offheap-resource/src/main/java/org/terracotta/offheapresource/OffHeapResourceImpl.java
mobasherul: will make it throw exception for sanity of this API

offheap-resource/src/main/java/org/terracotta/offheapresource/OffHeapResourceImpl.java
mobasherul: There is no differentiation between normal listener registation and the listeners getting registered as part of client reconnect.Also old listeners are of no use since they contain reference to old client descriptors. UUID collision came to my mind as well but jdk provides enough guarantee that it will not happen atleast for this usecase.

offheap-resource/src/main/java/org/terracotta/offheapresource/OffHeapResourceImpl.java
```diff
@@ -234,20 +248,34 @@ public boolean setCapacity(long size) throws IllegalArgumentException {
...
+    if ((Float.compare(offHeapUsageListener.getThreshold(), occupancy) <= 0)) {
+      OffHeapUsageEvent offHeapUsageEvent = new OffHeapUsageEventImpl(used, offHeapResourceState.getRemaining(), capacity, OffHeapUsageEventType.RISING);
+      offHeapUsageListener.getConsumer().accept(offHeapUsageEvent);
+      offHeapUsageListener.setFiringStatus(true);
+    }
+  }
+
+  @Override
+  public void removeUsageListener(UUID listenerUUID) throws IllegalArgumentException {
+    if (!listenerMap.containsKey(listenerUUID)) {
```
chrisdennis, line 275: Why not just use the `listenerMap.remove(...)` return value?

chrisdennis (approved):

offheap-resource/src/main/java/org/terracotta/offheapresource/OffHeapResourceImpl.java
mobasherul: Its not doing anything after removal.

chrisdennis: `listenerMap.remove(uuid) == null` is exactly the same as `containsKey()`.

PR #1035, opened 2021-07-27
Fix deadlock when a nomad transaction is committed through the Nomad entity at the same time a diagnostic call is made to query the state of the configuration system

mathieucarbou:
I have also updated the other synchronization in this class. Historically this class was containing a lot of states, which are now moved elsewhere, but I forgot to update the sync. So this PR is syncing with more granularity now.

Note: there can still be some corner cases happening... I.e. when nodes are not activated and supposing someone is concurrently exporting a topology to the same node while reading the topology at the same time. But well, clearly this is asking for troubles...

### DIFF
--- a/dynamic-config/server/configuration-provider/src/main/java/org/terracotta/dynamic_config/server/configuration/service/Licensing.java
+++ b/dynamic-config/server/configuration-provider/src/main/java/org/terracotta/dynamic_config/server/configuration/service/Licensing.java
@@ -1,0 +1,147 @@
+/*
+ * Copyright Terracotta, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.terracotta.dynamic_config.server.configuration.service;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.terracotta.dynamic_config.api.model.Cluster;
+import org.terracotta.dynamic_config.api.model.License;
+import org.terracotta.dynamic_config.server.api.InvalidLicenseException;
+import org.terracotta.dynamic_config.server.api.LicenseService;
+
+import java.io.IOException;
+import java.io.UncheckedIOException;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.StandardCopyOption;
+import java.util.Optional;
+import java.util.concurrent.locks.ReadWriteLock;
+import java.util.concurrent.locks.ReentrantReadWriteLock;
+
+import static java.util.Objects.requireNonNull;
+
+/**
+ * @author Mathieu Carbou
+ */
+class Licensing {
+
+  private static final Logger LOGGER = LoggerFactory.getLogger(Licensing.class);
+  private static final String LICENSE_FILE_NAME = "license.xml";
+
+  // guard access to the installed license
+  private final ReadWriteLock licenseLock = new ReentrantReadWriteLock();
+
+  private final LicenseService licenseService;
+  private final Path licenseFile;
+
+  public Licensing(LicenseService licenseService, Path licenseDir) {
+    this.licenseService = requireNonNull(licenseService);
+    this.licenseFile = requireNonNull(licenseDir).resolve(LICENSE_FILE_NAME);
+  }
+
+  public Path getLicenseFile() {
+    return licenseFile;
+  }
+
+  public boolean isInstalled() {
+    licenseLock.readLock().lock();
+    try {
+      return licenseFile.toFile().exists() && Files.isRegularFile(licenseFile) && Files.isReadable(licenseFile);
+    } finally {
+      licenseLock.readLock().unlock();
+    }
+  }
+
+  public Optional<String> read() throws UncheckedIOException, IllegalStateException {
+    licenseLock.readLock().lock();
+    try {
+      return licenseFile.toFile().exists() ?
+          Optional.of(new String(Files.readAllBytes(licenseFile), StandardCharsets.UTF_8)) :
+          Optional.empty();
+    } catch (IOException e) {
+      throw new UncheckedIOException(e);
+    } finally {
+      licenseLock.readLock().unlock();
+    }
+  }
+
+  public Optional<License> parse() {
+    licenseLock.readLock().lock();
+    try {
+      return isInstalled() ? Optional.of(licenseService.parse(licenseFile)) : Optional.empty();
+    } finally {
+      licenseLock.readLock().unlock();
+    }
+  }
+
+  /**
+   * @return false is license not installed, true if valid, {@code org.terracotta.dynamic_config.server.api.LicenseService#validate} will throw if invalid
+   */
+  public boolean validate(Cluster cluster) throws InvalidLicenseException {
+    licenseLock.readLock().lock();
+    try {
+      if (!isInstalled()) {
+        LOGGER.warn("Unable to validate cluster against license: license not installed: {}", cluster.toShapeString());
+        return false;
+      }
+      licenseService.validate(licenseFile, cluster);
+      LOGGER.debug("License is valid for cluster: {}", cluster.toShapeString());
+      return true;
+    } finally {
+      licenseLock.readLock().unlock();
+    }
+  }
+
+  public void install(String licenseContent, Cluster cluster) {
+    licenseLock.writeLock().lock();
+    try {
+      if (licenseContent != null) {
+        try {
+          final Path tempFile = Files.createTempFile("terracotta-license-", ".xml");
+          try {
+            Files.write(tempFile, licenseContent.getBytes(StandardCharsets.UTF_8));
+            licenseService.validate(tempFile, cluster);
+            LOGGER.info("License validated");
+            LOGGER.debug("Moving license file: {} to: {}", tempFile, licenseFile);
+            org.terracotta.utilities.io.Files.relocate(tempFile, licenseFile, StandardCopyOption.REPLACE_EXISTING);
+            LOGGER.info("License installed");
+          } finally {
+            try {
+              org.terracotta.utilities.io.Files.deleteIfExists(tempFile);
+            } catch (IOException ignored) {
+            }
+          }
+        } catch (IOException e) {
+          throw new UncheckedIOException(e);
+        }
+        LOGGER.info("License installation successful");
+      } else if (isInstalled()) {
+        try {
+          org.terracotta.utilities.io.Files.deleteIfExists(licenseFile);
+        } catch (IOException e) {
+          throw new UncheckedIOException("Error deleting existing license " + licenseFile + ": " + e.getMessage(), e);
+        }
+        LOGGER.info("License removal successful");
+      } else {
+        LOGGER.info("No license installed");
+      }
+    } finally {
+      licenseLock.writeLock().unlock();
+    }
+  }
+
+}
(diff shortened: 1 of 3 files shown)


dynamic-config/server/configuration-provider/src/main/java/org/terracotta/dynamic_config/server/configuration/service/Topologies.java
```diff
@@ -0,0 +1,193 @@
...
+   * - We cannot use the node name and stripe ID only, since the stripe ID can have changed in the new cluster with the attach/detach commands
+   * - Name could change following a set command when un-configured
+   * <p>
+   * So we try to find the best match we can...
+   */
+  private Node findMe(Cluster updatedCluster) {
+    final Node me = getUpcomingNodeContext().getNode();
+    for (Node node : updatedCluster.getNodes()) {
+      if (node.getUID().equals(me.getUID())
+          || node.getInternalAddress().equals(me.getInternalAddress())
```
chrisdennis, line 138: Why do we match on anything beyond UID?

dynamic-config/server/configuration-provider/src/main/java/org/terracotta/dynamic_config/server/configuration/service/Topologies.java
```diff
@@ -0,0 +1,193 @@
...
+ * @author Mathieu Carbou
+ */
+class Topologies {
+
+  private static final Logger LOGGER = LoggerFactory.getLogger(Topologies.class);
+
+  // guard access to the topologies
+  private final ReadWriteLock topologyLock = new ReentrantReadWriteLock();
+
+  private volatile NodeContext upcomingNodeContext;
```
chrisdennis, line 49: If these are guarded by the lock why are they volatile? Is Findbugs trying to tell you something?

chrisdennis (changes requested):

dynamic-config/server/configuration-provider/src/main/java/org/terracotta/dynamic_config/server/configuration/service/Licensing.java
```diff
@@ -0,0 +1,150 @@
...
+            try {
+              org.terracotta.utilities.io.Files.deleteIfExists(tempFile);
+            } catch (IOException ignored) {
+            }
+          }
+        }
+        LOGGER.info("License installation successful");
+
+      } else {
+        LOGGER.info("No license installed");
```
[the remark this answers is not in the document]
mathieucarbou: Yes, we can uninstall the license (remove it) if parameter is null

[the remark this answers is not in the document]
mathieucarbou: This line was there to confirm that there is no license installed. I will move it down a little.

dynamic-config/server/configuration-provider/src/main/java/org/terracotta/dynamic_config/server/configuration/service/Licensing.java
```diff
@@ -0,0 +1,150 @@
...
+    licenseLock.readLock().lock();
+    try {
+      return licenseFile.toFile().exists() && Files.isRegularFile(licenseFile) && Files.isReadable(licenseFile);
+    } finally {
+      licenseLock.readLock().unlock();
+    }
+  }
+
+  public Optional<String> read() throws UncheckedIOException, IllegalStateException {
+    if (licenseLock.readLock().tryLock()) {
```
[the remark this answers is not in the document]
mathieucarbou: I have converted it to a normal read lock/unlock, it's totally fine and better. Reading the license is done when doing a DS operation to setup the new stripe. In theory, there couldn't be any other license installation process at the same time since DC is supposed to be locked.

dynamic-config/server/configuration-provider/src/main/java/org/terracotta/dynamic_config/server/configuration/service/Topologies.java
```diff
@@ -0,0 +1,193 @@
...
+      c.accept(upcomingNodeContext);
+    } finally {
+      topologyLock.readLock().unlock();
+    }
+  }
+
+  public NodeContext getUpcomingNodeContext() {
+    topologyLock.readLock().lock();
+    try {
+      return upcomingNodeContext.clone();
```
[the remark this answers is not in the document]
mathieucarbou: Yes: the DC model implements that to return a copy of the topology for the callers because it is not immutable.

dynamic-config/server/configuration-provider/src/main/java/org/terracotta/dynamic_config/server/configuration/service/Topologies.java
```diff
@@ -0,0 +1,193 @@
...
+      } else {
+        // We have updated the topology and I am not part anymore of the cluster
+        // So we just reset the cluster object so that this node is alone
+        Node oldMe = upcomingNodeContext.getNode();
+        LOGGER.info("Node {} ({}) removed from pending topology: {}", oldMe.getName(), oldMe.getUID(), updatedCluster.toShapeString());
+        this.upcomingNodeContext = this.upcomingNodeContext.withOnlyNode(oldMe);
+      }
+
+      // When node is not yet activated, runtimeNodeContext == upcomingNodeContext
+      this.runtimeNodeContext = upcomingNodeContext;
```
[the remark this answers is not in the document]
mathieucarbou: That line is OK.

But thanks because there was a bug: `return upcomingNodeContext;` shouldn't have been with the if.

This the method called when installing the pushed topology when nodes are being configured (import) or activated. So at this time the runtime topology is equal to the upcoming topology. In case of activation, there will be a restart after Nomad gets acttivated.

dynamic-config/server/configuration-provider/src/main/java/org/terracotta/dynamic_config/server/configuration/service/Topologies.java
mathieucarbou: UIDs are internal settings hidden to the user except when he exports a config from a cluster. The user never has to set these. There are some commands allowing a user to push a new topology anmd we need to detected where is our node inside it.

dynamic-config/server/configuration-provider/src/main/java/org/terracotta/dynamic_config/server/configuration/service/Topologies.java
mathieucarbou: No, it's me that didn't remove it because the fields were moved from the old class to the new one without being updated.

dynamic-config/server/configuration-provider/src/main/java/org/terracotta/dynamic_config/server/configuration/service/Licensing.java
```diff
@@ -0,0 +1,150 @@
...
+    }
+  }
+
+  public void install(String licenseContent, Cluster cluster) {
+    licenseLock.writeLock().lock();
+    try {
+      if (licenseContent != null) {
+        Path tempFile = null;
+        try {
+          tempFile = Files.createTempFile("terracotta-license-", ".xml");
```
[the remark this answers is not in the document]
mathieucarbou: done

dynamic-config/server/configuration-provider/src/main/java/org/terracotta/dynamic_config/server/configuration/service/Topologies.java
```diff
@@ -0,0 +1,192 @@
...
+  public boolean areSame() {
+    topologyLock.readLock().lock();
+    try {
+      return runtimeNodeContext.equals(upcomingNodeContext);
+    } finally {
+      topologyLock.readLock().unlock();
+    }
+  }
+
+  @SuppressWarnings("NonAtomicOperationOnVolatileField")
```
chrisdennis, line 92: This should be unnecessary now.

dynamic-config/server/configuration-provider/src/main/java/org/terracotta/dynamic_config/server/configuration/service/Topologies.java
chrisdennis: Okay... but is the effect we want here more like:
1. If there is a UID match... take that.
2. If there isn't a UID match look for a **unique** name match.
3. If there isn't a UID or name match look for a **unique** internal address match.

dynamic-config/server/configuration-provider/src/main/java/org/terracotta/dynamic_config/server/configuration/service/Topologies.java
```diff
@@ -0,0 +1,192 @@
...
+      }
+    }
+    return null;
+  }
+
+  public boolean containsMe(Cluster maybeUpdatedCluster) {
+    return findMe(maybeUpdatedCluster) != null;
+  }
+
+  @SuppressWarnings("NonAtomicOperationOnVolatileField")
```
chrisdennis, line 149: No longer needed right?

dynamic-config/server/configuration-provider/src/main/java/org/terracotta/dynamic_config/server/configuration/service/Licensing.java
```diff
@@ -0,0 +1,146 @@
...
+        }
+        LOGGER.info("License installation successful");
+      } else {
+        try {
+          org.terracotta.utilities.io.Files.deleteIfExists(licenseFile);
+        } catch (IOException e) {
+          throw new UncheckedIOException("Error deleting existing license " + licenseFile + ": " + e.getMessage(), e);
+        }
+        // confirm that no license is installed when null is passed to uninstall it or do nothing
+        LOGGER.info("No license installed");
```
chrisdennis, line 139: Maybe I'm laboring the point... what was more getting at is that the voice/tense is wrong.

"License installation successful" vs "No license installed".

The former reports the success of an operation... the later asserts a state.

dynamic-config/server/configuration-provider/src/main/java/org/terracotta/dynamic_config/server/configuration/service/Licensing.java
```diff
@@ -0,0 +1,146 @@
...
+  }
+
+  public Path getLicenseFile() {
+    return licenseFile;
+  }
+
+  public boolean isInstalled() {
+    licenseLock.readLock().lock();
+    try {
+      return licenseFile.toFile().exists() && Files.isRegularFile(licenseFile) && Files.isReadable(licenseFile);
```
chrisdennis, line 63: `exists()` is redundant here.

dynamic-config/server/configuration-provider/src/main/java/org/terracotta/dynamic_config/server/configuration/service/Topologies.java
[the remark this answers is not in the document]
chrisdennis: I assume we have no immutable implementation we can convert to?

chrisdennis (approved):
Approved. Feel free to act on my remaining comments (or not).

dynamic-config/server/configuration-provider/src/main/java/org/terracotta/dynamic_config/server/configuration/service/Topologies.java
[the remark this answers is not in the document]
mathieucarbou: no...

dynamic-config/server/configuration-provider/src/main/java/org/terracotta/dynamic_config/server/configuration/service/Topologies.java
mathieucarbou: no because names could be auto-generated but internal address is more reliable since it is less likely to change and is auto-generated.
I.e. in configuration mode, when a node starts, UID and name are generated. If a user pushes a topology to this node, the more reliable information there to match is the internal address.

dynamic-config/server/configuration-provider/src/main/java/org/terracotta/dynamic_config/server/configuration/service/Topologies.java
mathieucarbou: thanks!

dynamic-config/server/configuration-provider/src/main/java/org/terracotta/dynamic_config/server/configuration/service/Topologies.java
mathieucarbou: thanks!

dynamic-config/server/configuration-provider/src/main/java/org/terracotta/dynamic_config/server/configuration/service/Licensing.java
mathieucarbou: Yes, I completely agree... But updating that requires to go over some test (changes current behavior).
So I have updated the PR to include this update.
- if parameter not null => `"License installation successful"`
- if license installed and parameter is null => `"License removal successful"`
- if license not installed and parameter is null => `"No license installed"`

PR #1038, opened 2021-08-03
config-tool unset -setting license-file

mathieucarbou:
There is no platform (OSS) system tests for that, obviously. They are in EE and this has been tested in EE.

### DIFF
--- a/dynamic-config/model/src/main/java/org/terracotta/dynamic_config/api/model/Setting.java
+++ b/dynamic-config/model/src/main/java/org/terracotta/dynamic_config/api/model/Setting.java
@@ -613,7 +613,7 @@ public enum Setting {
       o -> Optional.empty(),
       noop(),
       singletonList(
-          when(CONFIGURING, ACTIVATED).allow(SET).atLevel(CLUSTER)
+          when(CONFIGURING, ACTIVATED).allow(SET, UNSET).atLevel(CLUSTER)
       ),
       EnumSet.noneOf(Requirement.class),
       emptyList(),
--- a/dynamic-config/cli/dynamic-config-cli-api/src/main/java/org/terracotta/dynamic_config/cli/api/command/RemoteAction.java
+++ b/dynamic-config/cli/dynamic-config-cli-api/src/main/java/org/terracotta/dynamic_config/cli/api/command/RemoteAction.java
@@ -593,9 +593,9 @@ public abstract class RemoteAction implements Runnable {
 
   protected final void upgradeLicense(Collection<Endpoint> expectedOnlineNodes, Path licenseFile) {
     LOGGER.trace("upgradeLicense({}, {})", expectedOnlineNodes, licenseFile);
-    String xml;
+    final String xml;
     try {
-      xml = new String(Files.readAllBytes(licenseFile), StandardCharsets.UTF_8);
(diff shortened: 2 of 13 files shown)


dynamic-config/model/src/main/java/org/terracotta/dynamic_config/api/model/Setting.java
```diff
@@ -613,7 +613,7 @@ public boolean vetoRuntimeChange(NodeContext currentNode, Configuration configur
       o -> Optional.empty(),
       noop(),
       singletonList(
-          when(CONFIGURING, ACTIVATED).allow(SET).atLevel(CLUSTER)
+          when(CONFIGURING, ACTIVATED).allow(SET, UNSET).atLevel(CLUSTER)
```
tmesic99, line 616: don't forget to update the rules table in ConfigTool.adoc

tmesic99:
Is it possible that after unsetting the license, one or more nodes in the cluster may not successfully start?  I'm thinking of a scenario where the previously licensed cluster had been running with one or more licensed features, which after having been unlicensed, will prevent the node(s) from starting.  I'm not sure if any such licensed features exist that could trigger such a scenario.  If it is possible that such a scenario could happen, then should we identify this _before_ executing the unset command to warn the user that their cluster will no longer function as intended (and that nodes may not even be cable of starting).  Again, not sure if this scenario is even possible.

mathieucarbou:
> Not sure if I totally understand what the possible ramifications are for unlicensing a previously activated cluster. For example, is it possible that after unsetting the license, one or more nodes in the cluster may not successfully start? I'm thinking of a scenario where the previously licensed cluster had been running with one or more licensed features, which after having been unlicensed, will prevent the node(s) from starting. I'm not sure if any such licensed features exist that could trigger such a scenario. If it is possible that such a scenario could happen, then should we identify this _before_ executing the unset command to warn the user that their cluster will no longer function as intended (and that nodes may not even be cable of starting). Again, not sure if this scenario is even possible. Also, does unsetting a license automatically deactivate the cluster? If so, could any of these nodes, now having been reverted to diagnostic mode, not be able to startup?

That's ok :-) There is always a grace period of 24h before server stops itself if no license installed. Starting an EE server without a license works fine and it can be installed after.

chrisdennis:
Right - but there isn't a grace period for the removed capabilities is there?

mathieucarbou:
> Right - but there isn't a grace period for the removed capabilities is there?

Removing a license deactivate the license checks (in DC).
I will verify for the EE code but the code has to be dynamic like that too because it supports not having a license and adding one in real time plus upgrading.

But I will verify the code from a client point of view yes to make sure that when the license service is called from en client entity it won't fail and the grace period will be triggered dynamically.

mathieucarbou:
To clarify:

- if no license is installed: the grade period is triggered
- No more Ehcache / Store entities can be created
- Existing ones can still be used

Also:

`config-tool unset -setting license-file` is meant to be the opposite action of `config-tool set -setting license-file`. 
This is a way for the user to rollback what he did in case he finally wants to remove the file he placed on the nodes by mistake, instead of having to SSH on all nodes.

`config-tool unset -setting license-file` isn't meant to be used to switch / upgrade license: we just need to set a new one.

`config-tool unset -setting license-file` has more value when configuring a set of nodes or if a user wants to create a sort of "vanilla" activated cluster.

GaryWKeim (approved):

PR #1040, opened 2021-08-23
Use a more robust regex to parse the server build-id

chrisdennis:


mathieucarbou (approved):
There is Also another place to change on client side for backward compatibility

tmesic99:
> 
> 
> There is Also another place to change on client side for backward compatibility

Yes.  I believe the other location is in `DiagnosticServiceImpl.java`

chrisdennis:
Add the same code to the other location.

PR #1043, opened 2021-09-23
Passive redirection should consider the topology configuration to redirect to the right active endpoint

mathieucarbou:
Implementation of a `NetworkTranslator` for 10.x that will correctly pick the right endpoint for the active node when a passive sends a redirection back to a client.

A terracotta cluster is either meant to be used with the internal endpoints of the nodes, or the public configured endpoints if configured, but not both at the same time.

If public endpoints are not configured, all clients need to use the internal endpoints, and they must be reachable from all clients.

If public endpoints are configured, all clients need to use these public endpoints, not internal ones. All the clients will have to be able to reach the nodes by using these public endpoints.

Technically, it means that if all clients are in the same network as all nodes, public endpoints are not necessary. If clients are in another network and internal endpoints cannot be reached, then public endpoints can be configured to reach them. The user will be responsible to configure NAT / DNS / etc. 

If some clients need to be in the same network as nodes (so clients being both outside and inside), then the user will have to make sure that the public endpoints also resolve to the node internal endpoints, even within the same network.

The `NetworkTranslator` was always returning the "internal" endpoint of the active server, which of course is not always accessible from clients outside an isolated "terracotta network", which could prevent them from connecting to the cluster.

This fix works this way:
- If a public hostname/port is configured for the active node, it will be returned
- otherwise, the normal hostname/port is returned

CC @tglaeser

### DIFF
--- a/dynamic-config/server/services/pom.xml
+++ b/dynamic-config/server/services/pom.xml
@@ -62,6 +62,11 @@
       <scope>provided</scope>
     </dependency>
     <dependency>
+      <groupId>org.terracotta.internal</groupId>
+      <artifactId>server-spi</artifactId>
+      <scope>provided</scope>
+    </dependency>
+    <dependency>
       <groupId>org.terracotta</groupId>
       <artifactId>standard-cluster-services</artifactId>
       <scope>provided</scope>
--- a/dynamic-config/server/services/src/main/java/org/terracotta/dynamic_config/server/service/DynamicConfigNetworkTranslator.java
+++ b/dynamic-config/server/services/src/main/java/org/terracotta/dynamic_config/server/service/DynamicConfigNetworkTranslator.java
@@ -1,0 +1,91 @@
+/*
+ * Copyright Terracotta, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.terracotta.dynamic_config.server.service;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.terracotta.dynamic_config.api.model.Cluster;
+import org.terracotta.dynamic_config.api.model.Node;
+import org.terracotta.dynamic_config.api.service.TopologyService;
+import org.terracotta.inet.InetSocketAddressConverter;
+
+import java.net.InetSocketAddress;
+import java.util.Optional;
+
+/**
+ * This service implements a network translator to support TC deployment behind a NAT-ed network for example.
+ * <p>
+ * If TC nodes are deployed in a separate network, recommendation is to configure public hostname/port on the
+ * nodes and have all clients reach the nodes by using only public endpoints. The user will be responsible to
+ * setup his networks (NAT, DNS ,etc) accordingly so that public endpoints can resolve to the nodes whether the
+ * clients are "inside" or "outside" of the TC node network.
+ * <p>
+ * This is not possible to correctly support a mix of clients connecting to the cluster, some with public
+ * endpoints and some with internal endpoints, because the NetworkTranslator has no way to know in a reliable
+ * way which endpoint was used by the user in the URI to initiate the connection.
+ *
+ * @author Mathieu Carbou
+ */
+class DynamicConfigNetworkTranslator implements com.tc.spi.NetworkTranslator {
+
+  private static final Logger LOGGER = LoggerFactory.getLogger(DynamicConfigNetworkTranslator.class);
+
+  private final TopologyService topologyService;
+
+  public DynamicConfigNetworkTranslator(TopologyService topologyService) {
+    this.topologyService = topologyService;
+  }
+
+  /**
+   * This implementation will return the right endpoint to use when a passive node is redirecting a client to an active node.
+   * <p>
+   * If the cluster is configured with public addresses, the public endpoint of the active server will be used.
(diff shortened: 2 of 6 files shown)


dynamic-config/model/src/main/java/org/terracotta/dynamic_config/api/model/Cluster.java
```diff
@@ -514,6 +514,13 @@ public UID newUID(Random random) {
     return uuid;
   }
 
+  public Optional<Node> findNode(InetSocketAddress addr) {
+    return stripes.stream()
+        .map(stripe -> stripe.findNode(addr).orElse(null))
```
chrisdennis, line 519: Would this be better expressed as:
```java
stripes.stream().map(stripe -> stripe.findNode(addr)).filter(Optional::isPresent).map(Optional::get).findFirst()
```

dynamic-config/server/services/pom.xml
```diff
@@ -61,6 +61,11 @@
       <artifactId>management</artifactId>
       <scope>provided</scope>
     </dependency>
+    <dependency>
+      <groupId>org.terracotta.internal</groupId>
```
chrisdennis, line 65: I know Myron was pushing to have less dependency on internal artifacts coming out of core. Is this something we could eliminate?

dynamic-config/server/services/src/main/java/org/terracotta/dynamic_config/server/service/DynamicConfigNetworkTranslator.java
```diff
@@ -0,0 +1,92 @@
...
+   * @return the endpoint to use to connect to the active server
+   */
+  @Override
+  public String redirectTo(InetSocketAddress initiator, String serverHostPort) {
+    final Cluster cluster = topologyService.getRuntimeNodeContext().getCluster();
+    InetSocketAddress proposedRedirect;
+    try {
+      proposedRedirect = InetSocketAddressConverter.getInetSocketAddress(serverHostPort);
+    } catch (IllegalArgumentException e) {
+      // there is a bug in core. Core that does support Ipv6 addresses correctly.
```
chrisdennis, line 73: You mean core *doesn't* support IPv6 correctly? Is there an issue to reference so that:
* this workaround can get removed when fixed
* the correctness of this code can be judged.

dynamic-config/server/services/src/main/java/org/terracotta/dynamic_config/server/service/DynamicConfigServiceProvider.java
```diff
@@ -117,6 +118,9 @@ public boolean initialize(ServiceProviderConfiguration configuration, PlatformCo
 
   @Override
   public <T> T getService(long consumerID, ServiceConfiguration<T> configuration) {
+    if (configuration.getServiceType() == NetworkTranslator.class) {
```
chrisdennis, line 121: Why is this service the only service that needs special-casing?

chrisdennis (changes requested):

dynamic-config/model/src/main/java/org/terracotta/dynamic_config/api/model/Cluster.java
mathieucarbou: it's the same :P mine is just 1 line less and 9 characters less to type.

dynamic-config/server/services/src/main/java/org/terracotta/dynamic_config/server/service/DynamicConfigNetworkTranslator.java
mathieucarbou: Updated the comment (to match the commit comment).
Core is using such hacks also at some places too. This is due to the fact that core does not provide a way to represent a hostname/port combination (what we talked about) so everything goes through string, but the string representation of an `ip:port` is not the same for IPv4 and v6 and also changes across JVMs.

dynamic-config/server/services/src/main/java/org/terracotta/dynamic_config/server/service/DynamicConfigServiceProvider.java
mathieucarbou: Because `findService()` looks at the DC services (provided at bootstrap time) that are exposed inside `PlatformConfiguration` and made available to services when they get initialized (and entities through this SP).
`NetworkTranslator` is a core interface looked up by core - aka what is done for SSL buffers.

dynamic-config/model/src/main/java/org/terracotta/dynamic_config/api/model/Stripe.java
```diff
@@ -200,6 +200,10 @@ public Scope getScope() {
     return Scope.STRIPE;
   }
 
+  public Optional<Node> findNode(InetSocketAddress addr) {
+    return nodes.stream().filter(node -> node.hasAddress(addr)).findFirst();
```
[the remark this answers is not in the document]
mathieucarbou: What do you mean ? This is the configured topology by the user: this method tries to find the node matching the host/port as configured by the user.

Why would I check the bind address ?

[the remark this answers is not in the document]
myronkscott: the redirect address is the address that the server is available to clients on.  If the bind address is a distinct interface or address, the server publishes that address.  If the bind address is wildcard, the combination of the hostname and port is published.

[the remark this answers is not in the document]
myronkscott: yes

mathieucarbou:
@myronkscott @chrisdennis : so after discussions I have updated this PR to account for the bind address when not a wildcard one.

But we also settled on not merging this PR yet. but instead group it with other fixes / discussions about hostname / bind addr

dynamic-config/model/src/main/java/org/terracotta/dynamic_config/api/model/Cluster.java
mathieucarbou: > Not quite the same... yours assumes that `findNode(...)` never returns null.
Same for yours: `.filter(Optional::isPresent)` would throw a NPE

PR #1047, opened 2021-10-04
Update parent pom

mathieucarbou:


pom.xml
```diff
@@ -22,7 +22,7 @@
   <parent>
     <groupId>org.terracotta</groupId>
     <artifactId>terracotta-parent</artifactId>
-    <version>5.18</version>
+    <version>5.19</version>
```
akomakom, line 25: Let's use 5.20 since we're updating

akomakom (approved):

PR #1048, opened 2021-10-05
Compat' with future Angela version

mathieucarbou:


mathieucarbou:
@aurbroszniowski : PTI, build is passing locally with angela snapshot

PR #1049, opened 2021-10-08
Periodic logging about resource usage

mobasherul:


chrisdennis (changes requested):
It seems to me that what is being logged here is a generally valuable metric (that might want to be exposed in any M&M view). It also seems like there are many metrics that might be valuable to customers if periodically logged. With those statements in mind I'm thinking it makes more sense to frame this feature as the periodic logging of some monitorable metrics. In that sense the engine that drives this should probably be integrated with the M&M code. That would also prevent us from spawning two extra threads where each thread just logs a single line every 5 minutes.

data-root-resource/src/main/java/org/terracotta/config/data_roots/DataDirsConfigImpl.java
```diff
@@ -53,11 +59,17 @@
...
   private final String platformRootIdentifier;
   private final ConcurrentMap<String, DataDirs> serverToDataRoots = new ConcurrentHashMap<>();
   private final IParameterSubstitutor parameterSubstitutor;
   private final PathResolver pathResolver;
   private final Collection<EntityManagementRegistry> registries = new CopyOnWriteArrayList<>();
+  private final ScheduledExecutorService loggingExecutor = Executors.newSingleThreadScheduledExecutor(r -> {
+    Thread t = new Thread(r, "disk resources usage logger");
+    t.setDaemon(true);
+    return t;
+  });
```
mathieucarbou, line 72: This is not the right place to put a thread pool since this is only a configuration holder class.
Any action / computation should much more happen through a service.
I.e. in M&M we have a service tht has a singleton thread pool with one thread to handle stat collection. Services have access to configs so it would be possible to re-use the existing M&M services to do this sort of logging.

mobasherul:
> It seems to me that what is being logged here is a generally valuable metric (that might want to be exposed in any M&M view). It also seems like there are many metrics that might be valuable to customers if periodically logged. With those statements in mind I'm thinking it makes more sense to frame this feature as the periodic logging of some monitorable metrics. In that sense the engine that drives this should probably be integrated with the M&M code. That would also prevent us from spawning two extra threads where each thread just logs a single line every 5 minutes.

M&M code have their thread pools which do statistics collection but only when tmc is connected to tms otherwise M&M code doesn't do anything.

mobasherul:
Closing this in favor of https://github.softwareag.com/Terracotta/terracotta-enterprise/pull/170 as I am creating service provider which will take care of logging

PR #1051, opened 2021-10-14
New utility methods to better query the topology

mathieucarbou:


dynamic-config/model/src/main/java/org/terracotta/dynamic_config/api/model/Cluster.java
```diff
@@ -514,19 +515,26 @@ public UID newUID(Random random) {
     return uuid;
   }
 
+  public Optional<Node> findReachableNode(InetSocketAddress addr) {
```
tmesic99, line 518: Couldn't see where this method is called from. Maybe add a new test for it anyway?

mathieucarbou: Oh, it's for #1043. We had `hasAddress()` before, which had a flawed impl that is fixed with that.

mathieucarbou: This PR is just the exposure of topo information for other project + some fixes in the reported info and how checks are done.

PR #1054, opened 2021-10-18
Add server name folders in newly added Data-Dirs

mobasherul:


data-root-resource/src/main/java/org/terracotta/config/data_roots/DataDirsConfigImpl.java
```diff
@@ -122,6 +122,10 @@ public DataDirs getDataDirectoriesForServer(PlatformConfiguration platformConfig
   @Override
   public void addDataDirectory(String name, String path) {
     addDataDirectory(name, path, false);
+    // For newly added data-dirs as part of dynamic-config.
```
mobasherul, line 125: For newly added data-dirs node name folders was not created so in the mean time after adding new data-dirs if someone using DataDirs called getDataDirectory(<newly-added-data-dir>) it was failing because server folders were absent.

data-root-resource/src/main/java/org/terracotta/config/data_roots/DataDirsWithServerName.java
```diff
@@ -113,6 +113,17 @@ public void close() throws IOException {
...
 
+  @Override
+  public void updateDataDir(String name) {
+    Path resolved = wrapped.getRoot(name).resolve(serverName);
+    try {
+      wrapped.ensureDirectory(resolved);
+      lockDirectory(name, resolved);
+    } catch (IOException e) {
+      throw new DataDirsConfigurationException(e.toString(), e);
+    }
```
mathieucarbou, line 124: The logic is so internal that I am wondering if we should move this code directly in the for loop and remove the `updateDataDir` from the interface ?

mobasherul: ok then I will remove the api from the interface then

mathieucarbou (approved):

PR #1055, opened 2021-10-20
Fix endpoint handling in DC topology

mathieucarbou:


dynamic-config/cli/dynamic-config-cli-api/src/main/java/org/terracotta/dynamic_config/cli/api/command/RemoteAction.java
```diff
@@ -343,10 +343,17 @@ protected final Cluster getRuntimeCluster(InetSocketAddress expectedOnlineNode)
     }
   }
 
+  /**
+   * This method will connect to the node using teh provided address from teh user.
```
tmesic99, line 347: teh -> the

tmesic99 (approved):

PR #1058, opened 2021-11-29
Fix missing "instanceId" routing information for monitoring events and management calls

mathieucarbou:
Management: Allow more routing information to be passed in the client context.

Backward-compatible refactoring in management code to allow a client to add more routing information in its context used to identify pushed statistics and notifications. This will allow a client context to contain the previously missing "instanceId" routing information plus the manager name

### DIFF
--- a/management/cluster-topology/src/main/java/org/terracotta/management/model/cluster/AbstractManageableNode.java
+++ b/management/cluster-topology/src/main/java/org/terracotta/management/model/cluster/AbstractManageableNode.java
@@ -15,6 +15,7 @@
  */
 package org.terracotta.management.model.cluster;
 
+import org.terracotta.management.model.context.Context;
 import org.terracotta.management.model.context.Contextual;
 
 import java.util.Optional;
@@ -44,6 +45,11 @@ public abstract class AbstractManageableNode<P extends Contextual> extends Abstr
 
   public final boolean isManageable() {
     return managementRegistry != null;
+  }
+
+  @Override
+  public Context getContext() {
+    return super.getContext().with(getManagementRegistry().map(ManagementRegistry::getContext).orElse(Context.empty()));
   }
 
   @Override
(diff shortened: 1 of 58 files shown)


management/cluster-topology/src/main/java/org/terracotta/management/model/cluster/AbstractManageableNode.java
```diff
@@ -46,6 +47,11 @@ public final boolean isManageable() {
     return managementRegistry != null;
   }
 
+  @Override
+  public Context getContext() {
+    return super.getContext().with(getManagementRegistry().map(ManagementRegistry::getContext).orElse(Context.empty()));
+  }
```
mathieucarbou, line 53: `Client.getContext()` will now contain:
- clientId
- instanceId
- cacheManagerName / datasetManagerName (== context container key/value)

mathieucarbou: ... and this context will be of course used as the origin for all stats / notifications

management/management-registry/src/test/java/org/terracotta/management/registry/action/MyManagementProvider.java
```diff
@@ -23,7 +23,7 @@
  * @author Mathieu Carbou
  */
 @Named("TheActionProvider")
-@RequiredContext({@Named("cacheManagerName"), @Named("cacheName")})
+@RequiredContext({@Named("instanceId"), @Named("cacheManagerName"), @Named("cacheName")})
```
mathieucarbou, line 26: All client-side management providers have to include the instanceId in their context to be reached (aka the consumerID for server entities)

management/testing/integration-tests/src/test/java/org/terracotta/management/integration/tests/AbstractTest.java
```diff
@@ -291,7 +297,7 @@ protected void triggerClientStatComputation(long interval, TimeUnit unit) throws
         .clientStream()
         .filter(client -> client.getName().equals("pet-clinic"))
         .filter(AbstractManageableNode::isManageable)
-        .map(client -> client.getContext().with("appName", "pet-clinic"))
```
mathieucarbou, line 294: since the client context has more data now than just the clientId, no need when doign a management call to append the contextContainer inside since it will already be in the virtual cluster connection context

GaryWKeim (approved):

PR #1060, opened 2021-12-10
TDB-5488 - fix set/unset license-file when unconfigured

tmesic99:
Fix for defect: https://itrac.eur.ad.sag/browse/TDB-5488.
This problem started after 10.7 fix 3, resulting from a change to when a server's config folders get created on disk (698072a0).  With 10.7 fix 3 and earlier, config folders were created when the node was unconfigured .  But this behavior was changed to only create the config folder structure as part of activation.  This problem revealed itself when setting a license file on an unconfigured node with the server attempting to install the license into a non-existent folder throwing an exception (see TDB-5488).

The solution here holds the supplied 'licenseContent' string in memory, if set when unconfigured, until the node is activated.  Upon activation, the existing licenseContent is then used to create the license on disk.  Or, that in-memory license is superseded if the license is optionally supplied with the activate call.

Corresponding system tests will appear in a subsequent Enterprise PR.

### DIFF
--- a/dynamic-config/server/configuration-provider/src/main/java/org/terracotta/dynamic_config/server/configuration/service/Licensing.java
+++ b/dynamic-config/server/configuration-provider/src/main/java/org/terracotta/dynamic_config/server/configuration/service/Licensing.java
@@ -47,10 +47,13 @@ class Licensing {
 
   private final LicenseService licenseService;
   private final Path licenseFile;
+  private volatile String unconfiguredLicenseContent; // set only when the server is unconfigured (RO mode)
+  private final NomadServerManager nomadServerManager;
 
-  public Licensing(LicenseService licenseService, Path licenseDir) {
+  public Licensing(LicenseService licenseService, NomadServerManager nomadServerManager) {
     this.licenseService = requireNonNull(licenseService);
-    this.licenseFile = requireNonNull(licenseDir).resolve(LICENSE_FILE_NAME);
+    this.nomadServerManager = nomadServerManager;
+    this.licenseFile = requireNonNull(nomadServerManager.getConfigurationManager().getLicensePath()).resolve(LICENSE_FILE_NAME);
   }
 
   public Path getLicenseFile() {
@@ -66,7 +69,19 @@ class Licensing {
     }
   }
 
-  public Optional<String> read() throws UncheckedIOException, IllegalStateException {
+  public Optional<String> getLicenseContent() {
+    return Optional.ofNullable(read().orElseGet(() -> getUnconfiguredLicenseContent().orElse(null)));
+  }
+
+  /**
+   * Only return the unconfiguredLicenseContent if the server is unconfigured (i.e. RO mode).
+   */
+  private Optional<String> getUnconfiguredLicenseContent() {
+    return nomadServerManager.getNomadMode() == NomadMode.RO && unconfiguredLicenseContent != null ?
+        Optional.of(unconfiguredLicenseContent) : Optional.empty();
+  }
+
+  private Optional<String> read() throws UncheckedIOException, IllegalStateException {
     licenseLock.readLock().lock();
     try {
       return licenseFile.toFile().exists() ?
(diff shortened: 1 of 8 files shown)


mathieucarbou (changes requested):
@tmesic99 I am wondering if this is the right way to solve this issue...

In `Setting`:

`when(CONFIGURING, ACTIVATED).allow(SET, UNSET).atLevel(CLUSTER)`

makes no sense because from a UX perspective, an un-configured node is "nothing" so it's completely useless to license it Besides, there is no "folder" yet to write it.

When doing an activation, a consistency check amongst the nodes is done. But the license is not part of this check. Besides, this is a file that the user could remotely change when activated, but also just update locally in config folder.

Licensing needs to be done 24 hours max after activation (or before a cache or store client connects). So there is no rush in doing it early in the process.

So I am wondering if we should just set the validation to:

`when(CONFIGURING).allow(SET, UNSET).atLevel(CLUSTER)`

And then add some validations in the CLI instead, plus on the backend side (to throw if a license install is requested, but the config/license folder does not exist yet).

tmesic99:
> 

@mathieucarbou Yes, I had the exact same thoughts.  I opted for the solution I implemented in this PR because it preserves the behavior that's always been present in dynamic config.  That is, users have always been able to pre-configure their license before activation.  But if you prefer to take the approach of altogether removing the ability to set/unset licenses from CONFIGURED nodes, then that's fine.  Is that how you'd like to proceed?

mathieucarbou:
> > 
> 
> @mathieucarbou Yes, I had the exact same thoughts. I opted for the solution I implemented in this PR because it preserves the behavior that's always been present in dynamic config. That is, users have always been able to pre-configure their license before activation. But if you prefer to take the approach of altogether removing the ability to set/unset licenses from CONFIGURED nodes, then that's fine. Is that how you'd like to proceed?

No, that's right... I have checked the code and saw that the `set` was allowed since a while... :-(

Let's fix it like that...

dynamic-config/server/configuration-provider/src/main/java/org/terracotta/dynamic_config/server/configuration/service/Licensing.java
```diff
@@ -47,6 +47,7 @@
 
   private final LicenseService licenseService;
   private final Path licenseFile;
+  private String licenseContent;
```
mathieucarbou, line 50: field should be `volatile`

dynamic-config/server/configuration-provider/src/main/java/org/terracotta/dynamic_config/server/configuration/service/Licensing.java
```diff
@@ -106,14 +115,27 @@ public boolean validate(Cluster cluster) throws InvalidLicenseException {
...
+  /**
+   * Licenses can be configured (set/unset) when the server is ACTIVE or UNCONFIGURED.
+   * Only when the server is ACTIVE do we actually install the license (i.e. copy it to the config folder), since the
+   * config folder does not yet exist when the server is UNCONFIGURED.
+   * Update the licenseContent member only if a non-null licenseContent is supplied.  This supports the scenario whereby
+   * the license is set when the server is UNCONFIGURED and is not specified during activation (the supplied
+   * licenseContent would be null during the activate call).  In this case, we want to install the license that was
+   * established while the server was UNCONFIGURED.
+   */
+  public void install(String licenseContent, Cluster cluster, boolean isActive) {
```
mathieucarbou, line 127: ~~I would rename `isActive` to `nomadReady` (or something equivalent), because the installation does not require the server to be active, but only folders to be there.~~

Actually, you can remove this boolean and use the `NomadManager.nomadMode()` call (see other comment)

dynamic-config/server/configuration-provider/src/main/java/org/terracotta/dynamic_config/server/configuration/service/Licensing.java
```diff
@@ -106,14 +115,27 @@ public boolean validate(Cluster cluster) throws InvalidLicenseException {
...
     licenseLock.writeLock().lock();
     try {
       if (licenseContent != null) {
+        this.licenseContent = licenseContent;
+      }
```
mathieucarbou, line 132: I would be very careful there and only use this hack when in un-configured mode. Because when the server is activated, we really don't want this class field to be set, otherwise the getter method might return the value in case the user removed the license file from the disk. The value will be kept in memory and the server might still  licensed, when it's not anymore.

dynamic-config/server/configuration-provider/src/main/java/org/terracotta/dynamic_config/server/configuration/service/Licensing.java
mathieucarbou: I would rename it to something more meaningful like: `temporaryLicenseContent` or a better name, wit ha comment saying that this field is ONLY in unconfigured mode

mathieucarbou: Note: you could change the ctor of `Licensing` to have a ref inside of the `NomadServerManager`, to know the state of nomad with NomadServerManager.getNomadState(): RW == nomad system is up & running, we can do nomad changes. RO == unconfigured mode or repair.

`this.licensing = new Licensing(licenseService, nomadServerManager.getConfigurationManager());`

For the licensing stuff, there no notion of activated or not. The only concept to consider is whether the nomad system is RW or RO to be able to write or not the license file.

dynamic-config/server/configuration-provider/src/main/java/org/terracotta/dynamic_config/server/configuration/service/Licensing.java
```diff
@@ -66,7 +67,15 @@ public boolean isInstalled() {
     }
   }
 
-  public Optional<String> read() throws UncheckedIOException, IllegalStateException {
+  public Optional<String> getLicenseContent() {
+    Optional<String> content = read();
+    if (!content.isPresent() && licenseContent != null) {
```
mathieucarbou, line 72: I would check the nomadMode here too.

dynamic-config/server/configuration-provider/src/main/java/org/terracotta/dynamic_config/server/configuration/service/Licensing.java
mathieucarbou: You can use the `NomadManager.nomadMode()` to protect this

mathieucarbou (changes requested):

dynamic-config/server/configuration-provider/src/main/java/org/terracotta/dynamic_config/server/configuration/service/Licensing.java
mathieucarbou: perhaps a method like that would help:

```java
private Optional<String> findTemporarySetLicense() {
    return nomadManager.getNomadMode() == RW && temporaryLicense != null ? Optional.of(TemporaryLicense) : Optional.emopty();
}
```

Then `getLicenseContent()` would become something like:

```java
return Optional.ofNullable(read().orElseGet(() -> findTemporarySetLicense().orElse(null));
```

dynamic-config/server/configuration-provider/src/main/java/org/terracotta/dynamic_config/server/configuration/service/Licensing.java
tmesic99: done

dynamic-config/server/configuration-provider/src/main/java/org/terracotta/dynamic_config/server/configuration/service/Licensing.java
tmesic99: done

dynamic-config/server/configuration-provider/src/main/java/org/terracotta/dynamic_config/server/configuration/service/Licensing.java
tmesic99: done

dynamic-config/server/configuration-provider/src/main/java/org/terracotta/dynamic_config/server/configuration/service/Licensing.java
tmesic99: done

dynamic-config/server/configuration-provider/src/main/java/org/terracotta/dynamic_config/server/configuration/service/Licensing.java
```diff
@@ -106,30 +121,61 @@ public boolean validate(Cluster cluster) throws InvalidLicenseException {
     }
   }
 
+  /**
+   * Licenses can be configured (set/unset) when the server is ACTIVE or UNCONFIGURED.
+   * Since the config folder does not yet exist when the server is UNCONFIGURED, we only 'install' the license
+   * (i.e. copy the specified file to the config folder) when the server is ACTIVE.
+   * If install is called when the server is UNCONFIGURED (RO mode), we simply update the unconfiguredLicenseContent
+   * member with the supplied licenseContent.
+   * If install is called when the server is ACTIVE (RW mode), we use the license that was supplied during activation.
```
mathieucarbou, line 130: _when the server is **configured** (RW mode)_

dynamic-config/server/configuration-provider/src/main/java/org/terracotta/dynamic_config/server/configuration/service/Licensing.java
```diff
@@ -106,30 +121,61 @@ public boolean validate(Cluster cluster) throws InvalidLicenseException {
...
     licenseLock.writeLock().lock();
     try {
-      if (licenseContent != null) {
-        try {
-          final Path tempFile = Files.createTempFile("terracotta-license-", ".xml");
+      if (nomadServerManager.getNomadMode() == NomadMode.RO) {
+        // server is Unconfigured (or in Repair)
+        unconfiguredLicenseContent = licenseContent;
+      } else {
+        // server is Active
```
mathieucarbou, line 141: _server is Active_ => _server is **configured**_

mathieucarbou (approved):

PR #1064, opened 2022-01-28
Fixing NodeName in Attach Command

mobasherul:


dynamic-config/cli/dynamic-config-cli-api/src/main/java/org/terracotta/dynamic_config/cli/api/command/AttachAction.java
```diff
@@ -163,7 +163,7 @@ protected void validate() {
         addedNode.setUID(destinationCluster.newUID());
 
         if (destinationClusterActivated) {
-          NameGenerator.assignFriendlyNodeName(destinationCluster, addedNode);
+          NameGenerator.assignFriendlyNodeName(destinationStripe, addedNode);
         }
```
mathieucarbou, line 167: I thought we needed both the stripe AND cluster reference because the generated name ha to be unique in the destination cluster.

dynamic-config/cli/dynamic-config-cli-api/src/main/java/org/terracotta/dynamic_config/cli/api/command/AttachAction.java
```diff
@@ -112,14 +112,14 @@ protected void validate() {
      update the topology of "cluster A" so that the detached nodes won't be there anymore and can be attached somewhere else.
      "cluster A" could then be activated without impacting the "cluster B" used in the attach command.
      */
+    Stripe destinationStripe = destinationCluster.getStripeByNode(destination.getNodeUID()).get();
```
mathieucarbou, line 115: It would be better to move this line within the if and after the validation ?

dynamic-config/model/src/main/java/org/terracotta/dynamic_config/api/service/NameGenerator.java
```diff
@@ -85,9 +85,8 @@ public static void assignFriendlyStripeNames(Cluster cluster, Random random) {
     cluster.getStripes().forEach(stripe -> assignFriendlyStripeName(cluster, stripe, random));
   }
 
-  private static void assignFriendlyNodeName(Cluster cluster, Node node, Random random) {
+  private static void assignFriendlyNodeName(Stripe stripe, Node node, Random random) {
```
mathieucarbou, line 88: I think this implementation was already having a bug ?

I don't see any code that is verifying that the generated name is not already assigned to a node in the cluster passed as parameter ?

dynamic-config/model/src/main/java/org/terracotta/dynamic_config/api/service/NameGenerator.java
```diff
@@ -85,9 +85,8 @@ public static void assignFriendlyStripeNames(Cluster cluster, Random random) {
...
     if (nameCanBeSet(node)) {
-      Stripe stripe = cluster.getStripeByNode(node.getUID()).orElseThrow(IllegalArgumentException::new);
       List<String> used = stripe.getNodes().stream().map(Node::getName).collect(toList());
```
mathieucarbou, line 90: Here, used should be cluster.getNodes() I think ?

mobasherul: It should be stripe.GetNodes() as we are are attaching a node to stripe and the name should be unique within the stripe.

dynamic-config/cli/dynamic-config-cli-api/src/main/java/org/terracotta/dynamic_config/cli/api/command/AttachAction.java
mobasherul: can't be moved inside if block as I am using this in separate if block

dynamic-config/model/src/main/java/org/terracotta/dynamic_config/api/service/NameGenerator.java
mobasherul: pickRandomNodeName() was trying to assign a name which is not duplicated

dynamic-config/cli/dynamic-config-cli-api/src/main/java/org/terracotta/dynamic_config/cli/api/command/AttachAction.java
mobasherul: correct

dynamic-config/cli/dynamic-config-cli-api/src/main/java/org/terracotta/dynamic_config/cli/api/command/AttachAction.java
mathieucarbou: oh my bad! there is a validation already in the super class so it will work!

mathieucarbou (approved):

PR #1073, opened 2022-03-18
[main] Move to external slaves

chrisdennis:


akomakom (approved):

mathieucarbou:
This PR should not have been merged without fixing the tests first because A LOT of tests in this project can only be run on 1.11 with the `-Djdk.net.hosts.file` flag, and sadly the infra is different since the host file is not generated accordingly.

Will try to fix in https://github.com/Terracotta-OSS/terracotta-platform/pull/1074.

chrisdennis:
Sorry, the internal runners are now turned off. If this wasn't merged there would be no testing at all. The "passing threshold" for this PR was not "the tests pass", but instead: "the tests run".

PR #1075, opened 2022-03-21
relieve stress on inline servers

myronkscott:
new core uses direct memory for comms.  let's turn that back off for inline tests that churn servers.  I may be causing problems with direct memory GC.

### DIFF
--- a/dynamic-config/testing/support/src/main/java/org/terracotta/dynamic_config/test_support/DynamicConfigIT.java
+++ b/dynamic-config/testing/support/src/main/java/org/terracotta/dynamic_config/test_support/DynamicConfigIT.java
@@ -137,6 +137,7 @@ public class DynamicConfigIT {
           System.setProperty("com.tc.server.entity.processor.threads", "4");
           System.setProperty("com.tc.l2.tccom.workerthreads", "4");
           System.setProperty("com.tc.l2.seda.stage.stall.warning", "1000");
+          System.setProperty("com.tc.tc.messages.packup.enabled", "false");
         }
       });
 


dynamic-config/testing/support/src/main/java/org/terracotta/dynamic_config/test_support/DynamicConfigIT.java
```diff
@@ -137,6 +137,7 @@ protected void before(Description description) {
           System.setProperty("com.tc.server.entity.processor.threads", "4");
           System.setProperty("com.tc.l2.tccom.workerthreads", "4");
           System.setProperty("com.tc.l2.seda.stage.stall.warning", "1000");
+          System.setProperty("com.tc.tc.messages.packup.enabled", "false");
```
mathieucarbou, line 140: @myronkscott : why all these system properties are not in Angela when using inline mode ?

myronkscott: I don't think these properties should be used for all tests.  The dynamic config tests have a few characteristics for which these properties make sense.  Generally, the create many servers, they do a few short actions, they restart servers often usually the tasks are run on management key and they run in constrained environments.  For these reasons, these settings make sense for DynamicConfigIT

mathieucarbou: These properties are set for the whole JVM and I see no cleanup anywhere.... So all the tests of the jvm see them.
Since this is a super class, then this enforces anyone using this super class to run the tests inside a spawned jvm.
The right way would have been to add a `Properties` object in the angela configuration of TSA, which would be then passed in the ctor of `TCServerMain`

myronkscott: Its not pretty but it is producing the desired affect.  I'm just adding a property here not redesigning how tc-properties work.  TC properties usually come from the config.

mathieucarbou: I agree, but like angela, the config, here, of the jvm, is the maven pom... 
I would find it cleaner to move them in the maven pom, which is the place to configure the test jvm... And It's also cleaner to set them in the pom, where they can be configured per project / module, than hard-coded in a super class in an upstream project.

myronkscott: ok.  but you would have to duplicate in EE.

myronkscott: i'd merge this for now and create a new issue for a more proper cleanup.

mathieucarbou (approved):

PR #1074, opened 2022-03-21
Trying to make all jobs pass

mathieucarbou:

PR #1076, opened 2022-03-22
install custom tc properties into kit

myronkscott:


dynamic-config/testing/support/src/main/java/org/terracotta/dynamic_config/test_support/DynamicConfigIT.java
```diff
@@ -227,6 +218,30 @@ protected void before(Description description) {
...
+                props.setProperty("l1.healthcheck.l2.socketConnect", "false");
+                props.setProperty("l2.healthcheck.l2.socketConnect", "false");
+                props.setProperty("l2.election.timeout", "10");
+                ByteArrayOutputStream bos = new ByteArrayOutputStream();
+                props.store(bos, "custom tc properties");
+                bos.close();
+                folder.upload("tc.properties", new ByteArrayInputStream(bos.toByteArray()));
+              }
+            } catch (IOException exp) {
+              LOGGER.warn("unable to upload logback configuration", exp);
```
mathieucarbou, line 243: `tc.properties` perhaps ;-)

mathieucarbou (changes requested):
@myronkscott : the kit location needs to be read-only: It is a common location used for several concurrent tests / tests JVM.
Is there a way to put that in the maven pom (same behavior as before), or in the angela working Dir ?

mathieucarbou:
> tc.properties

Going with a `tc.properties` is OK, but only if the `tc.properties` file is created when the kit is created (which is I thunk in a maven task before the integration tests run. Same in gradle projects). There are some tals already there to setup a kit.

dynamic-config/testing/support/src/main/java/org/terracotta/dynamic_config/test_support/DynamicConfigIT.java
myronkscott: why do you not want this for all dynamic config tests?  Seems pointless to rewrite this file for every test execution.  That is why it is checking to see if the file already exists.

myronkscott: i moved it to a script but that doesn't seem very angela friendly.  isn't angela supposed to support remoting?

mathieucarbou: 1. remoting by making an agent download a kit
2. local, with or without ignite, downloading a kit at a common place
3. local, with or without ignite, downloading a kit at a common place, but copying the kit for each test (not on by default - this option should not be used)

Whatever the option, there is always a step where the kit is "prepared" before being accessed.

To make this preparation common between 1, 2 and 3, this is, like I said, by making this system properties common to the test jvm and to the angela.jvmm like I am doing in this PR: https://github.com/Terracotta-OSS/terracotta-platform/pull/1074/files#diff-1dec5e554ff5570ca9dcfd4d059bbeac8f87016fc10ceaf4adca3a11c8658631R33

The `angela.java.opts` is used to spawn processes with the wanted system proeprties. Even remotely (it is a holder for system proeprties we want to propagate).

So `angela.java.opts` **IS** this common place where to put system props, but they also need to be there in the test JVM **only if** using inline server... If not using inline server, `angela.java.opts` is used to start processes.

myronkscott: i moved it to a script like fake-hosts

dynamic-config/testing/system-tests/pom.xml
```diff
@@ -145,8 +145,24 @@
 
           <scripts>
             <script><![CDATA[
-new File(project.build.directory, "fake-hosts.txt").text = "127.0.0.1 ${InetAddress.getLocalHost().getHostName()} testhostname"
+new File(project.build.directory, "fake-hosts.txt").text = "127.0.0.1 ${InetAddress.getLocalHost().getHostName()} ${InetAddress.getLocalHost().getHostName().split('[.]')[0]} testhostname"
 println(new File(project.build.directory, "fake-hosts.txt").text)
+          ]]></script>
+            <script><![CDATA[
+new File("${project.build.directory}/angela/platform-kit-${project.version}/server/lib", "tc.properties").text = "\n\
+com.tc.server.entity.processor.threads=4\n\
```
mathieucarbou, line 153: no need for \n. In groovy, you have the `"""` quotes delimiting a multiline text:

```groovy
new File("${project.build.directory}/angela/platform-kit-${project.version}/server/lib", "tc.properties").text = """
com.tc.server.entity.processor.threads=4
com.tc.l2.tccom.workerthreads=4
...
"""
```

mathieucarbou (changes requested):

dynamic-config/testing/support/src/main/java/org/terracotta/dynamic_config/test_support/DynamicConfigIT.java
mathieucarbou: this won't solve the remote issue (only using `angela.java.opts` solves that) but for inline mode I think this is acceptable.

mathieucarbou (approved):
Approving so that you can merge after updating the script with `"""` (cleaner that \n)

mathieucarbou (approved):

dynamic-config/testing/system-tests/pom.xml
```diff
@@ -145,8 +145,20 @@
 
           <scripts>
             <script><![CDATA[
-new File(project.build.directory, "fake-hosts.txt").text = "127.0.0.1 ${InetAddress.getLocalHost().getHostName()} testhostname"
+new File(project.build.directory, "fake-hosts.txt").text = "127.0.0.1 ${InetAddress.getLocalHost().getHostName()} ${InetAddress.getLocalHost().getHostName().split('[.]')[0]} testhostname"
 println(new File(project.build.directory, "fake-hosts.txt").text)
+          ]]></script>
+            <script><![CDATA[
+new File("${project.build.directory}/angela/platform-kit-${project.version}/server/lib", "tc.properties").text = """
```
mathieucarbou, line 152: Script is failing on Windows:

```
[ERROR] Failed to execute goal org.codehaus.gmavenplus:gmavenplus-plugin:1.13.0:execute (fake-hosts) on project dynamic-config-system-tests: Error occurred while calling a method on a Groovy class from classpath. InvocationTargetException: startup failed:
[ERROR] Script2.groovy: 1: Unexpected input: '"' @ line 1, column 10.
[ERROR]    new File("D:\a\1\s\dynamic-config\testing\system-tests\target/angela/platform-kit-5.8-SNAPSHOT/server/lib", "tc.properties").text = """
[ERROR]             ^
[ERROR] 
[ERROR] 1 error
[ERROR] -> [Help 1]
[ERROR] 
```

To avoid that, you have to avoid using placeholders. Placeholders are replaced by maven before executing the script (which causes \" issues). But when using  properties like `project.build.directory` directly in a script, it tells groovy to read the script properties that are injected. The groovy plugin injects all maven properties in the groovy script.

Probably something like that would work: 

`new File(project.build.directory, "angela/platform-kit-${project.version}/server/lib/tc.properties")`

mathieucarbou (changes requested):

dynamic-config/testing/system-tests/pom.xml
```diff
@@ -145,8 +145,20 @@
 
           <scripts>
             <script><![CDATA[
-new File(project.build.directory, "fake-hosts.txt").text = "127.0.0.1 ${InetAddress.getLocalHost().getHostName()} testhostname"
+new File(project.build.directory, "fake-hosts.txt").text = "127.0.0.1 ${InetAddress.getLocalHost().getHostName()} ${InetAddress.getLocalHost().getHostName().split('[.]')[0]} testhostname"
```
mathieucarbou, line 148: We could have this change in my other PR, or, if we leave it heare, then could you please change: `InetAddress.getLocalHost().getHostName()` with `"hostname".execute().text.trim()` in both lines ? This change is required (comes from my other PR). 
Some tests were failing on Java 11 because `InetAddress.getLocalHost().getHostName()` is not returning the value of `hostname` on Azure vms... And what we need here is to reconstruct the right host file using the hostname.

mathieucarbou (changes requested):

azure-pipelines.yml
```diff
@@ -28,25 +28,25 @@ jobs:
     parameters:
       jobName: Linux_Java_8
       timeoutInMinutes: 180
-      options: -B -Djava.test.version=1.8
+      options: -B -Djava.test.version=1.8 -Dtest.parallel.forks=2
```
mathieucarbou, line 31: In my other PR, I saw that forkCount 4 is really too much for Azure...

dynamic-config/testing/system-tests/pom.xml
```diff
@@ -169,7 +181,7 @@ println(new File(project.build.directory, "fake-hosts.txt").text)
             <angela.rootDir>${project.build.directory}/angela</angela.rootDir>
             <angela.skipUninstall>true</angela.skipUninstall>
             <angela.kitCopy>false</angela.kitCopy>
-            <angela.tsa.fullLogging>true</angela.tsa.fullLogging>
+            <angela.tsa.fullLogging>false</angela.tsa.fullLogging>
```
mathieucarbou, line 184: avoid too many console logs

dynamic-config/testing/system-tests/pom.xml
mathieucarbou: Added a second commit to fix that :-)

dynamic-config/testing/system-tests/pom.xml
mathieucarbou: Added a second commit to fix that :-)

azure-pipelines.yml
mathieucarbou: I have closed #1074 (where 3/4 builds passed) to keep this PR only with all the changes.

dynamic-config/testing/system-tests/pom.xml
mathieucarbou: I have closed #1074 (where 3/4 builds passed) to keep this PR only with all the changes.

mathieucarbou (approved):

mathieucarbou:
@myronkscott : good news, everything works fine now, except one unstable test: `DetachInConsistency1x2IT#test_detach_active_node`. This is a timeout issue, so I added a third commit to fix it.

dynamic-config/testing/system-tests/pom.xml
mathieucarbou: Fixed in https://github.com/Terracotta-OSS/terracotta-platform/pull/1077

azure-pipelines.yml
mathieucarbou: Fixed in https://github.com/Terracotta-OSS/terracotta-platform/pull/1077

mathieucarbou (changes requested):
PR can be closed in favor of https://github.com/Terracotta-OSS/terracotta-platform/pull/1077.
All 3-4 latest builds were ok with the commit list from https://github.com/Terracotta-OSS/terracotta-platform/pull/1077.

Only some randomly failing tests to fix after that...

PR #1078, opened 2022-03-23
Improving test stability

mathieucarbou:
Only `test_detach_when_active_passive_disrupted` failed in [last build](https://dev.azure.com/TerracottaCI/terracotta/_build/results?buildId=9041&view=ms.vss-test-web.build-test-results-tab)

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -28,25 +28,25 @@ jobs:
     parameters:
       jobName: Linux_Java_8
       timeoutInMinutes: 180
-      options: -B -Djava.test.version=1.8 -Dtest.parallel.forks=2
+      options: -B -Djava.test.version=1.8 -Dtest.parallel.forks=3
       mavenGoals: 'clean verify'
   - template: build-templates/maven-common.yml@templates
     parameters:
       jobName: Linux_Java_11
       timeoutInMinutes: 180
-      options: -B -Djava.test.version=1.11 -Dtest.parallel.forks=2
+      options: -B -Djava.test.version=1.11 -Dtest.parallel.forks=3
       mavenGoals: 'clean verify'
   - template: build-templates/maven-common.yml@templates
     parameters:
       jobName: Windows_Java_8
       timeoutInMinutes: 180
       vmImage: 'windows-latest'
-      options: -B -Djava.test.version=1.8 -Dtest.parallel.forks=2
+      options: -B -Djava.test.version=1.8 -Dtest.parallel.forks=3
       mavenGoals: 'clean verify'
   - template: build-templates/maven-common.yml@templates
     parameters:
       jobName: Windows_Java_11
       timeoutInMinutes: 180
       vmImage: 'windows-latest'
-      options: -B -Djava.test.version=1.11 -Dtest.parallel.forks=2
+      options: -B -Djava.test.version=1.11 -Dtest.parallel.forks=3
       mavenGoals: 'clean verify'
--- a/common/test-utilities/src/main/java/org/terracotta/testing/JavaTool.java
+++ b/common/test-utilities/src/main/java/org/terracotta/testing/JavaTool.java
@@ -82,7 +82,7 @@ public class JavaTool {
     } catch (IOException e) {
       throw new UncheckedIOException(e);
     }
-    threadDumps(timeout).forEach(dump -> dump.writeTo(outputDir.resolve("thread-dump-" + dump.process.pid + "-" + dump.process.processName.replaceAll("\\W", "_") + ".txt")));
+    threadDumps(timeout).forEach(dump -> dump.writeTo(outputDir.resolve("thread-dump-" + dump.process.pid + "-" + dump.process.processName.replaceAll("\\W", "_") + ".log")));
   }
 
   public static void memoryDump(Process process, Path output, Duration timeout) {
--- a/dynamic-config/testing/support/src/main/java/org/terracotta/dynamic_config/test_support/DynamicConfigIT.java
+++ b/dynamic-config/testing/support/src/main/java/org/terracotta/dynamic_config/test_support/DynamicConfigIT.java
@@ -119,7 +119,6 @@ import static org.terracotta.utilities.test.matchers.Eventually.within;
 public class DynamicConfigIT {
   private static final Logger LOGGER = LoggerFactory.getLogger(DynamicConfigIT.class);
   private static final Duration DEFAULT_TEST_TIMEOUT = Duration.ofMinutes(2);
-  private static final Duration CONN_TIMEOUT = Duration.ofSeconds(30);
 
   protected static final String CLUSTER_NAME = "tc-cluster";
   protected static final AngelaOrchestratorRule angelaOrchestratorRule;
@@ -235,7 +234,9 @@ public class DynamicConfigIT {
             if (clusterDef.autoStart() && clusterDef.autoActivate()) {
               for (int stripeId = 1; stripeId <= clusterDef.stripes(); stripeId++) {
                 waitForActive(stripeId);
-                waitForPassives(stripeId);
+                LOGGER.info("1 ACTIVE server started for stripe: {}", stripeId);
+                int n = waitForPassives(stripeId);
+                LOGGER.info("{} PASSIVE servers started for stripe: {}", n, stripeId);
               }
             }
           }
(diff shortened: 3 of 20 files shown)


common/test-utilities/src/main/java/org/terracotta/testing/JavaTool.java
```diff
@@ -82,7 +82,7 @@ public static void threadDumps(Path outputDir, Duration timeout) {
     } catch (IOException e) {
       throw new UncheckedIOException(e);
     }
-    threadDumps(timeout).forEach(dump -> dump.writeTo(outputDir.resolve("thread-dump-" + dump.process.pid + "-" + dump.process.processName.replaceAll("\\W", "_") + ".txt")));
+    threadDumps(timeout).forEach(dump -> dump.writeTo(outputDir.resolve("thread-dump-" + dump.process.pid + "-" + dump.process.processName.replaceAll("\\W", "_") + ".log")));
```
mathieucarbou, line 85: Because azure is only collecting **/*.log and some other things. But not dumps in test-data/thread-dumps.

Will be easier to use .log for them since *.log files are collected by default.

dynamic-config/testing/system-tests/src/test/java/org/terracotta/dynamic_config/system_tests/network_disrupted/DetachInConsistency1x2IT.java
```diff
@@ -93,15 +93,16 @@ public void test_detach_when_active_passive_disrupted() throws Exception {
...
+      assertThat(getRuntimeCluster("localhost", getNodePort(1, 1)).getNodeCount(), is(equalTo(2)));
+
+      assertThat(getUpcomingCluster("localhost", getNodePort(1, 2)).getNodeCount(), is(equalTo(2)));
+      assertThat(getRuntimeCluster("localhost", getNodePort(1, 2)).getNodeCount(), is(equalTo(2)));
 
-    withTopologyService(1, 1, topologyService -> assertTrue(topologyService.isActivated()));
-    withTopologyService(1, 2, topologyService -> assertTrue(topologyService.isActivated()));
+      withTopologyService(1, 1, topologyService -> assertTrue(topologyService.isActivated()));
+      withTopologyService(1, 2, topologyService -> assertTrue(topologyService.isActivated()));
+    }
```
mathieucarbou, line 105: Moving the assertions before the `ServerToServerDisruptor` is closed to our servers are still working fine for the assertions

dynamic-config/testing/system-tests/src/test/java/org/terracotta/dynamic_config/system_tests/network_disrupted/DetachInConsistency1x2IT.java
```diff
@@ -92,16 +92,19 @@ public void test_detach_when_active_passive_disrupted() throws Exception {
...
+
       waitForPassive(1, passiveId);
-    }
-    assertThat(getUpcomingCluster("localhost", getNodePort(1, 1)).getNodeCount(), is(equalTo(2)));
-    assertThat(getRuntimeCluster("localhost", getNodePort(1, 1)).getNodeCount(), is(equalTo(2)));
+      waitForActive(1);
 
-    assertThat(getUpcomingCluster("localhost", getNodePort(1, 2)).getNodeCount(), is(equalTo(2)));
-    assertThat(getRuntimeCluster("localhost", getNodePort(1, 2)).getNodeCount(), is(equalTo(2)));
+      waitUntil(() -> getUpcomingCluster("localhost", getNodePort(1, 1)).getNodeCount(), is(equalTo(2)));
```
mathieucarbou, line 99: after the `undisrupt` call, it can take some seconds for the servers to recover. So we have to wait until we can connect back to the diagnostic port.

@myronkscott : FYI, It seems the diagnostic port is not available when the servers are in START_STATE. Here is the log of the passive below.

[stdout.txt](https://github.com/Terracotta-OSS/terracotta-platform/files/8339854/stdout.txt)

dynamic-config/testing/system-tests/pom.xml
```diff
@@ -139,16 +139,25 @@
...
 println(new File(project.build.directory, "fake-hosts.txt").text)
-          ]]></script>
-            <script><![CDATA[
+]]>
+                </script>
+              </scripts>
+            </configuration>
+          </execution>
+          <execution>
+            <id>tc.properties</id>
```
mathieucarbou, line 153: moved in its own execution for clarity

dynamic-config/testing/support/src/main/java/org/terracotta/dynamic_config/test_support/DynamicConfigIT.java
```diff
@@ -316,6 +316,11 @@ protected final void startNode(TerracottaServer node, Map<String, String> env, S
 
   protected final void stopNode(int stripeId, int nodeId) {
     angela.tsa().stop(getNode(stripeId, nodeId));
+    // waitForStopped is required because Ignite has a little cache
+    // and when the server is stopped, the angela server state read from the test jvm
+    // can still be an old state for some milliseconds.
+    // So this ensures the server has completed and also that this completion is seen on the test jvm
+    waitForStopped(stripeId, nodeId);
```
mathieucarbou, line 323: to know...

aurbroszniowski (approved):

dynamic-config/testing/system-tests/src/test/java/org/terracotta/dynamic_config/system_tests/network_disrupted/DetachInConsistency1x2IT.java
myronkscott: line 332?

```
2022-03-23 20:16:03,408 INFO - Action:MOVE_TO_ACTIVE not allowed because not enough servers are connected
2022-03-23 20:16:03,416 INFO - Terracotta Server instance has started diagnostic listening on  all interfaces (address:/0.0.0.0 port:13274)
2022-03-23 20:16:03,416 INFO - Moved to State[ START-STATE ]
```

mathieucarbou: Yes... Seems like de cannot access diagnostic port: is that possible ?

azure-pipelines.yml
```diff
@@ -28,25 +28,25 @@ jobs:
     parameters:
       jobName: Linux_Java_8
       timeoutInMinutes: 180
-      options: -B -Djava.test.version=1.8 -Dtest.parallel.forks=2
+      options: -B -Djava.test.version=1.8 -Dtest.parallel.forks=1
```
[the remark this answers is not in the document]
mathieucarbou: That is a test to check of everything runs fine. The problems I have with fork set to 2 is that :

1) some tests are really too heavy and run each one a lot of processes (with inline false) and we can easily have several severs running at the same time. In a container with limited resources this is just asking for problems

2) some processes died or were impossible to connect to. I saw that because some really simple test failed to connect in simple steps like condig-tool set or activate. These tests never failed in the past.

3) I really don't think that going to fork 2 will devide by 2 the build time. I have spent tremendous amount of time optimizing the TcDbTest projets from 2 hours to less than 30min now and inside VMs. And I didn't do that by increasing the fork count, which kills the resources.

Personally I prefer having a normal build running on a normal container, even if a little longer, than asking for troubles by squeezing the maximum of processes we can in a container which leads to nothing but random failure and unstable infra.

In this project, optimizing the build time can be done in a different way than using fork count. We could split the 3-4 sets of DC test in separate modules that we all build on separate slaves like I did for TcDbTest.

dynamic-config/testing/system-tests/src/test/java/org/terracotta/dynamic_config/system_tests/network_disrupted/DetachInConsistency1x2IT.java
myronkscott: i don't know.  i think that would be an underlying platform issue.  the listening port is open according to logging

mathieucarbou: This only happens when the system is very slow. Could it be possible that when angela sees `START-STATE`, then the test continue and the client starts to connect, but the port is not yet opened on the server ? Or it is 100% sure that the port will accept incoming connection when the log line is seen ?

azure-pipelines.yml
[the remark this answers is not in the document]
mathieucarbou: So reducing the conn and request timeouts reduces the tests from 2 hours to 1h30.
Now I am putting back forkCount == 3, but with `rerunFailingTestsCount == 2`.

PR #1080, opened 2022-03-28
Jackson 2.13.2.1 Upgrade

chrisdennis:
Upgrade to avoid https://nvd.nist.gov/vuln/detail/CVE-2020-36518


pom.xml
```diff
@@ -244,19 +244,11 @@
...
-        <version>${jackson.version}</version>
-      </dependency>
-      <dependency>
-        <groupId>com.fasterxml.jackson.datatype</groupId>
-        <artifactId>jackson-datatype-jdk8</artifactId>
-        <version>${jackson.version}</version>
-      </dependency>
-      <dependency>
-        <groupId>com.fasterxml.jackson.datatype</groupId>
-        <artifactId>jackson-datatype-jsr310</artifactId>
```
mathieucarbou, line 258: Why the removal of datatype-jdk8 and datatype-jsr310 ? If they are not in the classpath, they won't be picked up and we will break backward compatibility and eventually serialization / deserialisation in modules depending on `json-support`.

chrisdennis: This isn't the dependency section, it's the dependency management section, so it shouldn't affect that. I was however semi-forced to using the BOM because they haven't `2.13.2.1` versions of anything except core. That's why (I assume) the BOM has this date based version number... it depends on `2.13.2.1` of core and `2.13.2` of everything else.

mathieucarbou: oh I didn't notice sorry! thanks!

mathieucarbou (approved):

PR #1095, opened 2022-05-05
update voter test 

myronkscott:
because the server state may not update fast enough for the command to work the first time

### DIFF
--- a/dynamic-config/testing/system-tests/src/test/java/org/terracotta/dynamic_config/system_tests/activated/DetachCommandWithVoter1x2IT.java
+++ b/dynamic-config/testing/system-tests/src/test/java/org/terracotta/dynamic_config/system_tests/activated/DetachCommandWithVoter1x2IT.java
@@ -25,7 +25,6 @@ import org.terracotta.voter.ActiveVoter;
 import java.util.concurrent.CountDownLatch;
 
 import static org.hamcrest.Matchers.is;
-import static org.junit.Assert.assertThat;
 import static org.junit.Assert.assertTrue;
 import static org.terracotta.angela.client.support.hamcrest.AngelaMatchers.successful;
 
@@ -59,7 +58,7 @@ public class DetachCommandWithVoter1x2IT extends DynamicConfigIT {
       stopNode(1, passiveId);
       voted.await();
 
-      assertThat(configTool("detach", "-d", "localhost:" + getNodePort(1, activeId), "-s", "localhost:" + getNodePort(1, passiveId)), is(successful()));
+      waitUntil(()->configTool("detach", "-d", "localhost:" + getNodePort(1, activeId), "-s", "localhost:" + getNodePort(1, passiveId)), is(successful()));
 
       withTopologyService(1, activeId, topologyService -> assertTrue(topologyService.isActivated()));
     }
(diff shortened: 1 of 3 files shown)


mathieucarbou (approved):

dynamic-config/testing/system-tests/src/test/java/org/terracotta/dynamic_config/system_tests/activated/DetachCommandWithVoter1x2IT.java
```diff
@@ -59,7 +59,7 @@ public void testDetachAndVerifyWithVoter() throws InterruptedException {
       stopNode(1, passiveId);
       voted.await();
 
-      assertThat(configTool("detach", "-d", "localhost:" + getNodePort(1, activeId), "-s", "localhost:" + getNodePort(1, passiveId)), is(successful()));
+      waitUntil(configTool("detach", "-d", "localhost:" + getNodePort(1, activeId), "-s", "localhost:" + getNodePort(1, passiveId)), is(successful()));
```
mathieucarbou, line 62: wait you forgot the closure :-)
`waitUntil(() -> ..., is(successful()))`

dynamic-config/testing/system-tests/src/test/java/org/terracotta/dynamic_config/system_tests/activated/DetachCommandWithVoter1x2IT.java
```diff
@@ -82,7 +82,7 @@ public void testDetachAndAttachVerifyWithVoter() {
 
       startNode(1, passiveId);
 
-      assertThat(configTool("attach", "-d", "localhost:" + getNodePort(1, activeId), "-s", "localhost:" + getNodePort(1, passiveId)), is(successful()));
+      waitUntil(configTool("attach", "-d", "localhost:" + getNodePort(1, activeId), "-s", "localhost:" + getNodePort(1, passiveId)), is(successful()));
```
mathieucarbou, line 85: same

mathieucarbou (changes requested):

dynamic-config/testing/system-tests/src/test/java/org/terracotta/dynamic_config/system_tests/activated/DetachCommandWithVoter1x2IT.java
mathieucarbou: While being in platform... Could you please remove in the super class:
```java
  protected final void waitUntil(ToolExecutionResult result, Matcher<ToolExecutionResult> matcher) {
    angela.waitUntil(result, matcher);
  }
```
This method is unused and shouldn't be used now... I forgot to remove it following a refactor I guesss. We should use `waitUntil(() -> , is(...))`

mathieucarbou: Note: the reason is that this method is only compatible with the `InlineToolExecutionResult` which has a mistake btw because the `exitStatus` field is not volatile...
This method is not compatible with `ToolExecutionResult` (configTool not in inline mode)

mathieucarbou (approved):

PR #1100, opened 2022-06-29
Change kit

myronkscott:
Change the kit to be purely additive to the core kit.  We should not repackage the core kit as the platform components can be exploded on top of the core kit version.  This should make updating core in downstream projects not directly dependent on platform update and rebuild.


mathieucarbou:
thanks for removing the clustered map :P

dynamic-config/testing/system-tests/pom.xml
```diff
@@ -101,6 +101,21 @@
...
+                    </org.codehaus.plexus.components.io.filemappers.RegExpFileMapper>
+                  </fileMappers>
+                  <excludes>
+                    **\/server\/conf\/**\/*.xml,**\/server\/plugins\/lib\/default-configuration-*.jar
+                  </excludes>
+                  <groupId>org.terracotta.internal</groupId>
+                  <artifactId>terracotta-kit</artifactId>
+                  <version>${terracotta-core.version}</version>
+                  <type>tar.gz</type>
+                </artifactItem>
```
mathieucarbou, line 118: Why do the system tests (that ideally would treat a product as a near black box thing) need to know what to include / exclude from a kit dependency ?

Could we put that outside so that system tests only have ot point to 1 artifact to test ?

management/testing/integration-tests/pom.xml
```diff
@@ -122,6 +122,21 @@
...
+                    </org.codehaus.plexus.components.io.filemappers.RegExpFileMapper>
+                  </fileMappers>
+                  <excludes>
+                    **\/server\/conf\/**\/*.xml,**\/server\/plugins\/lib\/default-configuration-*.jar
+                  </excludes>
+                  <groupId>org.terracotta.internal</groupId>
+                  <artifactId>terracotta-kit</artifactId>
+                  <version>${terracotta-core.version}</version>
+                  <type>tar.gz</type>
+                </artifactItem>
```
mathieucarbou, line 139: same question, solving it would remove this code duplication

mathieucarbou (changes requested):

dynamic-config/testing/system-tests/pom.xml
myronkscott: the core kit will not include these things in the next release.  I can remove these excludes at that point.

mathieucarbou: That's why they were removed initially in platform-kit.xml, because this was the common point ;-) 

platform-kit.xml was acting as a bridge between the kit+plugins we want tt test, excluding the things we don't want from core.

myronkscott: excludes have been removed

myronkscott:
These changes are confirmed to work in ehcache3 and EE is unaffected.

diagnostic/client/pom.xml
```diff
@@ -57,7 +57,11 @@
       <artifactId>connection-api</artifactId>
       <scope>provided</scope>
     </dependency>
-
+    <dependency>
+      <groupId>org.slf4j</groupId>
+      <artifactId>slf4j-api</artifactId>
+      <scope>provided</scope>
```
mathieucarbou, line 63: What is the reason already why a required transitive dependency of a normal library used client -side which is not a server plugin would be provided ?

Does it mean everywhere in every module slf4j has to be provided now ?

dynamic-config/cli/cli-support/pom.xml
```diff
@@ -48,6 +48,7 @@
     <dependency>
       <groupId>ch.qos.logback</groupId>
       <artifactId>logback-classic</artifactId>
+      <scope>provided</scope>
```
mathieucarbou, line 51: What is the reason already why a required transitive dependency of a normal library used client -side which is not a server plugin would be provided ?

Does it mean everywhere in every module logging has to be provided now ? Why ? And why not the other dependencies?

dynamic-config/cli/config-tool/pom.xml
```diff
@@ -43,7 +43,7 @@
     <dependency>
       <groupId>org.terracotta.internal</groupId>
       <artifactId>client-runtime</artifactId>
-      <scope>runtime</scope>
+      <scope>provided</scope>
```
mathieucarbou, line 46: Same question

dynamic-config/cli/upgrade-tools/pom.xml
```diff
@@ -44,7 +44,11 @@
       <groupId>org.terracotta</groupId>
       <artifactId>terracotta-utilities-tools</artifactId>
     </dependency>
-
+    <dependency>
+      <groupId>ch.qos.logback</groupId>
+      <artifactId>logback-classic</artifactId>
+      <scope>test</scope>
```
mathieucarbou, line 50: Same question

dynamic-config/pom.xml
```diff
@@ -48,6 +48,7 @@
     <dependency>
       <groupId>org.slf4j</groupId>
       <artifactId>slf4j-api</artifactId>
+      <scope>provided</scope>
```
mathieucarbou, line 51: Why?

management/nms-agent-entity/nms-agent-entity-client/pom.xml
```diff
@@ -48,6 +48,7 @@
     <dependency>
       <groupId>org.slf4j</groupId>
       <artifactId>slf4j-api</artifactId>
+      <scope>provided</scope>
```
mathieucarbou, line 51: Why ? This is a normal library, used client -side which has a transitive dep requires to work...

voltron-proxy/voltron-proxy-client/pom.xml
```diff
@@ -45,6 +45,11 @@
       <artifactId>entity-common-api</artifactId>
       <scope>provided</scope>
     </dependency>
+    <dependency>
+      <groupId>org.slf4j</groupId>
+      <artifactId>slf4j-api</artifactId>
+      <scope>provided</scope>
+    </dependency>
```
mathieucarbou, line 52: Why ? It's client side

diagnostic/client/pom.xml
myronkscott: If we don't do this, the api jar gets pulled into the kit too many places.  If you are using this, it is in the context of some form of the client-runtime which also pulls slf4j and any other logging support.

dynamic-config/cli/cli-support/pom.xml
myronkscott: see above

diagnostic/client/pom.xml
myronkscott: We rely on connection-api and slf4j-api for compile but the runtime implementation and linking is determined elsewhere.

mathieucarbou: oh I remember the discussion yes. And this only applies to entity clients.

mathieucarbou (approved):

PR #1102, opened 2022-07-04
[daggy] NEVER ZAP a node when stopping it

mathieucarbou:
This bug has been introduced 30 April 2020.

When a node is stopped through DC API, it is also zapped.

The stop method is used:

1. From the repair command in the cluster-tool
2. From the detach command


mathieucarbou:
replaced by https://github.com/Terracotta-OSS/terracotta-platform/pull/1103

PR #1107, opened 2022-07-13
Update dependency versions -- logging component update

cljohnso:
This commit updates dependencies to consolidate the SLF4J and Logback
dependencies in upstream dependencies and this project.

* The versions of the following are updated:
** terracotta-configuration
** terracotta-core
** tc-passthrough-testins
** tc-tripwire
** galvan
** angela
** statistics
** terracotta-utilities
** logback
  The logback "base" version is set to 1.2.11
** slf4j
  The SLF4J "base" version is set to 1.7.32 (the minimum needed by
  logback 1.2.11

* Version ranges
  Version ranges are introduced for logback, slf4j, and
  terracotta-utilities.  The version of Logback is capped at
  1.2.9999 to avoid the early versions of 1.3.x.  The version of
  SLF4J is capped at 1.7.9999 to  avoid picking up 1.8.0-betaX
  releases which will not work with released versions of Logback
  before 1.3.x.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -31,18 +31,21 @@
 
   <properties>
     <maven.version>3.6.3</maven.version>
-    <slf4j.version>1.7.32</slf4j.version>
-    <logback.version>1.2.11</logback.version>
+    <slf4j.base.version>1.7.32</slf4j.base.version>
+    <slf4j.range.version>[${slf4j.base.version},1.7.9999)</slf4j.range.version>
+    <logback.base.version>1.2.11</logback.base.version>
+    <logback.range.version>[${logback.base.version},1.2.9999)</logback.range.version>
     <terracotta-apis.version>1.8.2</terracotta-apis.version>
-    <terracotta-configuration.version>10.7.2</terracotta-configuration.version>
-    <passthrough-testing.version>1.8.3</passthrough-testing.version>
-    <terracotta-core.version>5.9.4</terracotta-core.version>
-    <tripwire.version>1.0.3</tripwire.version>
-    <galvan.version>1.6.4</galvan.version>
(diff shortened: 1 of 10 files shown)


cljohnso:
Mostly exclusions for SLF4J in strategic places ...

mathieucarbou (commented):
@cljohnso : can't you move the exclusions in the parent pom where the depedencies are declared (eventually declaring them if needed) ?

Pretty sure we will mistakenly forgive one during maintenance or addition dev tests if these are not put in common.

cljohnso:
> @cljohnso : can't you move the exclusions in the parent pom where the depedencies are declared (eventually declaring them if needed) ?
> 
> Pretty sure we will mistakenly forgive one during maintenance or addition dev tests if these are not put in common.

Unless I missed some, the bulk of the exclusions are for modules declared in _this_ project -- Maven stupidly chooses the base version of the range when performing dependency convergence assessment for "local" modules.  If I missed a exclusion which should have been in dependencyManagment, please point it out and I'll correct it.

pom.xml
```diff
@@ -270,7 +273,13 @@
       <dependency>
         <groupId>ch.qos.logback</groupId>
         <artifactId>logback-classic</artifactId>
-        <version>${logback.version}</version>
+        <version>${logback.range.version}</version>
+        <exclusions>
```
chrisdennis, line 277: Worth a comment?

chrisdennis (approved):
Approved. It might be worth putting a comment on the logbook exclusion but that is optional.

mathieucarbou (approved):
thanks for the exclusion at the top...

PR #1108, opened 2022-07-21
[DO NOT MERGE] log directory can be unset so its presence is not required.

mathieucarbou:


tmesic99:
What happens if we start an unconfigured node with no config-dir defined, and then activate it - still without a defined config-dir.  Is that a desirable/valid state we want to support?

mathieucarbou:
> 

you mean log-dir instead of config-dir ? If yes, then yes we have always supported not having any log dir because the value is UNSET everywhere like backup dir.

mathieucarbou:
Note: I was able to test locally that this works:

`./kit/target/platform-kit-5.9-SNAPSHOT/platform-kit-5.9-SNAPSHOT/server/bin/start-tc-server.sh -auto-activate -cluster-name=foo -log-dir=`

We get the line:

`2022-07-21 23:12:19,866 INFO - Logging directory is not set. Logging only to the console`

mathieucarbou:
Needs discussion because break backward compat'

mathieucarbou:
I will recreate a new PR

PR #1110, opened 2022-07-22
Fix server plugins packaging:

mathieucarbou:
- Only add Class-Path in plugin jars, not all jars
- Fix missing provided scopes on some server provided jars (i.e. SLF4J) in plugins POM

I have verified with the following that we do not have nay ref to logback or slf4j inside class-path entries of jars on the server/plugins path:

```bash
for jar in `find kit/target/platform-kit-5.9-SNAPSHOT/platform-kit-5.9-SNAPSHOT/server/plugins -name "*.jar"`; do echo $jar; unzip -p $jar META-INF/MANIFEST.MF | grep -A 8 -i Class-Path | grep -i slf4j; done;

for jar in `find kit/target/platform-kit-5.9-SNAPSHOT/platform-kit-5.9-SNAPSHOT/server/plugins -name "*.jar"`; do echo $jar; unzip -p $jar META-INF/MANIFEST.MF | grep -A 8 -i Class-Path | grep -i logback; done;
```

### DIFF
--- a/client-message-tracker/pom.xml
+++ b/client-message-tracker/pom.xml
@@ -28,6 +28,22 @@
 
   <artifactId>client-message-tracker</artifactId>
 
+  <build>
+    <plugins>
+      <plugin>
+        <groupId>org.apache.felix</groupId>
+        <artifactId>maven-bundle-plugin</artifactId>
+        <configuration>
+          <archive>
+            <manifest>
+              <addClasspath>true</addClasspath>
+            </manifest>
+          </archive>
+        </configuration>
+      </plugin>
+    </plugins>
+  </build>
+
   <dependencies>
     <dependency>
       <groupId>org.terracotta</groupId>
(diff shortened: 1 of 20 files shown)


chrisdennis (approved):
This makes sense. Did you validate it works in the TC-EE kit still?

mathieucarbou:
> This makes sense. Did you validate it works in the TC-EE kit still?

Sadly I cannot without Clifford's work but considering the EE kit is just this kit with some more plugins, I don't see any reasons why it should not.

chrisdennis:
> considering the EE kit is just this kit with some more plugins

That's not really true any more... but I agree that if the platform kit is working it's highly likely the EE one will too.

mathieucarbou:
> > considering the EE kit is just this kit with some more plugins
> 
> That's not really true any more... but I agree that if the platform kit is working it's highly likely the EE one will too.

Before the PR I've looked at the jars of the EE kit and only the ones coming from platform were wrong

chrisdennis:
It's true that we use the platform JARs... but we pull them as dependencies... we don't extract them from the kit.

PR #1112, opened 2022-07-28
Use the classloader of common-json for the ObjectMappers

chrisdennis:


common/json-support/src/main/java/org/terracotta/json/ObjectMapperFactory.java
```diff
@@ -82,6 +83,7 @@ public ObjectMapperFactory withModules(Module... modules) {
 
   public ObjectMapper create() {
     ObjectMapper mapper = JsonMapper.builder()
+        .typeFactory(TypeFactory.defaultInstance().withClassLoader(getClass().getClassLoader()))
```
mathieucarbou, line 86: Makes total sense.

Did you test that ?

Just as a side node: I know Jackson is based on some java service discovery, but in DC we don't use any of that. All the Java modules are programmatically added to the object mapper.

So I know there is no module discovery, but I also know that jackson is using  service discovery for other kind of services too, like the ObjectMapper itself.

mathieucarbou (approved):

PR #1129, opened 2022-11-02
Further HostPort refactorings to decouple the parsing from the utility classes with the introduction of HostPort#parse methods

mathieucarbou:
I was trying to use HostPort in some voter / config-tool refac.

Here are some further changes regarding `HostPort` to move the parse() methods to the HostPort class in order to avoid to deal with the utility class each time we parse a HostPort.

Note: I specifically changes the method names to break compatibility in order to change the client code accordingly, in order to be able to remove the default values set for the ports.

It is up to the caller to know which port fits.

2 flavors of parse methods now:
- parse(String, int) => takes a default port in case none is supplied
- parse(String) => expect a port to be there (i.e. case for deserialisers for example)


common/inet-support/src/test/java/org/terracotta/inet/HostPortTest.java
```diff
@@ -45,4 +49,219 @@ private void assertHostPortException(String host) {
...
+        is(throwing(Matchers.instanceOf(IllegalArgumentException.class))
+            .andMessage(containsString("Missing port"))));
+
+    assertThat(() -> HostPort.parse(""),
+        is(throwing(Matchers.instanceOf(IllegalArgumentException.class))
+            .andMessage(containsString("Missing port"))));
+
+    assertThat(() -> HostPort.parse(":"),
+        is(throwing(Matchers.instanceOf(IllegalArgumentException.class))
+            .andMessage(containsString("Server must be an RFC 1123 compliant hostname or a valid IP address"))));
```
mathieucarbou, line 265: only added this test method. other methods were moved there from `InetSocketAddressConverterTest.java`

tmesic99 (approved):

PR #1130, opened 2022-11-03
[5.9] Some voter fixes regarding CLI parsing and topology used to start voter threads

mathieucarbou:
- Change the CLI parsing logic to treat comma as a normal separator, **to be consistent with other tools**: `-connect-to foo,bar -connect-to baz` will lead to a list of 3 items: foo, bar baz now

- Use the user-input addresses to fetch the topology, and not to determine how to start voter threads

- Start the voter thread based on the topology read and not based on user input (which can become obsolete with DC)


voter/src/main/java/org/terracotta/voter/cli/TCVoterMain.java
```diff
@@ -63,6 +84,41 @@ public void processArgs(String[] args) {
...
+    DiagnosticServiceProvider diagnosticServiceProvider = createDiagnosticServiceProvider(options);
+    ConcurrencySizing concurrencySizing = createConcurrencySizing();
+    MultiDiagnosticServiceProvider multiDiagnosticServiceProvider = new ConcurrentDiagnosticServiceProvider(diagnosticServiceProvider, null, concurrencySizing);
+
+    final Map<String, InetSocketAddress> addresses = options.getServersHostPort()
+        .stream()
+        .collect(toMap(identity(), hostPort -> InetSocketAddressConverter.parseInetSocketAddress(hostPort, 9410), (value1, value2) -> value1));
+
+    try (final DiagnosticServices<String> diagnosticServices = multiDiagnosticServiceProvider.fetchAnyOnlineDiagnosticService(addresses)) {
+      return diagnosticServices.getOnlineEndpoints()
```
myronkscott, line 111: Why only online?  When offline servers become available, a voter thread should be started for it.

myronkscott: oh.  nvm.  looks like this is just querying one of the online servers for the topology

mathieucarbou: Cluster tool is doing the exact same code to fetch a topology faster.

voter/src/main/java/org/terracotta/voter/cli/TCVoterMain.java
```diff
@@ -71,12 +127,21 @@ protected Optional<Properties> getConnectionProperties(Options option) {
...
+    validateStripesLimit(cluster);
+
+    // The endpoints we should use (public or internal) based on the user-provided list
+    final String[] hostPorts = cluster.determineEndpoints(options.getServersHostPort().stream().map(hostPort -> HostPort.parse(hostPort, 9410)).collect(toList()))
+        .stream()
+        .map(Node.Endpoint::getHostPort)
+        .map(HostPort::toString)
+        .toArray(String[]::new);
+
     startVoter(connectionProps, hostPorts);
```
mobasherul, line 145: how are u gonna handle this with multi stripe in enterprise ? Currently voter threads are equal to number of stripes and one voter thread is for 1 stripe

mathieucarbou: @mobasherul : actually... let me just refactor the code... I can make an impl that will work both for EE and OSS

mobasherul: got it so now only one of the online node in -connec-to is enough since we can fetch all info about topology and will start voter thread equal to number of stripes in the topology at that time,

mathieucarbou: yes! the code will be more robust. 

But for cloud, in kube, we will need to call the voter with ALL the known pods like this: `-connect-to node-1-1,node-1-2,node-2-2,node-2-1` to be sure the voter will be able to fetch a topology from at least one of them

mobasherul (approved):

voter/src/main/java/org/terracotta/voter/cli/TCVoterMain.java
```diff
@@ -71,13 +129,18 @@ protected Optional<Properties> getConnectionProperties(Options option) {
...
+    validateCLuster(cluster);
+
+    // The endpoints we should use (public or internal) based on the user-provided list
+    final List<HostPort> initiators = options.getServersHostPort().stream().map(hostPort -> HostPort.parse(hostPort, 9410)).collect(toList());
+    cluster.determineEndpoints(initiators)
+        .stream()
+        .collect(groupingBy(
+            endpoint -> cluster.getStripeByNode(endpoint.getNodeUID()).get().getUID(),
+            mapping(endpoint -> endpoint.getHostPort().toString(), toList())))
+        .forEach((uid, endpoints) -> startVoter(connectionProps, endpoints.toArray(new String[0])));
```
mathieucarbou, line 143: @mobasherul : this code can remain the same both for OSS and EE

voter/src/main/java/org/terracotta/voter/cli/TCVoterMain.java
```diff
@@ -88,9 +151,9 @@ protected void startVoter(Optional<Properties> connectionProps, String... hostPo
     new ActiveVoter(ID, new CompletableFuture<>(), connectionProps, hostPorts).start();
   }
 
-  protected void validateStripesLimit(String[] args) {
-    if (args.length > 1) {
-      throw new RuntimeException("Usage of multiple -connect-to options not supported");
+  protected void validateCLuster(Cluster cluster) {
+    if (cluster.getStripeCount() > 1) {
+      throw new RuntimeException("Usage of multiple stripes is not supported");
```
mathieucarbou, line 156: @mobasherul : we can just override that in EE

PR #1132, opened 2022-11-03
[5.9] Fixed Cluster#determineEndpoint(UID, Collection)

mathieucarbou:


dynamic-config/model/src/main/java/org/terracotta/dynamic_config/api/model/Cluster.java
```diff
@@ -540,17 +540,17 @@ public Collection<Endpoint> determineEndpoints() {
...
-    return determineEndpoint(node, Arrays.asList(initiators));
+  public Optional<Endpoint> determineEndpoint(UID nodeUID, HostPort... initiators) {
+    return determineEndpoint(nodeUID, Arrays.asList(initiators));
   }
 
-  public Optional<Endpoint> determineEndpoint(UID node, Collection<? extends HostPort> initiators) {
+  public Optional<Endpoint> determineEndpoint(UID nodeUID, Collection<? extends HostPort> initiators) {
     final Collection<? extends Endpoint> results = search(initiators);
     if (results.isEmpty()) {
-      return getNode(uid).map(Node::determineEndpoint);
```
mathieucarbou, line 550: was always returning `empty()` becasue `uid` is actually the cluster object uid... The wrong uid was used here...

dynamic-config/model/src/main/java/org/terracotta/dynamic_config/api/model/Cluster.java
```diff
@@ -540,17 +540,17 @@ public Collection<Endpoint> determineEndpoints() {
...
+      return getNode(nodeUID).map(Node::determineEndpoint);
     } else {
       final Endpoint endpoint = results.iterator().next();
-      return getNode(uid).map(n -> n.determineEndpoint(endpoint));
```
mathieucarbou, line 553: was always returning `empty()` becasue `uid` is actually the cluster object uid... The wrong uid was used here...

chrisdennis (approved):

PR #1135, opened 2022-11-03
[5.9] Introduced EndpointType

mathieucarbou:
This daggy fix is based on #1132


dynamic-config/model/src/main/java/org/terracotta/dynamic_config/api/model/Cluster.java
```diff
@@ -502,16 +501,31 @@ public Optional<Node> findReachableNode(HostPort hostPort) {
...
         .collect(toList());
   }
 
+  public EndpointType determineEndpointType(HostPort... initiators) {
+    return determineEndpointType(Arrays.asList(initiators));
+  }
+
+  public EndpointType determineEndpointType(Collection<? extends HostPort> initiators) {
+    // search in the topology the endpoints matching exactly the initiators
+    final Collection<? extends Endpoint> results = search(initiators);
```
chrisdennis, line 516: Why not:
```
Set<EndpointType> endpointTypes = search(initiators).map(Endpoint::getEndpointType).collect(toSet());
```

mathieucarbou: because the current code does the same thing but with less operations ?

mathieucarbou:
Updated daggy following previous merge

dynamic-config/model/src/main/java/org/terracotta/dynamic_config/api/model/Cluster.java
chrisdennis: I mainly meant to catch and cover the scenarios where we match multiple types of endpoints.

chrisdennis (approved):

dynamic-config/model/src/main/java/org/terracotta/dynamic_config/api/model/Cluster.java
mathieucarbou: yeah... In that case, I really don't know what to do... If we find a good solution for that I agree we have to change to a set and handle that.

I was thinking initially, if I see 2 different types or more, to fallback to the default determine() method, which is putting public endpoints in favor of internal ones, if they exist.... But the problem is that there is another type of existing endpoint, which is when the user sets a bind address equal to the hostname or its ip address (so not wildcard)... So the logic would not work.

mathieucarbou: I have changed the underlying implementation of `node.findEndpoints(addr)` to return a list of endpoints matching the address. 
- `determine*` methods are taking a decision
- `search` and `find` methods should not and should return correct results

So thanks to this fix, the search() methods now has a list of ALL node endpoints matching a given user input. So a node having both the same config for internal and public addresses will have its 2 endpoints returned.

Doing so, in the `determineEndpointType` method I can now return the endpointType that is having the maximum occurrences of endpoints.

It would then take a really badly configured network with several nodes having same internal hostnames and ports (like different stripes in subnets) and another node having this same address set as public, in order to break that logic... This cannot happen in cloud / Kube.

So this is not 100% perfect, but better than before.

chrisdennis (approved):

PR #1141, opened 2022-11-09
`config-tool activate -stripe <[name/]hostname[:port]|hostname[:port]|…>`

mathieucarbou:
Allows to quickly activate a cluster when all nodes are already started with the right parameters

**Examples:**

`config-tool -cluster-name tc-cluster -stripe node-1-1:9410|node-1-2,node-2-1:9410|node-2-2`

`config-tool -cluster-name tc-cluster -stripe node-1-1:9410|node-1-2 -stripe node-2-1:9410|node-2-2`

`config-tool -cluster-name tc-cluster -stripe stripe1/node-1-1:9410|node-1-2,stripe2/node-2-1:9410|node-2-2`

**Notes:**

- the separator characters used (`/` and `|`) are not a choice but more constrained coming from the [set of characters that we do not allow to have in both hostnames and stripe names](https://github.com/Terracotta-OSS/terracotta-platform/blob/master/dynamic-config/model/src/main/java/org/terracotta/dynamic_config/api/service/ClusterValidator.java#L58-L68). Also `,` is already taken by JCommander to split parameter values.
- Nodes are expected to have the exact same cluster-wide settings , otherwise validation failure. It means the nodes have to be started, like in a cloud env, with similar cluster-wide settings but can have each one different node-specific settings
- Nodes have to be started with an empty topology (they should be alone)

### DIFF
--- a/dynamic-config/cli/cli-support/src/main/java/org/terracotta/dynamic_config/cli/converter/ShapeConverter.java
+++ b/dynamic-config/cli/cli-support/src/main/java/org/terracotta/dynamic_config/cli/converter/ShapeConverter.java
@@ -1,0 +1,46 @@
+/*
+ * Copyright Terracotta, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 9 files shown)


mathieucarbou:
@mobasherul : this is still in progress but 80% done. I just need to add some tests and verify the logic.

You can start reviewing in case you see something I missed in the logic ;-)

Thanks!

mobasherul (approved):

mathieucarbou:
@mobasherul : finished adding a couple of tests. All works fine, PR ready to review and merge :-)

PR #1142, opened 2022-11-14
Refreshable Endpoints

mathieucarbou:
Introduced the concept of `Endpoints` when returning a collection of endpoints, so that we can re-use it in other projects to dynamically refresh the endpoints to use to connect to when a topology change occurs.

Ehcache PR: https://github.com/ehcache/ehcache3/pull/3079

### DIFF
--- a/dynamic-config/model/src/main/java/org/terracotta/dynamic_config/api/model/Cluster.java
+++ b/dynamic-config/model/src/main/java/org/terracotta/dynamic_config/api/model/Cluster.java
@@ -532,7 +532,12 @@ public class Cluster implements Cloneable, PropertyHolder {
         .collect(collectingAndThen(groupingBy(Endpoint::getEndpointType, counting()), map -> map.entrySet().stream()))
         .max(Map.Entry.comparingByValue())
         .map(Map.Entry::getKey)
-        .orElseGet(() -> determineEndpoints().iterator().next().getEndpointType());
+        .orElseGet(this::determineEndpointType);
+  }
+
+  public EndpointType determineEndpointType() {
+    // nodes all have public addresses or not have them set
+    return getStripes().get(0).getNodes().get(0).determineEndpoint().getEndpointType();
   }
 
   /**
@@ -546,16 +551,16 @@ public class Cluster implements Cloneable, PropertyHolder {
    *
    * @param initiators Addresses used to load this class, can be null.
    */
-  public Collection<Endpoint> determineEndpoints(HostPort... initiators) {
+  public Endpoints determineEndpoints(HostPort... initiators) {
     return determineEndpoints(Arrays.asList(initiators));
   }
 
-  public Collection<Endpoint> determineEndpoints(Collection<? extends HostPort> initiators) {
-    return determineEndpoints(determineEndpointType(initiators));
+  public Endpoints determineEndpoints(Collection<? extends HostPort> initiators) {
+    return new Endpoints(this, determineEndpointType(initiators));
   }
 
-  public Collection<Endpoint> determineEndpoints() {
-    return getNodes().stream().map(Node::determineEndpoint).collect(toList());
+  public Endpoints determineEndpoints() {
+    return new Endpoints(this, determineEndpointType());
   }
 
   public Optional<Endpoint> determineEndpoint(UID nodeUID, HostPort... initiators) {
@@ -571,11 +576,11 @@ public class Cluster implements Cloneable, PropertyHolder {
     return getNode(nodeUID).map(node -> node.determineEndpoint(endpointType));
   }
 
-  public Collection<Endpoint> determineEndpoints(Endpoint initiator) {
+  public Endpoints determineEndpoints(Endpoint initiator) {
     return determineEndpoints(initiator.getEndpointType());
   }
 
-  public Collection<Endpoint> determineEndpoints(EndpointType endpointType) {
-    return getNodes().stream().map(node -> node.determineEndpoint(endpointType)).collect(toList());
+  public Endpoints determineEndpoints(EndpointType endpointType) {
+    return new Endpoints(this, endpointType);
   }
 }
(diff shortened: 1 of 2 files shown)


mathieucarbou:
rebased

dynamic-config/model/src/main/java/org/terracotta/dynamic_config/api/model/Endpoints.java
```diff
@@ -0,0 +1,117 @@
...
+
+  public Iterator<Node.Endpoint> iterator(UID stripeUID) {
+    final Cluster c = this.cluster.get();
+    return c.getStripe(stripeUID)
+        .map(stripe -> endpoints().filter(endpoint -> stripe.containsNode(endpoint.getNodeUID())))
+        .orElse(Stream.empty())
+        .iterator();
+  }
+
+  private Stream<Node.Endpoint> endpoints() {
```
chrisdennis, line 72: If we also add:
```java
private Stream<Node.Endpoint> endpoints(UID stripeUID) {
  return cluster.getStripe(stripeUID).map(s -> s.getNodes().stream().map(node -> node.determineEndpoint(endpointType))).orElse(Stream.empty());
}
```
then we. can switch to:
```java
public Iterable<InetSocketAddress> asInetSocketAddresses() {
  return () -> endpoints().map(e -> e.getHostPort().createInetSocketAddress()).iterator();
}

public Iterable<HostPort> asHostPorts() {
  return () -> endpoints().map(Node.Endpoint::getHostPort).iterator();
}

public Iterable<InetSocketAddress> asInetSocketAddresses(UID stripeUID) {
  return () -> endpoints(stripeUID).map(e -> e.getHostPort().createInetSocketAddress()).iterator();
}

public Iterable<HostPort> asHostPorts(UID stripeUID) {
  return () -> endpoints(stripeUID).map(Node.Endpoint::getHostPort).iterator();
}
```

dynamic-config/model/src/main/java/org/terracotta/dynamic_config/api/model/Endpoints.java
```diff
@@ -0,0 +1,117 @@
...
+
+  public Endpoints(Cluster cluster) {
+    this(cluster, cluster.determineEndpointType());
+  }
+
+  public Endpoints(Cluster cluster, Collection<? extends HostPort> initiators) {
+    this(cluster, cluster.determineEndpointType(requireNonNull(initiators)));
+  }
+
+  public Endpoints(Cluster cluster, EndpointType endpointType) {
```
chrisdennis, line 46: There seems to be some confusion between the set of constructors exposed here and the set of methods available on `Cluster`. I'm thinking it might be cleaner to rationalize this just this constructor here, and lean on the methods in Cluster for the rest?

dynamic-config/model/src/main/java/org/terracotta/dynamic_config/api/model/Cluster.java
```diff
@@ -571,11 +576,11 @@ public Optional<Endpoint> determineEndpoint(UID nodeUID, EndpointType endpointTy
     return getNode(nodeUID).map(node -> node.determineEndpoint(endpointType));
   }
 
-  public Collection<Endpoint> determineEndpoints(Endpoint initiator) {
-    return determineEndpoints(initiator.getEndpointType());
+  public Endpoints determineEndpoints(Endpoint initiator) {
+    return new Endpoints(this, initiator.getEndpointType());
```
chrisdennis, line 580: Perhaps 
```java
return determineEndpoints(initiator.getEndpointType());
```
?

chrisdennis (changes requested):

dynamic-config/model/src/main/java/org/terracotta/dynamic_config/api/model/Endpoints.java
```diff
@@ -0,0 +1,117 @@
...
+import java.util.stream.Stream;
+
+import static java.util.Objects.requireNonNull;
+
+/**
+ * @author Mathieu Carbou
+ */
+public final class Endpoints extends AbstractCollection<Node.Endpoint> {
+
+  private final AtomicReference<Cluster> cluster = new AtomicReference<>();
```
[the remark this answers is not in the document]
mathieucarbou: In our use case when in a callback thread from core, yes, but should we limit to this only use case ?

dynamic-config/model/src/main/java/org/terracotta/dynamic_config/api/model/Endpoints.java
mathieucarbou: thanks! far easier lol!

dynamic-config/model/src/main/java/org/terracotta/dynamic_config/api/model/Endpoints.java
mathieucarbou: yes we can do that, That is tricky otherwise for the mind. But it will make callers do more stuff also. I wanted to simplify the usages fro callers because they can call the ctors directly.

dynamic-config/model/src/main/java/org/terracotta/dynamic_config/api/model/Endpoints.java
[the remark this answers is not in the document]
chrisdennis: The semantics of an atomic set and a volatile write are the same.

> The memory effects for accesses and updates of atomics generally follow the rules for volatiles, as stated in The Java Language Specification (17.4 Memory Model) :
> * get has the memory effects of reading a volatile variable.
> * set has the memory effects of writing (assigning) a volatile variable.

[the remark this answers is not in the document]
mathieucarbou: Oh !!! I didn't know... But it makes sense. Sorry!

chrisdennis (approved):

PR #1144, opened 2022-11-18
Fix MultiDiagnosticServiceProvider which was incorrectly throwing exceeption and implementing the needed APIs

mathieucarbou:


dynamic-config/cli/dynamic-config-cli-api/src/main/java/org/terracotta/dynamic_config/cli/api/command/RemoteAction.java
```diff
@@ -400,7 +400,7 @@ protected final Cluster getRuntimeCluster(Collection<HostPort> nodes) {
   }
 
   protected final <V> V withAnyOnlineDiagnosticService(Collection<HostPort> nodes, BiFunction<HostPort, DiagnosticService, V> fn) {
-    try (DiagnosticServices<HostPort> diagnosticServices = multiDiagnosticServiceProvider.fetchAnyOnlineDiagnosticService(toAddr(nodes, identity(), HostPort::createInetSocketAddress), null)) {
+    try (DiagnosticServices<HostPort> diagnosticServices = multiDiagnosticServiceProvider.fetchAnyOnlineDiagnosticService(toAddr(nodes, identity(), HostPort::createInetSocketAddress))) {
       final Map.Entry<HostPort, DiagnosticService> entry = diagnosticServices.getOnlineEndpoints().entrySet().iterator().next();
       return fn.apply(entry.getKey(), entry.getValue());
```
mobasherul, line 405: should we add check here that we got atleast one online before calling fn ?

mathieucarbou: no need: look at `fetchAnyOnlineDiagnosticService ` javadoc: `fetchAnyOnlineDiagnosticService ` will throw if no online node found.

diagnostic/client/src/main/java/org/terracotta/diagnostic/client/connection/ConcurrentDiagnosticServiceProvider.java
```diff
@@ -49,104 +44,138 @@ public ConcurrentDiagnosticServiceProvider(DiagnosticServiceProvider diagnosticS
...
+      // some connection timeout decisions from user:
+      // - either timeout (long or short)
+      // - either null => default core timeout is used
+      // - either interruption
       try {
-        // capture the task output
-        int count = addresses.size();
-        while (count-- > 0) {
-          // we do not need to handle any timeout here during a take or get because they are handled in the submitted tasks
-          Future<Tuple3<K, DiagnosticService, DiagnosticServiceProviderException>> completed = completionService.take();
```
mobasherul, line 108: @mathieucarbou How the connection timeout works tests have long time out so these take calls will block and tests will time out for tests trying to connect to offline nodes .

diagnostic/client/src/main/java/org/terracotta/diagnostic/client/connection/ConcurrentDiagnosticServiceProvider.java
```diff
@@ -49,104 +44,138 @@ public ConcurrentDiagnosticServiceProvider(DiagnosticServiceProvider diagnosticS
...
         } catch (Exception e) {
-          return tuple3(id, null, new DiagnosticServiceProviderException("Failed to create diagnostic connection to " + address, e));
+          onFailure.accept(k, new DiagnosticServiceProviderException("Failed to create diagnostic connection to: " + address + ": " + e.getMessage(), e));
         }
       }));
-
-      Map<K, DiagnosticService> online = new HashMap<>(addresses.size());
-      Map<K, DiagnosticServiceProviderException> offline = new HashMap<>(addresses.size());
-
+      executor.shutdown();
```
mobasherul, line 157: We are not waiting here for connections to be completed. I am also intrigued how the tests were passing earlier when trying to connect to offline nodes as take()  in ExecutioncompletionService will block there as well .

diagnostic/client/src/main/java/org/terracotta/diagnostic/client/connection/ConcurrentDiagnosticServiceProvider.java
mathieucarbou: in the config tool we use null when we don't know if some nodes are offline or not. null means default core timeout of 5 sec. I need to fix osmething because tests are waiting too long now I think.

diagnostic/client/src/main/java/org/terracotta/diagnostic/client/connection/ConcurrentDiagnosticServiceProvider.java
mathieucarbou: they were passing because when we connect to a set of nodes when we don't know the states, we connnect with a nul ll timeout (== 5 sec timeout). We don't need to wait because the wait is done in the close() : this is a try-resource. Wait is done whether shutdown or shutdownNow iscalled.

mathieucarbou:
@mobasherul : please look at my latest changes here: https://github.com/Terracotta-OSS/terracotta-platform/compare/0ce655c11ae9ea5152d295e8726eed4562be1b93..3e964db3f6809998d01c4b33e6dac2ff381b660d

Test were failing because of this: https://github.com/Terracotta-OSS/terracotta-platform/compare/0ce655c11ae9ea5152d295e8726eed4562be1b93..3e964db3f6809998d01c4b33e6dac2ff381b660d#diff-fd2d62ddaa2a6b2a659c9ff9f46f05dbc66265e3ded1a1868d1787ed7f7cf754L126-L127 : I was constructing the fetcher with the field `connctionTimeout`, not the user-overridden value `connTimeout` which can be set to null when callers want short timeouts

diagnostic/client/src/main/java/org/terracotta/diagnostic/client/connection/ConcurrentDiagnosticServiceProvider.java
mathieucarbou: yes: every time we do a discovery where it is possible that somes nodes are down, we use `null`. Example: the call to `getLogicalServerStates` that returns a map because some noes can be unreachable.

But in all other calls were we expect all nodes to be online, we can use the user-provided (and potentially long) timeout

mobasherul: Ah for explicit connection also we can pass some other timeout . I was thinking we always use configured timeouts got it. Hopefully the tests would pass now

mobasherul (approved):

PR #1146, opened 2022-12-08
config-tool attach -stripe-shape [name/]hostname[:port]|hostname[:port] -to-cluster <hostname[:port]>

mathieucarbou:
This is the introduction of the `-stripe-shape` parameter to be able to attach a new stripe to a cluster (activated or not) in one shot without having to prepare the topology first by attaching new nodes one by one.

The stripe is formed dynamically thanks to the list of nodes (and eventually stripe name) passed to the cli.

Required for the Kube Operator.


mobasherul (approved):

mathieucarbou:
@mobasherul : added some system tests, all works fine. I will test the snapshot in EE and merge if OK.

PR #1151, opened 2023-01-26
[daggy-5.9] Correctly rethrow all exceptions seen instead of the first one seen

mathieucarbou:


mathieucarbou:
equivalent master PR merged

PR #1152, opened 2023-01-26
[daggy-master] Voter fixes

mathieucarbou:


chrisdennis (approved):
Feels like this code is due for an overhaul... but these changes are moving it in a positive direction.

mathieucarbou:
> Feels like this code is due for an overhaul... but these changes are moving it in a positive direction.

yes... we really have troubles with Haque to follow the voter code... It's very hard.

myronkscott (approved):

PR #1157, opened 2023-01-26
[daggy-5.9] Bugfix: wait at voter startup for a topology of an activated cluster to be available.

mathieucarbou:
This fix prevents voter to fetch at startup a cluster topology from a node in diagnostic mode, and ensures the voter will start only when it can fetch a topology from an activated cluster. Some nodes passed to the CLI could still be or become in diagnostic mode.


mathieucarbou:
equivalent master PR merged

PR #1163, opened 2023-02-07
Introduced a validation to prevent further scale out operation when a previous one was rolled back

mathieucarbou:
Once user fixed the problem, he can allow further scale out operation to be done by executing: `config-tool repair -force allow_scale_out`


mathieucarbou:
closed as he work required is a little larger and needs also to be coupled with some EE changes in *Actions.

PR #1164, opened 2023-02-13
[5.9] Prevent further scale in/out process when a previous one failed, independently of the lock/unlock state of the config

mathieucarbou:
### Overview

- Introduced some validations to prevent scale in/out operation to be performed if one failed previously
- Introduced a repair mode for the user to allow the operation to proceed with `config-tool repair -force allow_scaling`
- Refactored the lock/unlock logic in attach and detach to correctly catch exceptions and add markers to deny any further scale in/out in case of errors
- Introduced some flags in the diagnostic output to show if scale in/out process is allowed or not
- Refactored the inheritance logic for attach and detach between OSS and EE to consolidate the lock/unlock/marker logic at one place and only let EE override the difference in behaviour

### Scale in flow

**detach command (CLI)**

1. validation: fails if we find: a `deny scale in` marker
2. `lock`
3. trigger rebalancing
    - on failure:
        * try to place a marker `deny scale in` to prevent replaying the detach
        * try to `unlock`

**on rebalancing success (server-side)**

1. `detach` the stripe
2. `unlock` on nomad tx success

**on rebalancing failure (server-side)**

1. place a `deny scale in` marker
2. `unlock`

### Scale out flow

**attach command (CLI)**

1. validations
    - fails if we find: a `deny scale out` marker
3. `lock`
4. `attach`
    - on failure:
        * try to place a marker `deny scale out` to prevent replaying the attach
        * try to `unlock`
        * in any case, either a marker is placed or the config is kept locked
6. trigger rebalancing on nomad success
    - on failure:
        * config is kept locked

**on rebalancing success (server-side)**

1. `unlock`

**on rebalancing failure (server-side)**

1. TRY `detach`
2. FINALLY place a `deny scale out` marker
3. FINALLY `unlock`

### Why adding a marker if the `attach` or `detach` CLI fails ?

Because attach and detach are triggering 2 Nomad tx to lock and unlock (discovery/prepare/commit) and replaying would cause 2 problems:
- un-necessary append-log entries would fill the append-log
- un-necessary nomad transactions triggered would increase the chance to collide with a concurrent user transaction which aims are doing a valid config change or repair

### Repairing

We can re-allow a scale op to be retried by running:

`config-tool repair -force allow_scaling`

### DIFF
--- a/dynamic-config/cli/dynamic-config-cli-api/src/main/java/org/terracotta/dynamic_config/cli/api/command/TopologyAction.java
+++ b/dynamic-config/cli/dynamic-config-cli-api/src/main/java/org/terracotta/dynamic_config/cli/api/command/TopologyAction.java
@@ -19,6 +19,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.terracotta.diagnostic.model.LogicalServerState;
 import org.terracotta.dynamic_config.api.model.Cluster;
+import org.terracotta.dynamic_config.api.model.LockTag;
 import org.terracotta.dynamic_config.api.model.Node.Endpoint;
 import org.terracotta.dynamic_config.api.model.nomad.TopologyNomadChange;
 import org.terracotta.dynamic_config.api.service.ClusterValidator;
@@ -29,6 +30,7 @@ import java.util.Collection;
 import java.util.HashSet;
 import java.util.Map;
 import java.util.Set;
+import java.util.function.Consumer;
 import java.util.function.Supplier;
 
 import static java.lang.System.lineSeparator;
@@ -41,6 +43,9 @@ import static org.terracotta.dynamic_config.api.model.ClusterState.CONFIGURING;
 public abstract class TopologyAction extends RemoteAction {
   private static final Logger LOGGER = LoggerFactory.getLogger(TopologyAction.class);
 
+  @Injector.Inject
+  public final UnlockConfigAction unlockAction = new UnlockConfigAction();
+
   protected OperationType operationType = OperationType.NODE;
   protected HostPort destinationHostPort;
   protected boolean force;
@@ -50,6 +55,26 @@ public abstract class TopologyAction extends RemoteAction {
   protected Map<Endpoint, LogicalServerState> destinationOnlineNodes;
   protected boolean destinationClusterActivated;
   protected Cluster destinationCluster;
+
+  protected boolean useLock; // if set, a lock will try to be created before the operation and unlocked after
+  protected String createdLockToken;
+  private boolean unlock;
+
+  public void setLock(boolean lock) {
+    this.useLock = lock;
+  }
+
+  public boolean isLockRequired() {
+    return useLock;
+  }
+
+  public boolean isUnlockRequired() {
+    return unlock;
+  }
+
+  public void setUnlock(boolean unlock) {
+    this.unlock = unlock;
+  }
 
   public void setOperationType(OperationType operationType) {
     this.operationType = operationType;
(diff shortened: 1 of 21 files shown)


dynamic-config/cli/dynamic-config-cli-api/src/main/java/org/terracotta/dynamic_config/cli/api/command/TopologyAction.java
```diff
@@ -79,10 +104,29 @@ protected void validate() {
...
+  protected final boolean isScaleInOrOut() {
+    return destinationClusterActivated && operationType == OperationType.STRIPE;
+  }
+
   @Override
-  public void run() {
+  public final void run() {
     validate();
-    execute();
+    final Runnable scaleTask = () -> tryCatch(this::execute, onExecuteError());
```
mobasherul, line 114: what is the use of deny lock tags for onExecuteError ? Suppose u already are firing detach with lockAwareNomadManager meaning attach stripe failed as part of rollback u are firing detach and that detach failed ?

mathieucarbou: putting a marker here avoids going through the start of the rebalancing process again in case triggerring the rebalancing fails, so that the operator does not keep trying again and again.

For detach, the steps are:

1. validation
2. lock
3. trigger rebalancing

on rebalancing success:
1. detach
2. unlock (on nomad sucsess)

on rebalancing failure:
1. unlock

`onExecuteError` is the callback that is called if the execute method fails, in the case of detach in EE, if the step 3 fails.

If step 3 fails, we put a marker to avoid replaying the scale in op until the user allows it.

Line 123: `tryCatch(scaleTask, mustLock ? e -> unlock(destinationCluster, destinationOnlineNodes) : e -> {});` makes sure the config gets unlocked in case of error if a lock was placed before. This is required to make sure the user can trigger some config changes or repair. This fixes the bug that was there before where no unlock was called in the catch. The markers will take care of preventing the replay of a scale op.

dynamic-config/cli/dynamic-config-cli-api/src/main/java/org/terracotta/dynamic_config/cli/api/command/TopologyAction.java
```diff
@@ -79,10 +104,29 @@ protected void validate() {
...
+    final boolean mustLock = isLockRequired() && destinationClusterActivated && !isLockAwareNomadManager();
+    if (mustLock) {
+      // if we need to lock the operation, lock it first
+      createdLockToken = lock(destinationCluster, destinationOnlineNodes, LockTag.OWNER_PLATFORM, buildLockTag());
+      output.out("Config lock with token: " + createdLockToken);
+      destinationCluster = getUpcomingCluster(destination);
+    }
+    // then we run the scale task. In case of error, we unlock, but only if we placed a lock
+    tryCatch(scaleTask, mustLock ? e -> unlock(destinationCluster, destinationOnlineNodes) : e -> {});
```
mobasherul, line 123: In topologyAction.execute() we are not throwing any exception so StripeAdditionNomadChange/StripeRemovalNomadChange fails we have to unlock

dynamic-config/cli/dynamic-config-cli-api/src/main/java/org/terracotta/dynamic_config/cli/api/command/TopologyAction.java
mobasherul: But any way we are not unlocking if detach fails as part of attach stripe failure and if detach succeeds as part of attach stripe failure we have the seq to stop retrying

mathieucarbou: @mobasherul : agree but this is another kind of failure (server-side). we were talking about the failure within the CLI when starting the rebalancing. I've just updated the description of the PR to add a lot of details regarding the new flow and error handling. I will let you have a look and if you still see some things not handled let me know 😄

mobasherul: I feel "lock unlock" is enough for CLI whether attach/detach and we might not need this explicit marker and it doesn't matter whether we scale up/down if the previous scale operation failed we should not allow any scaling operation . WDYT?

mathieucarbou: scale in and out are different operations: what are the reasons to block a scale in if a previous scale out failed or was rolled back ?

Is `lock/unlock` enough ? No... If you don't put a marker, the operator or a user manually or through a script could retry the operation indefinitely, which will trigger a serie of "lock, operation, failure, unlock" indefinitely. This can cause a deny of service of the configuration system and ultimately the servers.

mobasherul: @mathieucarbou I guess becuase we want manual intervention if the scaling operation fails. lock/unlock is what u are using for detach operation right ?

mathieucarbou: Let's talk about that tomorrow in lead whether we should block all scale op or just one type. I don't see any reasons why we should block. Currently, we don't and we allow. So why would we remove a feature ? The only thing we discussed in previous meeting was about **preventing the user to replay a previously failed scaling op.**

`detach` command used the flow as described above in the pr description... I've inserted a marker in case CLI fails and added the missing unlock in case the CLI fails, which was missing before.

mathieucarbou: discussed, we will use markers everywhere.

mathieucarbou:
@mobasherul @chrisdennis : ready for review.

mobasherul (approved):

PR #1169, opened 2023-03-07
Re-implement voter

myronkscott:
This is an attempt to cleanup voter code.  The actions are all the same, just some re-organization of the pieces and some additions/changes to help tests.  There are three main parts.  VotingGroup is in charge of voting structures for each stripe. It continuously polls the topology and tries to maintain registrations with all the members that are alive.   VotingGroup also makes sure only one server is being voted for during a voting session.  This replaces ActiveVoter.  ClientVoterThread is in charge of registration and heartbeating for each server in the stripe.  ClientVoteManger is the connection plumbing to each server.


dynamic-config/testing/system-tests/pom.xml
```diff
@@ -214,6 +214,7 @@ println(new File(project.build.directory, "angela/platform-kit-${project.version
             <org.terracotta.voter.topology.fetch.interval>9000</org.terracotta.voter.topology.fetch.interval>
             <jdk.net.hosts.file>${project.build.directory}/fake-hosts.txt</jdk.net.hosts.file>
             <IGNITE_UPDATE_NOTIFIER>false</IGNITE_UPDATE_NOTIFIER>
+            <org.terracotta.voter.topology.fetch.interval>5000</org.terracotta.voter.topology.fetch.interval>
```
mathieucarbou, line 217: Can we move it to a CLI parameter like it was done with the other timeout values ?

voter/src/main/java/org/terracotta/voter/ClientVoterManagerImpl.java
```diff
@@ -22,24 +22,24 @@
...
 
   public static final String REQUEST_TIMEOUT = "Request Timeout";
+  public static final String MBEAN_NAME = "VoterManager";
 
   private final String hostPort;
   public Diagnostics diagnostics;
 
   private volatile boolean voting = false;
-  private volatile long generation = 0;
+  private volatile long generation = -2;
```
mathieucarbou, line 42: Why it has to be initialized to -2 ? What -2 means ? Introduce constant maybe ?

voter/src/main/java/org/terracotta/voter/ClientVoterThread.java
```diff
@@ -0,0 +1,112 @@
...
+        voteRequest.accept(mgr);
+      }
+    }, 0L, HEARTBEAT_INTERVAL, TimeUnit.MILLISECONDS);
+    this.operational.complete(null);
+  }
+
+  // For testing
+  private final CompletableFuture<?> operational = new CompletableFuture<>();
+
+  public CompletableFuture<?> operational() {
```
mathieucarbou, line 109: It could just be that to simply usage and avoids leaking internals ? 
```java
public void awaitOperational() throws ... {
    this.latch.await();
}
```

voter/src/main/java/org/terracotta/voter/ClientVoterThread.java
```diff
@@ -0,0 +1,112 @@
...
+        LOGGER.warn("Heart-beating with {} not connected", mgr.getTargetHostPort(), run);
+        close();
+        voteRequest.accept(mgr);
+      }
+    }, 0L, HEARTBEAT_INTERVAL, TimeUnit.MILLISECONDS);
+    this.operational.complete(null);
+  }
+
+  // For testing
+  private final CompletableFuture<?> operational = new CompletableFuture<>();
```
mathieucarbou, line 107: latch not enough ?

voter/src/main/java/org/terracotta/voter/ClientVoterThread.java
```diff
@@ -0,0 +1,112 @@
...
+  public void operate(Consumer<ClientVoterManager> voteRequest) throws ConnectionException, TimeoutException {
+    if (currentTask != null && !currentTask.isDone()) {
+      throw new IllegalStateException();
+    }
+    if (!register()) {
+      throw new RuntimeException("unable to register");
+    }
+    currentTask = exec.scheduleAtFixedRate(() -> {
+      try {
+        long election = mgr.heartbeat(id);
```
mathieucarbou, line 79: why is the latch (or future) not decremented (or completed) after the heartbeat call to make sure it is really operational ?

voter/src/main/java/org/terracotta/voter/VotingGroup.java
```diff
@@ -0,0 +1,410 @@
...
+  private synchronized void setTargets(String[] nodes) {
+    targets = nodes;
+  }
+  
+  public synchronized Set<String> getExistingTopology() {
+    return new HashSet<>(Arrays.asList(targets));
+  }
+  
+  private VoterStatus getStatus() {
+    return new VoterStatus() {
```
mathieucarbou, line 389: could be a private final field since there is no state per method call (`getStatus()` )

voter/src/main/java/org/terracotta/voter/ClientVoterThread.java
mathieucarbou, line 79: since heartbeat() does not throw interrupted exception, should we check for currentTask.isCancelled() and Thread.currentThread.isInterrupted juste after to not do anything ? Since this could be closed by the voterGroup ?

voter/src/main/java/org/terracotta/voter/VotingGroup.java
```diff
@@ -0,0 +1,410 @@
...
+  public void stop() {
+    LOGGER.info("Stopping {}", this);
+    reset();
+    sharedExecutor.shutdown();
+    try {
+      sharedExecutor.awaitTermination(30, TimeUnit.SECONDS);
+    } catch (InterruptedException ie) {
+      LOGGER.info("shutdown interrupted", ie);
+    }
+    this.voter.interrupt();
```
mathieucarbou, line 262: this.voter.join() also ?

voter/src/main/java/org/terracotta/voter/VotingGroup.java
```diff
@@ -0,0 +1,410 @@
...
+
+  @Override
+  public String toString() {
+    return "VotingGroup{" + nodes.keySet() + "}";
+  }
+  
+  // Below is all cruft for testing help
+  private final List<Consumer<String>> votingListeners = new CopyOnWriteArrayList<>();
+  private final CompletableFuture<?> bootstrapped = new CompletableFuture<>();
+  private CompletableFuture<?> pollingFuture = new CompletableFuture<>();
```
mathieucarbou, line 346: volatile ? Reset through another thread

dynamic-config/testing/system-tests/pom.xml
myronkscott: i'm fine with that.  Can we do that in a subsequent PR?  UI is not a strength of mine and some discussion on parameter names should be had.

voter/src/main/java/org/terracotta/voter/ClientVoterManagerImpl.java
```diff
@@ -51,31 +51,29 @@ public String getTargetHostPort() {
...
   @Override
-  public long registerVoter(String id) throws TimeoutException {
+  public boolean register(String id) throws TimeoutException {
     String result = processInvocation(diagnostics.invokeWithArg(MBEAN_NAME, "registerVoter", id));
     try {
-      return Long.parseLong(result);
+      generation = Long.parseLong(result);
+      return isRegistered();
     } catch (NumberFormatException ne) {
+      generation = -1L;
```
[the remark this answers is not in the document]
myronkscott: see above

voter/src/main/java/org/terracotta/voter/ClientVoterManagerImpl.java
```diff
@@ -175,6 +173,16 @@ public synchronized boolean isConnected() {
...
   }
+  
+  @Override
+  public boolean isRegistered() {
+    return generation >= 0;
+  }
+  
+  @Override
+  public long generation() {
+    return generation;
```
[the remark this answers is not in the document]
myronkscott: Any state is OK here.  It's just used for comparison in voting so negatives will never get voted for.  They can't anyway since they are no longer connected.

voter/src/main/java/org/terracotta/voter/ClientVoterThread.java
myronkscott: CompletableFuture is better so that we can complete exceptionally if needed.  Will make the call package private

voter/src/main/java/org/terracotta/voter/ClientVoterThread.java
myronkscott: Want to be able to complete exceptionally

voter/src/main/java/org/terracotta/voter/ClientVoterThread.java
myronkscott: This is part of the scheduled operation and outside the scope of operational.  Once the voter is registered, it is operational

myronkscott: I think this is covered in VotingGroup::handleVoteRequest as it checks if the thread is still connected

voter/src/main/java/org/terracotta/voter/VotingGroup.java
myronkscott: sure

voter/src/main/java/org/terracotta/voter/VotingGroup.java
```diff
@@ -0,0 +1,410 @@
...
+      sharedExecutor.awaitTermination(30, TimeUnit.SECONDS);
+    } catch (InterruptedException ie) {
+      LOGGER.info("shutdown interrupted", ie);
+    }
+    this.voter.interrupt();
+  }
+
+  private String[] pollTopology(String[] existingTopo, Function<String, ClientVoterManager> factory, Properties connectionProps) {
+    ClientVoterManager active = getVoteOwner();
+    while (active != null) {
```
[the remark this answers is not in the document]
myronkscott: the only reason this thread should be interrupted is on stop.  on stop we first set vote owner to null.  I don't think we should add too much interrupt handling is it can be avoided.

voter/src/main/java/org/terracotta/voter/ClientVoterThread.java
mathieucarbou: Ok! I didn't follow this part of code.

mathieucarbou (approved):

mathieucarbou:
Do you plan on cherry picking that on the 5.9 branch also after it is merged? It would be nice to have both branches in sync for this part I think.
(Taking as hypothesis that 5.9 is still used in EE 10.7 branch...?)

myronkscott:
i didn't think about that. it should be a pretty clean port. i'll try it.

mathieucarbou:
Anyone else want / need to review ?

myronkscott:
hold off on merging just yet. There is one thing I want to fix plus squash the commits so it is easier to cherry-pick

voter/src/main/java/org/terracotta/voter/VotingGroup.java
```diff
@@ -0,0 +1,433 @@
...
+    return nodes.size();
+  }
+  
+  private synchronized CompletableFuture<?> refreshPollingFuture() {
+    pollingFuture = new CompletableFuture<>();
+    notify();
+    return pollingFuture;
+  }
+  
+  private synchronized CompletableFuture<?> waitForRefresh() {
```
mathieucarbou, line 378: Is it OK for this method to be synchronized ? Since it is called from the public one `forceTopologyUpdate()`, use would go in this while loop waiting, so it will block any call to `refreshPollingFuture()` done from `pollTopology()` (there is a while loop) right ? Or I missed something ? Is it expected ?

mathieucarbou: I was wondering if it is not better to introduce a self-resettable future that will automatically reset itself once completed ? Something like that:

```diff
Index: voter/src/main/java/org/terracotta/voter/VotingGroup.java
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/voter/src/main/java/org/terracotta/voter/VotingGroup.java b/voter/src/main/java/org/terracotta/voter/VotingGroup.java
--- a/voter/src/main/java/org/terracotta/voter/VotingGroup.java	(revision 5065a8f7d0236e8442a7c7330e5c87c1cf6a93d7)
+++ b/voter/src/main/java/org/terracotta/voter/VotingGroup.java	(date 1678720327446)
@@ -15,11 +15,12 @@
  */
 package org.terracotta.voter;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.terracotta.connection.ConnectionException;
+
 import java.util.Arrays;
 import java.util.Collections;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
@@ -35,11 +36,11 @@
 import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
+import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.Consumer;
 import java.util.function.Function;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
-import org.terracotta.connection.ConnectionException;
 
 
 /**
@@ -79,6 +80,8 @@
     this.id = id;
     this.sharedExecutor = Executors.newScheduledThreadPool(hostPorts.length);
     this.voter = voterThread(connectionProps, factory, hostPorts);
+
+    selfResetPollingFuture();
   }
   
   public VoterStatus start() {
@@ -271,7 +274,7 @@
   private String[] pollTopology(String[] existingTopo, Function<String, ClientVoterManager> factory, Properties connectionProps) {
     ClientVoterManager active = getVoteOwner();
     while (active != null) {
-      CompletableFuture<?> pollMark = refreshPollingFuture();
+      CompletableFuture<?> pollMark = pollingFuture.get();
       try {
         Set<String> existingTopology = nodes.keySet();
         Set<String> newTopology = Collections.unmodifiableSet(active.getTopology());
@@ -350,7 +353,7 @@
   // Below is all cruft for testing help
   private final List<Consumer<String>> votingListeners = new CopyOnWriteArrayList<>();
   private final CompletableFuture<?> bootstrapped = new CompletableFuture<>();
-  private CompletableFuture<?> pollingFuture = new CompletableFuture<>();
+  private final AtomicReference<CompletableFuture<?>> pollingFuture = new AtomicReference<>();
   private String[] targets;
   
   public void addVotingListener(Consumer<String> voter) {
@@ -358,7 +361,7 @@
   }
   
   private void fireVotingListeners(String voter) {
-    votingListeners.forEach(c->c.accept(voter));
+    votingListeners.forEach(c -> c.accept(voter));
   }
 
   public Set<String> getConnectedServers() {
@@ -368,30 +371,18 @@
   public int countConnectedServers() {
     return nodes.size();
   }
-  
-  private synchronized CompletableFuture<?> refreshPollingFuture() {
-    pollingFuture = new CompletableFuture<>();
-    notify();
-    return pollingFuture;
-  }
-  
-  private synchronized CompletableFuture<?> waitForRefresh() {
-    CompletableFuture<?> current = pollingFuture;
-    try {
-      while (current == pollingFuture) {
-        notifySleepTimer();
-        this.wait();
-      }
-      return pollingFuture;
-    } catch (InterruptedException ie) {
-      throw new RuntimeException(ie);
-    }
+
+  private void selfResetPollingFuture() {
+    final CompletableFuture<?> old = this.pollingFuture.get();
+    final CompletableFuture<Object> future = new CompletableFuture<>();
+    future.whenComplete((o, throwable) -> selfResetPollingFuture());
+    this.pollingFuture.compareAndSet(old, future);
   }
-  
+
   public CompletableFuture<?> forceTopologyUpdate() {
-    return waitForRefresh();
+    return pollingFuture.get();
   }
-    
+
   private synchronized void setTargets(String[] nodes) {
     targets = nodes;
   }
```

PR #1181, opened 2023-06-28
Introduce "platform-layout" bundle to be use by downstream projects for the kit layout

mathieucarbou:
We need a usable platform kit to do some proper testing manually, across branches or across versions within the scope of the platform project / DC / M&M.

Downstream projects are already building their own scripts and kit layout and depend on all the platform dependencies so they don't need to depend on the platform kit and can continue build their own without this dependency.

## PR CONTENT

- Revert previous commit changing the platform kit
- Introduce a `platform-layout`, which contains only the platform stuff
- The `platform-kit` is made from `core-kit` + `platform-layout`. It is deployed and serves as testing purposing, within the same project or could be used to test backward compatibility in inter-project tests


mathieucarbou:
I added a commit as a workaround of an Angela bug (or requirement ?). Right now, Angela fails if a kit does not contain a client folder.

kit/src/main/assembly/platform-kit.xml
```diff
@@ -30,161 +30,19 @@
...
-      <directory>src/main/kit</directory>
-      <fileMode>0644</fileMode>
+      <directory>${project.build.directory}/terracotta-kit-${terracotta-core.version}/terracotta-${terracotta-core.version}</directory>
       <includes>
-        <include>**/*</include>
+        <include>server/**</include>
       </includes>
       <excludes>
-        <exclude>**/*.sh</exclude>
-        <exclude>**/*.bat</exclude>
```
myronkscott, line 47: These shouldn't be necessary.  They aren't part of cores kit.  If they are showing up there it is a bug.

myronkscott (approved):

PR #1182, opened 2023-06-28
Jackson => Gson

mathieucarbou:
These are some code modifications required for the migration which can be put as an isolated separated PR. They are backward compatible and do not impact anything.


dynamic-config/model/src/main/java/org/terracotta/dynamic_config/api/model/License.java
```diff
@@ -66,22 +74,22 @@ public boolean hasCapability(String capability) {
...
   public String getType() {
-    return flagsMap.entrySet().stream().filter(Map.Entry::getValue).findFirst().get().getKey();
+    return flags.entrySet().stream().filter(Map.Entry::getValue).findFirst().get().getKey();
   }
 
   @Override
   public String toString() {
     return "License{" +
-        "capabilityLimitMap=" + capabilityLimitMap +
+        "capabilityLimitMap=" + capabilities +
```
GaryWKeim, line 90: Not clear why we want to change the field names but leave the old names in the `toString` result.

dynamic-config/model/src/main/java/org/terracotta/dynamic_config/api/model/NodeContext.java
```diff
@@ -30,16 +30,19 @@ public class NodeContext implements Cloneable {
...
+    nodeUID = null;
+  }
 
   public NodeContext(Cluster cluster, UID nodeUID) {
     requireNonNull(nodeUID);
     this.nodeUID = requireNonNull(nodeUID);
     this.cluster = requireNonNull(cluster);
-    this.node = cluster.getNode(nodeUID)
-        .orElseThrow(() -> new IllegalArgumentException("Node UID: " + nodeUID + " not found in cluster: " + cluster.toShapeString()));
-    this.stripe = cluster.getStripeByNode(nodeUID).get();
```
GaryWKeim, line 42: It seems strange to move this validation into an accessor, but if `nodeUID` can change, I guess it makes sense.

diagnostic/client/src/test/java/org/terracotta/diagnostic/client/DiagnosticServiceImplTest.java
```diff
@@ -344,6 +344,11 @@ public static class Beef extends Meat implements Serializable {
     private static final long serialVersionUID = 1L;
     final String quality;
 
+    // For Json
+    private Beef() {
```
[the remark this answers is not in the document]
mathieucarbou: Gson does not work like Jackson. Gson is only field based, reflection based, and required to be able to construct objects. They expect users to either add a default ctor (whatever the visibility), or provide instanciators for all objects.
I prefer private ctors because we are controlling the code, it reduces the code to write in json modules and this is not affecting anything.

https://github.com/google/gson/blob/main/UserGuide.md

dynamic-config/api/src/main/java/org/terracotta/dynamic_config/api/model/nomad/Applicability.java
```diff
@@ -50,4 +52,195 @@ static Applicability node(UID nodeUID) {
...
+    public boolean equals(Object o) {
+      if (this == o) return true;
+      if (o instanceof Applicability
+          && ((Applicability) o).getLevel() == CLUSTER
+          && getLevel() == CLUSTER) {
+        return true;
+      }
+      if (!(o instanceof V2)) return false;
+      V2 that = (V2) o;
+      return getLevel() == that.getLevel() &&
```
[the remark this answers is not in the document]
mathieucarbou: I agree. this code was not created but moved from previous class, so I will reformat it.

dynamic-config/model/src/main/java/org/terracotta/dynamic_config/api/model/License.java
mathieucarbou: Related to my previous explanation on how Gson works ;-)

dynamic-config/model/src/main/java/org/terracotta/dynamic_config/api/model/NodeContext.java
mathieucarbou: This is also due to how Gson works: this class being serialised, this is to avoid having computed fields or transient fields that won't be recovered. 

But you're wright I can improve this and split the validation in the ctor instead.

PR #1183, opened 2023-06-29
Jackson => Gson (2/2)

mathieucarbou:
This PR:

1. removes Jackson from platform and uses Gson instead.
2. Removes any Json-related code from these modules: 
 * Nomad
 * Sanskrit
 * iNet
 * Struct
 * DC model
 * DC API
 * DC Server API

No change in the communication format, test ran against a previous platform kit using Jackson.

Some important notes: 

- Gson works completely differently than Jackson. It is field-based and reflection-based
- Gson is really SIMPLE. It does its best to prevent user from serialising when it knows it won't be able to deserialise. 
- As a consequence, Gson does not support polymorphism
- Gson does also not support class loading and mapping `Class` objects
- Gson also does not support raw / wildcard types since it cannot deserialise
- Gson has no mixin support
- Gson factory order is important
- Gson is strongly types and makes a string difference between declared types (used to select type adapters) and runtime types

All these missing features had to be added. So a lot of code was written to cope with missing Gson features, and also allow class loading and unknown type deserialisation, **through a mechanism where we only specify the type names we allow for security reasons.**

Gson is really easy to use. The `*Module` classes in different modules are stripped down compared to Jackson ones and writing custom type adapters is easier too.


mathieucarbou:
moving to draft - refactoring of how the `GsonModule` works internally.

common/json-support/src/main/java/org/terracotta/json/gson/internal/PathTypeAdapter.java
```diff
@@ -0,0 +1,62 @@
...
+public class PathTypeAdapter extends TypeAdapter<Path> {
+  @Override
+  public String toString() {
+    return PathTypeAdapter.class.getName();
+  }
+
+  @Override
+  public void write(JsonWriter out, Path value) throws IOException {
+    out.beginArray();
+    for (Path segment : value) {
```
GaryWKeim, line 38: Wondering why this is a for-loop.

dynamic-config/model/src/test/java/org/terracotta/dynamic_config/api/service/ClusterFactoryTest.java
```diff
@@ -350,57 +340,6 @@ public void test_toProperties() {
     }));
   }
 
-  @Test
```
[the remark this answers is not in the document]
GaryWKeim: NM: I see that they moved to a different class.

common/json-support/src/main/java/org/terracotta/json/gson/UnsafeObjectTypeAdapterFactory.java
```diff
@@ -0,0 +1,355 @@
...
+import java.util.Optional;
+
+import static java.util.Collections.emptyMap;
+import static java.util.Objects.requireNonNull;
+
+/**
+ * Type adapter that can be used to map a field without knowing its type in advance,
+ * by adding a class information in the serialized json.
+ * <p>
+ * Since deserialization requires to load a class, this adapter is unsafe and should only be
```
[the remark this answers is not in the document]
mathieucarbou: You have an example in `DiagnosticJsonModule`.
Gson is strongly types and does not allow the serialisation of an object if it cannot deserialise. And it uses the declared types to determine how to deserialise. 
But some classes do not declare types (ie. raw collections, raw types like Object or Object[]). 
For such cases, jackson was automatically appending the `@class` information next to the serialised data and was doing a classloader call when deserialising.
Gson does not support this, because is prevents serialising something "unknown", and does not support class loading (one of the reason why there is so few security issues with Gson).
So this class allows to mark some types or annotated fields on mixins to support adding the class information next to the serialised object exactly like Jackson was doing.

common/json-support/src/main/java/org/terracotta/json/gson/internal/HierarchyTypeAdapterFactory.java
```diff
@@ -0,0 +1,317 @@
...
+import static java.util.Objects.requireNonNull;
+import static org.terracotta.json.gson.Adapters.nullSafe;
+
+/**
+ * <pre>
+ * NOTE FROM TERRACOTTA DEVS
+ *
+ * This class has been copied from the Gson project at:
+ * <a href="https://github.com/google/gson/blob/main/extras/src/main/java/com/google/gson/typeadapters/RuntimeTypeAdapterFactory.java">https://github.com/google/gson/blob/main/extras/src/main/java/com/google/gson/typeadapters/RuntimeTypeAdapterFactory.</a>
+ * These extra classes are not available in any Maven repo and are meant as samples to be copied and adapter.
```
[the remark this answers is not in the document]
mathieucarbou: true! that wasn't the case at first but I had to...

management/testing/integration-tests/src/test/java/org/terracotta/management/integration/tests/ServerCacheManagementIT.java
```diff
@@ -123,8 +121,10 @@ public void can_receive_server_statistics() throws Exception {
         .map(o -> o.<Long>getLatestSampleValue("Cluster:HitCount"))
         .anyMatch(counter -> counter.get() == 1L));
 
-    Object expected = readJson("stats.json");
-    queryAllRemoteStatsUntil(stats -> toJson(stats).equals(expected));
+    queryAllRemoteStatsUntil(stats -> {
+      System.out.println(toJson(stats));stats.get(0);
```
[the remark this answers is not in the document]
mathieucarbou: sout removed - was there to debug the incoming data

dynamic-config/server/configuration-repository/src/test/java/org/terracotta/dynamic_config/server/configuration/nomad/persistence/DefaultHashComputerTest.java
```diff
@@ -33,7 +33,7 @@
 public class DefaultHashComputerTest {
```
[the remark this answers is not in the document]
mathieucarbou: Dic and spell check says computer ?
https://www.askdifference.com/computor-vs-computer/

dynamic-config/model/src/test/java/org/terracotta/dynamic_config/api/service/ClusterFactoryTest.java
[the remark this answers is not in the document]
mathieucarbou: yes nothing was removed ;-)

dynamic-config/json/src/main/java/org/terracotta/dynamic_config/api/json/DynamicConfigJsonModule.java
```diff
@@ -0,0 +1,155 @@
...
+        return Applicability.V1.class.getName() + "|" + Applicability.V2.class.getName();
+      }
+
+      @Override
+      public Class<? extends Applicability> apply(JsonObject jsonObject) {
+        return jsonObject.has("scope") ? Applicability.V1.class : Applicability.V2.class;
+      }
+    });
+
+    // These are all the DC classes allowed when serializing Class type ot deserializing @class marker into an object for unknown declared types.
```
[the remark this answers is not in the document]
mathieucarbou: done

common/sanskrit/src/test/java/org/terracotta/persistence/sanskrit/json/TestModule.java
```diff
@@ -15,51 +15,33 @@
...
+  @Override
+  public void configure(GsonBuilder builder) {
+    builder.writeNull(MapSanskritVisitor.Null.class);
+
+    builder.registerSuperType(TestData.Vegie.class)
+        .withSubtype(TestData.Tomato.class)
+        .withSubtype(TestData.Pepper.class);
+
+    builder.registerSuperType(TestData.CookingManual.class)
+        .withSubtype(TestData.CookingManual.class)
```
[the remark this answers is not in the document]
mathieucarbou: ahah! good catch! my mistake!

common/json-support/src/test/java/org/terracotta/json/SerializationTest.java
```diff
@@ -0,0 +1,58 @@
...
+import java.util.Map;
+
+import static org.hamcrest.Matchers.equalTo;
+import static org.hamcrest.Matchers.is;
+import static org.junit.Assert.assertThat;
+
+public class SerializationTest {
+
+  @Test
+  public void test_any_subtyoes() {
```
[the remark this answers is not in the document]
mathieucarbou: fixed

common/json-support/src/main/java/org/terracotta/json/gson/internal/PathTypeAdapter.java
mathieucarbou: How would you do it otherwise ? The goal here is to write all the path segments within an array using the streaming api for Gson.

common/json-support/src/main/java/org/terracotta/json/gson/internal/HierarchyTypeAdapterFactory.java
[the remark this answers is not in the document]
GaryWKeim: I meant that the comment says "copied and adapter"

common/json-support/src/main/java/org/terracotta/json/gson/internal/PathTypeAdapter.java
GaryWKeim: I see, it's `Path.iterator`.. subtle.

common/json-support/src/main/java/org/terracotta/json/gson/internal/HierarchyTypeAdapterFactory.java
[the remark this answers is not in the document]
mathieucarbou: Yeah, a Path is an Iterable, weirdly

common/json-support/src/main/java/org/terracotta/json/gson/internal/PathTypeAdapter.java
mathieucarbou: Yeah, a Path is an Iterable, weirdly

common/json-support/src/main/java/org/terracotta/json/gson/internal/HierarchyTypeAdapterFactory.java
```diff
@@ -0,0 +1,318 @@
+/*
```
chrisdennis, line 1: This needs to have the original Google copyright and license header. We can add ours in addition to cover our additions... but the original needs to be there.

common/json-support/src/main/java/org/terracotta/json/gson/internal/MixinTypeAdapterFactory.java
```diff
@@ -0,0 +1,473 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.TreeMap;
+
+import static java.util.Collections.emptyMap;
+import static java.util.Objects.requireNonNull;
+
+/**
+ * This Gson adapter is highly inspired from
```
chrisdennis, line 65: Probably needs the original Google copyright header to be safe here then.

common/json-support/src/main/java/org/terracotta/json/gson/UnsafeObjectTypeAdapterFactory.java
```diff
@@ -0,0 +1,364 @@
...
+  public String toString() {
+    return "Factory[type=" + declaredTypeToken + ",adapter=" + UnsafeObjectTypeAdapter.class.getName() + "]";
+  }
+
+  private boolean isWildTypeToken() {
+    return Adapters.isWildTypeToken(declaredTypeToken);
+  }
+
+  @SuppressWarnings("unchecked")
+  private <E> JsonElement serializeArray(Gson gson, TypeToken<E> token, Object value) {
```
chrisdennis, line 116: Generic typing seems wrong here... `E` is doing double duty as both the array type and the component type.

common/nomad/src/main/java/org/terracotta/nomad/messages/TakeoverMessage.java
```diff
@@ -18,6 +18,12 @@
 import java.time.Instant;
 
 public class TakeoverMessage extends MutativeMessage {
+
+  // For Json
+  TakeoverMessage() {
```
chrisdennis, line 23: Why... what's this for, and how is GSon initializing the fields? Reflection?

diagnostic/common/src/main/java/org/terracotta/diagnostic/common/json/DiagnosticJsonModule.java
```diff
@@ -15,74 +15,25 @@
...
-    super(DiagnosticJsonModule.class.getSimpleName(), new Version(1, 0, 0, null, null, null));
-
-    setMixInAnnotation(DiagnosticRequest.class, DiagnosticRequestMixin.class);
-    setMixInAnnotation(DiagnosticResponse.class, DiagnosticResponseMixin.class);
+public class DiagnosticJsonModule implements GsonModule {
+  @Override
+  public void configure(GsonConfig config) {
+    config.registerUnsafeTypeAdapters(Object[].class);
+    config.allowClassLoading(Object[].class);
+    config.registerMixin(DiagnosticResponse.class, DiagnosticResponseMixin.class);
```
chrisdennis, line 32: What happened to the `DiagnosticRequestMixin`?

dynamic-config/testing/system-tests/src/test/resources/logback-ext-test.xml
```diff
@@ -17,8 +17,9 @@
 
 -->
 <included>
-  <!--  <logger name="org.terracotta.dynamic_config" level="TRACE"/>-->
-  <!--  <logger name="org.terracotta.persistence.sanskrit" level="TRACE"/>-->
-  <!--  <logger name="org.terracotta.nomad" level="TRACE"/>-->
+    <logger name="org.terracotta.dynamic_config" level="TRACE"/>
```
chrisdennis, line 20: Did you intend to commit these?

dynamic-config/testing/system-tests/src/test/resources/logback-test.xml
```diff
@@ -37,16 +37,16 @@
     <appender-ref ref="STDOUT"/>
   </root>
 
-  <logger name="org.terracotta.dynamic_config" level="INFO" additivity="false">
+  <logger name="org.terracotta.dynamic_config" level="TRACE" additivity="false">
```
chrisdennis, line 40: Did you intend to commit these?

pom.xml
```diff
@@ -235,11 +234,9 @@
...
       </dependency>
       <dependency>
-        <groupId>com.fasterxml.jackson</groupId>
-        <artifactId>jackson-bom</artifactId>
-        <version>${jackson.version}</version>
-        <scope>import</scope>
-        <type>pom</type>
+        <groupId>com.google.code.gson</groupId>
+        <artifactId>gson</artifactId>
+        <version>2.10.1</version>
```
chrisdennis, line 239: Should GSon get a version property?

chrisdennis (changes requested):

common/json-support/src/main/java/org/terracotta/json/gson/internal/HierarchyTypeAdapterFactory.java
mathieucarbou: Ok!

common/json-support/src/main/java/org/terracotta/json/gson/internal/MixinTypeAdapterFactory.java
mathieucarbou: Ok!
Note: This one is a pretty big adaptation of the original class.

common/json-support/src/main/java/org/terracotta/json/gson/internal/NullTypeAdapter.java
```diff
@@ -0,0 +1,39 @@
...
+
+public class NullTypeAdapter<T> extends TypeAdapter<T> {
+  @Override
+  public String toString() {
+    return getClass().getName();
+  }
+
+  @Override
+  public void write(JsonWriter out, T value) throws IOException {
+    out.jsonValue("null");
```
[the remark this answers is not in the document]
mathieucarbou: No because the value will of type Class<T> ( a type we want to use as a way to force null)

common/json-support/src/main/java/org/terracotta/json/gson/internal/HierarchyTypeAdapterFactory.java
mathieucarbou: I will add their license in another class too (runtime type adapter).

common/json-support/src/main/java/org/terracotta/json/gson/Adapters.java
```diff
@@ -0,0 +1,454 @@
...
+      ReflectionAccessFilter.BLOCK_INACCESSIBLE_JAVA,
+      ReflectionAccessFilter.BLOCK_ALL_ANDROID,
+      ReflectionAccessFilter.BLOCK_ALL_PLATFORM
+  ));
+
+  public static final ToNumberStrategy NUMBER_STRATEGY = in -> {
+    String value = in.nextString();
+    if (value.contains(".") || SPECIAL_NUMBERS.contains(value)) {
+      try {
+        float n = Float.parseFloat(value);
```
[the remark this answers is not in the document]
mathieucarbou: I agree with you... This is something all parsers do : they make a choice. I will look but if I remember I am reproducing here what Gonna or Jackson is doing. 
There might be a way to avoid that with the lazy number class of Gson. I will check.

common/json-support/src/main/java/org/terracotta/json/gson/Adapters.java
```diff
@@ -0,0 +1,454 @@
...
+
+    @Override
+    public <T> TypeAdapter<T> create(Gson gson, TypeToken<T> type) {
+      final Class<? super T> rawType = type.getRawType();
+      if (rawType.isPrimitive()
+          || rawType.isArray()
+          || Number.class.isAssignableFrom(rawType)
+          || String.class == type.getRawType()
+          || Boolean.class == type.getRawType()
+          || Collection.class.isAssignableFrom(type.getRawType())
```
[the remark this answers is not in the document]
mathieucarbou: ah! missed that one :P

common/json-support/src/main/java/org/terracotta/json/gson/UnsafeObjectTypeAdapterFactory.java
mathieucarbou: Crap yes there is a mistake. 
E is the array element type. So it should be `TypeToken<?> token` instead.

common/nomad/src/main/java/org/terracotta/nomad/messages/TakeoverMessage.java
mathieucarbou: Yeah, see my comments in the PR... 

Gson being totally field based and reflection based, for classes we control, it requires to be able to call a no-arg ctor, or to provide some class instanciators, which are sort of factories were we call the existing ctors with dummy values and gson takes the instance and then replace the field values. Which is as ugly and requires more code than just adding a private ctor to our classes like that...

That's one of the little price to pay for using Gson (one or the other solution above). I prefer setting private ctors because for some cases this is even not possible to instantiate with dummy values.

diagnostic/common/src/main/java/org/terracotta/diagnostic/common/json/DiagnosticJsonModule.java
mathieucarbou: We don't need anymore, like a bunch of other mixins.

`registerUnsafeTypeAdapters(Object[].class)` makes sure that Gson wraps the unknown objects inside with their runtime type next to the serialised form.

dynamic-config/testing/system-tests/src/test/resources/logback-ext-test.xml
mathieucarbou: ouch no!

dynamic-config/testing/system-tests/src/test/resources/logback-test.xml
mathieucarbou: no! forgot them! debugging purposes...

pom.xml
mathieucarbou: So I would say no, because this is 1 lib only and like all other lib, when only 1 lib, we don't have a property.

common/json-support/src/main/java/org/terracotta/json/gson/Adapters.java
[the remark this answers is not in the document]
mathieucarbou: Ok so what I said is half-true for the non decimal values and false for decimal values. So I will change this parser and grab also your suggestion below for the integers and longs.

```
import com.fasterxml.jackson.databind.ObjectMapper
ObjectMapper json = new ObjectMapper()
println json.readValue(json.writeValueAsString([n: Float.MAX_VALUE]), Map).n.class // Double
println json.readValue(json.writeValueAsString([n: Double.MAX_VALUE]), Map).n.class // Double
println json.readValue(json.writeValueAsString([n: 0.1f]), Map).n.class // Double
println json.readValue(json.writeValueAsString([n: 0.1d]), Map).n.class // Double
println json.readValue(json.writeValueAsString([n: Byte.MAX_VALUE]), Map).n.class // Integer
println json.readValue(json.writeValueAsString([n: Short.MAX_VALUE]), Map).n.class // Integer
println json.readValue(json.writeValueAsString([n: Integer.MAX_VALUE]), Map).n.class // Integer
println json.readValue(json.writeValueAsString([n: Long.MAX_VALUE]), Map).n.class // Long
println json.readValue(json.writeValueAsString([n: Long.MAX_VALUE]), Map).n.class // Long
```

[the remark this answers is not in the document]
mathieucarbou: Note:  we currently don't have use case were we serialise floats and have to deserialise to floats. I even think we don't even have any decimal numbers going through json currently, but only ints or longs.

I am changing the code to copy the Jackon behviour, but if we think we want to change this behaviour for floats, in order to use floats, this would be possible and won't break backward compatibility.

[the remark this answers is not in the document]
mathieucarbou: Also I am noticing that the plain value notation we use (at least in TMC) is wrong because 1.0 is serialized as 1 and then would be deserialized as an int. 1.0f or 1.0d should be serialized as 1.0.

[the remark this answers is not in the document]
mathieucarbou: ok so I've pushed a better version, with also a change in `FloatingPointTypeAdapter`

common/json-support/src/main/java/org/terracotta/json/DefaultJsonFactory.java
```diff
@@ -68,211 +83,209 @@ public DefaultJsonFactory pretty() {
...
 
-    @Override
-    public Object parse(String json, Type type) {
-      try {
-        return mapper.readValue(json, mapper.constructType(type));
-      } catch (JsonProcessingException e) {
-        throw new UncheckedIOException(e);
-      }
-    }
+  private static class DirectedGraph<T> {
```
chrisdennis, line 252: I was struggling to understand this code in the context of a directed graph, or its usage. I tried to rewrite things in this area just to test my understanding and ended up with the following:
```java
private static class DirectedGraph<T> {
  private final Map<T, Collection<T>> children = new LinkedHashMap<>();

  public void addEdge(T from, T to) {
    children.computeIfAbsent(from, k -> new LinkedHashSet<>()).add(to);
  }

  public Stream<T> depthFirstTraversal(final T root) {
    return depthFirstTraversal(root, new HashSet<>(children.size()));
  }

  private Stream<T> depthFirstTraversal(final T current, final Set<T> visited) {
    if (visited.add(current)) {
      return Stream.concat(children.getOrDefault(current, emptyList()).stream().flatMap(child -> depthFirstTraversal(child, visited)), Stream.of(current));
    } else {
      return empty();
    }
  }

  @Override
  public String toString() {
    return " - " + children.entrySet().stream().map(Object::toString).collect(joining(lineSeparator() + " - "));
  }
}
```
and then driving it:
```java
final DirectedGraph<Class<? extends Module>> graph = new DirectedGraph<>();
resolvedModules.keySet().forEach(type -> {
  graph.addEdge(RootModule.class, type);
  // we only override if the module to override has to be loaded.
  getOverrides(type).filter(resolvedModules::containsKey).forEach(overridden -> graph.addEdge(type, overridden));
  getDependencies(type).forEach(dependency -> graph.addEdge(type, dependency));
});

LOGGER.trace("Graph:{}{}", lineSeparator(), graph);

return graph.depthFirstTraversal(RootModule.class).map(resolvedModules::get).collect(toList());
```
This behaves the same as the original code with one small change. In the test where `C` overrides `A` we end up configuring `C` after `A` instead of before. This seems more logical to me... but I don't know enough about Gson to know if it's correct:
```java
@Test
public void test_override() {
  CALLS.clear();
  new DefaultJsonFactory().withModule(new B()).withModule(new C()).create();
  assertThat(CALLS, is(equalTo(asList("A", "B", "C"))));
}
```

mathieucarbou: @chrisdennis : this is the code I wrote first (depth first traversal) but it did work because the resulting list is not correctly sorted. The sorting algorithm is a little bit different.

It is possible to compare the output of traces when you put slf4j simple in trace mode.

@Dependency is just to declare modules that are required. So logically registered before but this is not required.

**@Override** is to make sure a module gets **configured before** another one.

This is linked to the documentation I've put in GsonFactory class. Gson has an internal list of factories, and when searching for a type adapter, it takes the first one it finds (there is one exception though when searching for the next factory, skipping the first one found).

So the logic is that in Gson, things having a higher priority should come first.

So that is why the sorting makes sure that overriding modules are configured first (in this case `C`) and then other ones.

chrisdennis: If we just reverse the edge between `overridden` and `type` won't that get us where we want to be then?

mathieucarbou: @chrisdennis I ran a test and I cannot reproduce what I saw with your code. I might have made a mistake in my DFT code, or, I might have put the edges in the wrong order.

See:

```java
public class Main {
  public static void main(String[] args) {
    // A->B means:
    // - A is a dependency of B
    // - or A overrides B

    // A->B->C
    // A->D->C
    // D->E
    // F->A

    {
      final DefaultJsonFactory.DirectedGraph<String> graph = new DefaultJsonFactory.DirectedGraph<>();
      Stream.of("A", "B", "C", "D", "E", "F", "G").forEach(v -> {
        graph.addVertex(v);
        graph.addEdge("R", v);
      });

      graph.addEdge("A", "B");
      graph.addEdge("A", "D");
      graph.addEdge("B", "C");
      graph.addEdge("D", "C");
      graph.addEdge("D", "E");
      graph.addEdge("F", "A");

      System.out.println("SORT: " + graph.sort("R"));
      // [R, G, F, A, D, E, B, C] => OK
    }

    {
      final DefaultJsonFactory.DirectedGraph<String> graph = new DefaultJsonFactory.DirectedGraph<>();
      Stream.of("A", "B", "C", "D", "E", "F", "G").forEach(v -> {
        graph.addVertex(v);
        graph.addEdge("R", v);
      });

      graph.addEdge("A", "B");
      graph.addEdge("A", "D");
      graph.addEdge("B", "C");
      graph.addEdge("D", "C");
      graph.addEdge("D", "E");
      graph.addEdge("F", "A");

      final List<String> list = graph.depthFirstTraversal("R").collect(toList());
      Collections.reverse(list);
      System.out.println("DFT: " + list);
      // [R, G, F, A, D, E, B, C] => OK
    }
  }
}
```

I will update the code with yours and run the tests again... Your implementation is cleaner ;-)

mathieucarbou: Pushed a few changes :-)

common/json-support/src/main/java/org/terracotta/json/util/DirectedGraph.java
```diff
@@ -0,0 +1,58 @@
...
+import java.util.Set;
+import java.util.stream.Stream;
+
+import static java.lang.System.lineSeparator;
+import static java.util.Collections.emptyList;
+import static java.util.stream.Collectors.joining;
+import static java.util.stream.Stream.empty;
+
+public class DirectedGraph<T> {
+  private final Map<T, Collection<T>> vertexEdges = new LinkedHashMap<>();
```
chrisdennis, line 32: I think this can be a regular `HashMap`

chrisdennis (approved):

common/json-support/src/main/java/org/terracotta/json/util/DirectedGraph.java
mathieucarbou: I wanted to preserve user ordering when a user declares 2 dependencies on a module

PR #1200, opened 2025-01-30
new core

myronkscott:
These are follow on changes to adopt new core once it is merged.  Terracotta-OSS/terracotta-core#1355


common/sanskrit/build.gradle
```diff
@@ -5,8 +5,9 @@ plugins {
 }
 
 dependencies {
-  implementation "org.terracotta:terracotta-utilities-tools:$terracottaUtilitiesToolsVersion"
+  compileOnly "org.terracotta:terracotta-utilities-tools:$terracottaUtilitiesToolsVersion"
```
mathieucarbou, line 8: why ?

communicator/client/build.gradle
```diff
@@ -5,7 +5,7 @@ plugins {
 }
 
 dependencies {
-  api "org.terracotta:entity-client-api:$terracottaApisVersion"
+  compileOnly "org.terracotta:client-api:$terracottaApisVersion"
```
mathieucarbou, line 8: why ? this is unrelated to voltron plugin and a mere library with dependency.

communicator/common/build.gradle
```diff
@@ -5,7 +5,7 @@ plugins {
 }
 
 dependencies {
-  api "org.terracotta:entity-common-api:$terracottaApisVersion"
+  compileOnly "org.terracotta:common-api:$terracottaApisVersion"
```
mathieucarbou, line 8: why ? this is unrelated to voltron plugin and a mere library with dependency.

communicator/server/build.gradle
```diff
@@ -6,7 +6,7 @@ plugins {
 
 dependencies {
   api project(':communicator:common')
-  api "org.terracotta:standard-cluster-services:$terracottaApisVersion"
+  compileOnly "org.terracotta:server-api:$terracottaApisVersion"
```
mathieucarbou, line 9: why ? this is unrelated to voltron plugin and a mere library with dependency.

management/testing/integration-tests/build.gradle
```diff
@@ -1,21 +1,48 @@
...
+  pluginsImpl project(':dynamic-config:server:config-provider')
+  pluginsImpl project(':dynamic-config:server:services')
+  pluginsImpl project(':management:testing:entity')
+  pluginsImpl project(':management:entities:nms:server')
+  pluginsImpl project(':management:entities:nms-agent:server')
+  pluginsImpl project(':management:server:services')
+  pluginsImpl project(':lease:server')
+  pluginsImpl project(':client-message-tracker')
+  pluginsImpl project(':diagnostic:server:services')
+  pluginsImpl project(':platform-base')
```
mathieucarbou, line 45: Why would management need to know all these ? 
This will create a potential duplication in all integration tests like that ?
management just need to know about a test kit where it adds its own stuff.

mathieucarbou (changes requested):
I kind of disagree with the compileOnly change for a normal library.

Maybe this can be discussed with Chris ?

A normal library should declare its transitive api dependencies and implementation dependencies as usual.

If this causes issues at some point at other places, the issue has to be solved where the library is used.

Otherwise what's the point of using a dependency management system ? And let's declare everything compileOnly.

common/sanskrit/build.gradle
myronkscott: utilities was getting pulled into both the api and implementation directories.  This prevents it from getting pulled into the implant directory.  It's brought into api via dynamic-config:api

communicator/client/build.gradle
myronkscott: client-api doesn't mean anything without the implementation and shouldn't be part of the plugin's api.  Separating the api from the runtime allows for updating the container without having to recompile and redeploy the plugin.  Final packaging is the runtime + plugins which makes pulling in the api separately unwanted.

communicator/server/build.gradle
myronkscott: server's api is not part of the plugin's api.

management/testing/integration-tests/build.gradle
myronkscott: This is how the "kit" is built now for galvan tests.  Galvan provides the server.  Consumer tells galvan what plugins go were and the "kit" is built dynamically.  For the tests.  Some effort has been made to reduce the impact of building multiple kits.

myronkscott:
> I kind of disagree with the compileOnly change for a normal library.
> 
> Maybe this can be discussed with Chris ?
> 
> A normal library should declare its transitive api dependencies and implementation dependencies as usual.
> 
> If this causes issues at some point at other places, the issue has to be solved where the library is used.
> 
> Otherwise what's the point of using a dependency management system ? And let's declare everything compileOnly.

My view is that plugins are not normal java libraries.  They only work and exist in the context of a container (passthrough, live server).  It makes sense to compile against the api of the container and not worry about the runtime implementation.  It is up to the container to maintain the api and work as documented.  This allows building of a kit or server or client solely by combining the plugins with the appropriate runtime as long as the api is compatible.  This is why core apis are compileOnly or compileApi

communicator/client/build.gradle
mathieucarbou: It is not a plugin here: this is a library used client-side, like a normal maven lib with transitive dependencies.
If you have an XML lib, it depends on jaxb api. Whatever the jaxb impl is used by the lib user.
Same thing here.

communicator/common/build.gradle
mathieucarbou: So this is a packaging concern - and the place to fix that is at packaging level - libraries should not be impacted by how they are used... common here requires common-api in api scope because this module exposes common-api classes in its api.

This change put the burden on the consumer of this lib to know how it works.

If packaging has special requirements, then they have to be handled in the build.gradle that is doing the packaging.

And probably that in this build.gradle script you can put in common all required exclusions / replacements globally, instead of doing that per library and impacting each one.

communicator/server/build.gradle
mathieucarbou: yes but same answer: this lib is a normal lib: this is not a voltron plugin here ,so this lib uses and exposes some objects from its dependencies in API scope, as defined by gradle. Consumers of such lib should have it in their project.

Now, if there is a packaging concern, it has to be handled where the packaging is done. But the packaging concerns should not be handled directly in each single library of platform project...

If I develop an app using jaxb api, jmx api or any java or j2ee api, the api is put in api scope. 
Even for java ee objects deployed server-side.

The packaging job should be responsible to know how the project is packaged, which container, etc and then exclude or replace what's needed.

management/testing/integration-tests/build.gradle
mathieucarbou: I agree, but we have a test kit in platform specially for that. 
All these things should be in platform test kit, and management and DC just need to depend on the test kit.
They don't need to know how the test kit is built.

communicator/server/build.gradle
myronkscott: We can change it to compileOnlyApi but I still think that is an incorrect use of the dependency chain.  The concern is that this library which is part of a plugin should not but exporting as api an api it is consuming.

communicator/common/build.gradle
myronkscott: client-api and server-api already include common so this common implementation should not need to export the api it consumes.

communicator/client/build.gradle
myronkscott: this is not a normal library if it requires the apis.  It can only be used in the context of the container implementation.

common/sanskrit/build.gradle
myronkscott: I will look into refactoring the api libs to try and fix this but there are many interdependencies here and this was the quickest fix.

management/testing/integration-tests/build.gradle
myronkscott: the goal here is to not require kits for testing purposes.  Passthrough and galvan's purpose is to facilitate plugin testing.  We can use Angela for final delivery kit testing in ehcache and EE

myronkscott: core no longer provides a kit and galvan knows on its own how to start inline and process servers.  Scripts and README and documentation are no longer the responsibility of core and belong solely in the final kit delivery packaging.

communicator/client/build.gradle
myronkscott: https://blog.gradle.org/introducing-compile-only-dependencies

management/server/api/build.gradle
```diff
@@ -2,12 +2,12 @@ plugins {
...
   id 'org.terracotta.build.convention.deploy'
   id 'biz.aQute.bnd.builder'
+  id 'org.terracotta.build.convention.voltron-server'
 }
 
 dependencies {
   api project(':management:registry')
-  api "org.terracotta:entity-server-api:$terracottaApisVersion"
-  api "org.terracotta:standard-cluster-services:$terracottaApisVersion"
+  api project(':management:model')
```
mathieucarbou, line 10: should come transitively from the one above

dynamic-config/api/build.gradle
```diff
@@ -2,14 +2,15 @@ plugins {
...
+  id 'org.terracotta.build.convention.voltron-server'
 }
 
 dependencies {
   api project(':dynamic-config:model')
   api project(':diagnostic:model')
   api project(':common:nomad')
 
-  implementation "org.terracotta:terracotta-utilities-tools:$terracottaUtilitiesToolsVersion"
+  api ("org.terracotta:terracotta-utilities-tools:$terracottaUtilitiesToolsVersion")
```
mathieucarbou, line 13: why ? not sure dc api exposes anything from utilities. It just uses it.

management/entities/nms-agent/client/src/main/java/org/terracotta/management/entity/nms/agent/client/DefaultNmsAgentService.java
```diff
@@ -70,7 +70,7 @@ public class DefaultNmsAgentService implements EndpointListener, MessageListener
   private long timeoutMs = 5000;
   private Executor managementCallExecutor = Runnable::run;
   private final ManagementProvider<?> diagnosticProvider = new DiagnosticProvider(DiagnosticUtility.class);
-  private BiConsumer<Operation, Throwable> onOperationError = (op, err) -> LOGGER.trace("Failed to call management entity. Message will be lost. Error: {}", err.getMessage(), err);
+  private BiConsumer<Operation, Throwable> onOperationError = (op, err) -> LOGGER.error("Failed to call management entity. Message will be lost. Error: {}", err.getMessage(), err);
```
mathieucarbou, line 73: this can flood the logs...

mathieucarbou (commented):
a few comments

management/entities/nms-agent/client/src/main/java/org/terracotta/management/entity/nms/agent/client/DefaultNmsAgentService.java
myronkscott: Trace hid an error in the proxy implementation that was very difficult to track down since this was not appearing in the logs.  This shouldn't happen unless there is a serious build problem correct?  This can't be caused by anything environmental can it?

management/entities/nms-agent/common/build.gradle
```diff
@@ -5,7 +5,7 @@ plugins {
 }
 
 dependencies {
-  api project(':management:model')
+  implementation project(':management:model')
```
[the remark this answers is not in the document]
myronkscott: :management:server:api exposes this as part if its service (plugins/api).

management/server/api/build.gradle
myronkscott: removed

management/server/services/build.gradle
```diff
@@ -1,14 +1,14 @@
...
   id 'org.terracotta.build.convention.java-library'
   id 'org.terracotta.build.convention.deploy'
-  id 'org.terracotta.build.convention.voltron'
+  id 'org.terracotta.build.convention.voltron-server'
 }
 
 dependencies {
   service project(':management:server:api')
   service project(':dynamic-config:api')
-  service project(':common:json')
```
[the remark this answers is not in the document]
myronkscott: reverted

platform-layout/build.gradle.kts
```diff
@@ -74,31 +72,19 @@ val serverPluginLibs = dependencies.extensions.create<ServerPluginExtension>("se
...
-      }
-    }
-  }
+  serverLibs.name("org.terracotta.internal:server-runtime:$terracottaRuntimeVersion")
 
   // voltron server API
   serverPluginApis(project(":common:json"))
   serverPluginApis(project(":diagnostic:server:api"))
   serverPluginApis(project(":dynamic-config:server:api"))
+  serverPluginApis(project(":dynamic-config:json"))
```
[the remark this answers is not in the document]
myronkscott: reverted but it seems strange that common:json is a service api and dynamic-config json is not

settings.gradle
```diff
@@ -125,4 +125,21 @@ include ':common:test-utilities',
...
+            substitute(module("org.terracotta:common-api")).using(project(":common-api"))
+            substitute(module("org.terracotta.internal:client-runtime")).using(project(":client-runtime"))
+            substitute(module("org.terracotta.internal:server-runtime")).using(project(":server-runtime"))
+            substitute(module("org.terracotta:passthrough-server")).using(project(":passthrough-server"))
+            substitute(module("org.terracotta:galvan")).using(project(":galvan"))
+            substitute(module("org.terracotta.internal:test-interfaces")).using(project(":test-interfaces"))
+            substitute(module("org.terracotta.internal:galvan-support")).using(project(":galvan-support"))
+        }
+    }
+}
```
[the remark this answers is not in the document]
myronkscott: This is more a preview of a merged repository.  we can possibly do something with gradle init scripts but this is closer to what Gary was asking for and I guess Chris too with one mega mono repo.   I'm fine if we want to take it out of here.

dynamic-config/api/build.gradle
mathieucarbou: In that case this is data roots that should depend on it as implementation. 

I even think that more generally, terracotta-utilities-tools should in theory always be in implementation scope since these are only utilities.

management/entities/nms-agent/common/build.gradle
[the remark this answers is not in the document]
mathieucarbou: so this is api then ? all what is in plugins/api is api.

and a client library using nms-agent entity client needs model to come transitively.

settings.gradle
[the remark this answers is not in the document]
myronkscott: I have removed this section.  Will revisit later

dynamic-config/api/build.gradle
myronkscott: the problem is that API here depends on it so it live in the plugins/api directory.  If data-roots and others depend on it as implementation, then they will want to pull it into the plugins/lib directory as well.  That's not desirable.  You can say that the file exclusion in kit building will fix this but it's cleaner to not pull it at all and allow the gradle mechanisms to put it in the correct place.

dynamic-config/entities/nomad/client/build.gradle
```diff
@@ -2,11 +2,13 @@ plugins {
...
   id 'org.terracotta.build.convention.deploy'
   id 'biz.aQute.bnd.builder'
+  id 'org.terracotta.build.convention.voltron-client'
 }
 
 dependencies {
   api project(':dynamic-config:entities:nomad:common')
-  api "org.terracotta:entity-client-api:$terracottaApisVersion"
+  // above has below as compile only
+  runtimeOnly project(':common:nomad')
```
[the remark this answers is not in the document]
myronkscott: I moved client to api and common to compileOnly and server to service.  This satisfies the correct dependency in each domain.

management/entities/nms-agent/common/build.gradle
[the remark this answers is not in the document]
myronkscott: following the model of dynamic-config common has moved to compileOnly, client exposes api and server uses service dependency

dynamic-config/cli/api/src/main/java/org/terracotta/dynamic_config/cli/api/restart/RestartService.java
```diff
@@ -191,10 +191,10 @@ public Map<Node.Endpoint, Exception> getErrors() {
   }
 
   private void shutdown(ExecutorService executorService) {
-    executorService.shutdownNow();
+    executorService.shutdown();
```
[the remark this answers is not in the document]
myronkscott: JDK bug.  calling both

myronkscott:
Addressed every comment.  Please review and reopen any resolved conversation if more discussion is required.

mathieucarbou-ibm:
> > I kind of disagree with the compileOnly change for a normal library.
> > Maybe this can be discussed with Chris ?
> > A normal library should declare its transitive api dependencies and implementation dependencies as usual.
> > If this causes issues at some point at other places, the issue has to be solved where the library is used.
> > Otherwise what's the point of using a dependency management system ? And let's declare everything compileOnly.
> 
> My view is that plugins are not normal java libraries. They only work and exist in the context of a container (passthrough, live server). It makes sense to compile against the api of the container and not worry about the runtime implementation. It is up to the container to maintain the api and work as documented. This allows building of a kit or server or client solely by combining the plugins with the appropriate runtime as long as the api is compatible. This is why core apis are compileOnly or compileOnlyApi

My comment was referring to other left overs impacting **client side dependencies and their transitive dependencies**.

myronkscott:
Besides, dyanmic-config api is most above else a common and normal library used both client side and server side. So I would expect that any app that depends on this module correctly sees its transitive graph, regardless of what magic you need to do in the scripts to build the kit...

That's fine.  I switched from compileOnly to api...Are you ok with that?

diagnostic/server/services/build.gradle
```diff
@@ -1,11 +1,10 @@
 plugins {
   id 'org.terracotta.build.convention.java-library'
   id 'org.terracotta.build.convention.deploy'
-  id 'org.terracotta.build.convention.voltron'
+  id 'org.terracotta.build.convention.voltron-server'
 }
 
 dependencies {
-  service project(':common:json')
```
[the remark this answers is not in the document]
myronkscott: I added it back

PR #1202, opened 2025-02-18
Update Maven coordinates

mathieucarbou:
This PR updates all maven coordinates as is:

- everything under `org.terracotta`
- all artifact names prefixed by `terracotta-*`
- artifact names match the folder structure

This helps organizing our libraries in the kit / layout and avoids naming conflicts.

image
image

### DIFF
--- a/voter/build.gradle
+++ b/voter/build.gradle
@@ -25,6 +25,6 @@ tool {
 
 deploy {
   groupId = 'org.terracotta'
-  artifactId = 'voter'
+  artifactId = 'terracotta-voter'
   name = 'Voter CLI'
 }
--- a/client-message-tracker/build.gradle
+++ b/client-message-tracker/build.gradle
@@ -6,6 +6,6 @@ plugins {
 
 deploy {
   groupId = 'org.terracotta'
-  artifactId = 'client-message-tracker'
+  artifactId = 'terracotta-client-message-tracker'
   name = 'Client Message Tracker Server Plugin'
 }
--- a/common/inet/build.gradle
+++ b/common/inet/build.gradle
@@ -9,7 +9,7 @@ dependencies {
 }
 
 deploy {
-  groupId = 'org.terracotta.common'
-  artifactId = 'common-inet-support'
+  groupId = 'org.terracotta'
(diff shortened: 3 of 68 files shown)


myronkscott (commented):
I'm fine with this but I'd like someone with more expertise in this to approve.

mathieucarbou:
@chrisdennis : can you have a quick look please ?
We'd like to merge that so that I can then look to deploy a release or at least start an Ehcache PR to integrate the changes, and then do the same for EE.
This will unlock also core because Myron needs to test core in downstream projects also.
Thanks!

chrisdennis (commented):
I understand everything this does... I'm just not sure I fully understand the why here. This seems to move every single artifact to new coordinates. I guess to figure out the motivation I might need to know:

* What collisions were we seeing?
* What's the advantage of a single flat groupId?

mathieucarbou:
> I understand everything this does... I'm just not sure I fully understand the why here. This seems to move every single artifact to new coordinates. I guess to figure out the motivation I might need to know:

Yes, since some artifacts changed during the switch,, we talked in previous meeting to take the opportunity to also review / fix some of the artefact names we have: some are not meaningful, some are weird, some are too long or with prefixes.

This cleans up all that.

> * What collisions were we seeing?

none

> * What's the advantage of a single flat groupId?

you should ask the opposite question: what's the advantage of using groupid ? => none because it just create a bunch of sub folders and it does not remove the fact that the artifact name has better be unique enough to avoid collisions during copies or just review in a list of lib within others to easily recognise it. If you have 10 artifacts called client.jar, it does not help

voter/build.gradle
```diff
@@ -25,6 +25,6 @@ tool {
 
 deploy {
   groupId = 'org.terracotta'
-  artifactId = 'voter'
+  artifactId = 'terracotta-voter'
```
chrisdennis, line 28: This one is the only one I can imagine might have external impacts.

chrisdennis (approved):
Approved... but should we think about rolling back the voter change if we want to use this in maintenance lines?

PR #1214, opened 2025-08-18
Switched from Nexus to Artifactory for deployment

mathieucarbou:
Switched from Nexus to Artifactory for deployment.

Tested with:

`gw --stacktrace --info publish`

and

`gw --stacktrace --info :platform-layout:publish`

to verify that the layout distribution is also deployed.


mathieucarbou:
> Do we care about test failures? If not I will merge.

no, we can merge

PR #1220, opened 2025-09-17
Adding support for line continuation

tmesic99:
Prior to this change, the newer .cfg file format did not support line continuation, which is supported with the original .properties format.  This deficiency was noticed whenever a tc-properties setting possessed several entries and everything had to be entered on a single line, which made it difficult to clearly identify each property.  This change now allows each property to be continued on separate line.

### DIFF
--- a/dynamic-config/model/src/main/java/org/terracotta/dynamic_config/api/service/ConfigPropertiesTranslator.java
+++ b/dynamic-config/model/src/main/java/org/terracotta/dynamic_config/api/service/ConfigPropertiesTranslator.java
@@ -27,6 +27,7 @@ import java.io.Reader;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.util.Iterator;
 import java.util.LinkedHashMap;
 import java.util.LinkedHashSet;
 import java.util.List;
@@ -111,19 +112,34 @@ public class ConfigPropertiesTranslator {
    * Warning: caller who created the stream is also responsible for closing it
    */
   public Properties convert(Reader stream) throws IOException {
-
     // Read each line in the file stream, ignoring all line comments
     // Line format is: [[stripe|node :]<stripe_or_node_name>:]<setting>=<value>
-    // splits happen at '='
+    // splits only happen at '='
+    // line continuations are supported
 
     errors.clear();
     inputConfigSettings.clear();
     try (BufferedReader reader = new BufferedReader(stream)) {
-      reader.lines()
-          .map(String::trim)
-          .filter(line -> !line.isEmpty() && !line.startsWith("#") && !line.startsWith("!"))
-          .map(line -> line.split(EQUALS))
-          .forEach(splits -> inputConfigSettings.put(splits[0].trim(), splits.length > 1 ? splits[1].trim() : ""));
+      List<String> lines = reader.lines()
+        .filter(line -> {
+           line = line.trim();
+           return !line.isEmpty() && !line.startsWith("#") && !line.startsWith("!");
+        })
+        .collect(Collectors.toList());
+      Iterator<String> itr = lines.iterator();
+      String line = "";
+      while (itr.hasNext()) {
+        String thisLine = itr.next();
+        int continuationIndex = thisLine.lastIndexOf('\\');
+        if (continuationIndex == -1) {
+          line += thisLine;
+          String[] splits = line.split("=");
+          inputConfigSettings.put(splits[0].trim(), splits.length > 1 ? splits[1].trim() : "");
+          line = "";
+        } else {
+          line += thisLine.substring(0, continuationIndex);
+        }
+      }
     } catch (Exception ex) {
       throw errorsWith(ex.getMessage());
     }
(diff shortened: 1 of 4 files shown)


dynamic-config/model/src/main/java/org/terracotta/dynamic_config/api/service/ConfigPropertiesTranslator.java
```diff
@@ -111,19 +112,32 @@ public Properties load(Path configFile) {
...
     errors.clear();
     inputConfigSettings.clear();
     try (BufferedReader reader = new BufferedReader(stream)) {
-      reader.lines()
-          .map(String::trim)
-          .filter(line -> !line.isEmpty() && !line.startsWith("#") && !line.startsWith("!"))
-          .map(line -> line.split(EQUALS))
-          .forEach(splits -> inputConfigSettings.put(splits[0].trim(), splits.length > 1 ? splits[1].trim() : ""));
+      List<String> lines = reader.lines()
+        .map(String::trim)
```
mathieucarbou-ibm, line 124: this global trim is problematic because it will remove the spaces at the start of continued indented lines. 

trim should only be used when testing for a line starting with # or ! or is blank, or when sanitizing properties once the parsing is done

dynamic-config/model/src/main/java/org/terracotta/dynamic_config/api/service/ConfigPropertiesTranslator.java
```diff
@@ -111,19 +112,32 @@ public Properties load(Path configFile) {
...
+        .filter(line -> !line.isEmpty() && !line.startsWith("#") && !line.startsWith("!"))
```
mathieucarbou-ibm, line 125:
```suggestion
        .filter(line -> {
            line = line.trim();
            return !line.isEmpty() && !line.startsWith("#") && !line.startsWith("!")
        })
```

dynamic-config/model/src/main/java/org/terracotta/dynamic_config/api/service/ConfigPropertiesTranslator.java
```diff
@@ -111,19 +112,32 @@ public Properties load(Path configFile) {
...
+        .collect(Collectors.toList());
+      Iterator<String> itr = lines.iterator();
+      String line = "";
+      while (itr.hasNext()) {
+        String thisLine = itr.next();
+        char c = thisLine.charAt(thisLine.length() - 1);
+        if (c == '\\') {
+          line += thisLine.substring(0, thisLine.length()-1).trim();
```
mathieucarbou-ibm, line 133: I wouldn't trim here, to keep the user input as much as possible
```suggestion
          line += thisLine.substring(0, thisLine.length()-1);
```

dynamic-config/model/src/main/java/org/terracotta/dynamic_config/api/service/ConfigPropertiesTranslator.java
```diff
@@ -111,19 +112,32 @@ public Properties load(Path configFile) {
...
+        } else {
+          line += thisLine;
+          String[] splits = line.split("=");
+          inputConfigSettings.put(splits[0], splits.length > 1 ? splits[1].trim() : "");
```
mathieucarbou-ibm, line 137: this is where I would trim: at the very end, when creating the properties
```suggestion
          inputConfigSettings.put(splits[0].trim(), splits.length > 1 ? splits[1].trim() : "");
```

dynamic-config/model/src/test/resources/tc-config.cfg
```diff
@@ -0,0 +1,82 @@
...
+l2.healthcheck.l2.socketConnectCount:5,\
+#\
+l2.healthcheck.l1.ping.enabled:true,\
+l2.healthcheck.l1.ping.idletime:5000,       \
+l2.healthcheck.l1.ping.interval:1000,\
+# comment3\
+l2.healthcheck.l1.ping.probes:3,\
+l2.healthcheck.l1.socketConnect:true ,\
+l2.healthcheck.l1.socketConnectTimeout:2,\
+l2.healthcheck.l1.socketConnectCount:5,\
```
mathieucarbou-ibm, line 43:
```suggestion
        \
        # comment3 \
        l2.healthcheck.l1.ping.probes:3, \
        l2.healthcheck.l1.socketConnect:true , \
        l2.healthcheck.l1.socketConnectTimeout:2, \
        l2.healthcheck.l1.socketConnectCount:5, \
```

Usually continued lines are indented and the backslash is separated from the previous characters by a space

mathieucarbou-ibm (changes requested):

dynamic-config/model/src/main/java/org/terracotta/dynamic_config/api/service/ConfigPropertiesTranslator.java
tmesic99: updated as suggested.

dynamic-config/model/src/main/java/org/terracotta/dynamic_config/api/service/ConfigPropertiesTranslator.java
tmesic99: Agreed.

dynamic-config/model/src/main/java/org/terracotta/dynamic_config/api/service/ConfigPropertiesTranslator.java
tmesic99: updated as suggested.

mobasherul-ibm (approved):

PR #1265, opened 2026-03-24
[5.11] TDB-20080 : Lock the DC configuration until the end of config sync

mathieucarbou-ibm:
This PR is a proposal implementation using a lock to queue access to any DC / Nomad API during the sync of the configuration.

This is an alternative implementation of https://github.com/Terracotta-OSS/terracotta-core/pull/1403 and https://github.com/Terracotta-OSS/terracotta-core/pull/1403 which @myronkscott and I prefer compared to the others because it strictly prevents access to the configuration when it is not available.

The config sync process can happen:
- at startup
- remotely from diagnostic port during a stripe activation (stripe addition)

Changes:

- renamed method names to be more correct
- added some javadoc
- moved the locking mechanism directly into `DynamicConfigNomadServer` because DC requires anyway single thread access to the Nomad system
- Introduce reserve() and release() to block the Nomad system


jhouserizer (approved):
the locking logic looks safe/correct to me.

dynamic-config/server/config-provider/src/main/java/org/terracotta/dynamic_config/server/configuration/DynamicConfigConfigurationProvider.java
```diff
@@ -223,14 +227,21 @@ public void sync(byte[] bytes) {
                 + "(please change the logging config to see more details)" + lineSeparator()
                 + "==============================================================================================================================================" + lineSeparator(),
             e.getMessage());
+        // first release before stopping
+        nomadServerManager.getNomadServer().release();
```
chrisdennis: Shouldn't this be in a finally block?

chrisdennis (changes requested):

myronkscott (approved):

chrisdennis (approved):

PR #1277, opened 2026-04-09
[5.11] TDB-20032: Fix Lock and Unlock commands

mathieucarbou-ibm:
This PR fixes several issues in the `lock` and `unlock` commands.

1. **Missing validation to check that the cluster is fully activated**. This validation was not there so it was possible to trigger a `lock` or `unlock` command on a non activated cluster (or partial) which would lead to a cryptic Nomad error for the user

2. **Missing validation about online nodes**: `lock` and `unlock` commands were allowed as long as actives were up. They were not checking the states of passive servers. This problems can cause a mismatch configuration under certain cases and a loss of append log entries in case of an unfortunate failover.

3. **Fixed a typo** in the user displayed message

Note: point 2 was solved already for `set` / `unset` commands (mutative operations on the append log) but the fix was not ported to `lock` and `unlock` commands, which are also mutative operations on the append log.

Also, like we did for `set` / `unset`, a **hidden** `-force` option was added to the `lock` and `unlock` commands exactly like it was also done for the `set` / `unset` commands in order to bypass this validation and still be able to force a `lock` or `unlock` if some passive nodes are down and if the user knows what he is doing.

### DIFF
--- a/dynamic-config/cli/api/src/main/java/org/terracotta/dynamic_config/cli/api/command/LockUnlockConfigAction.java
+++ b/dynamic-config/cli/api/src/main/java/org/terracotta/dynamic_config/cli/api/command/LockUnlockConfigAction.java
@@ -1,0 +1,70 @@
+/*
+ * Copyright Terracotta, Inc.
+ * Copyright IBM Corp. 2026
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.terracotta.dynamic_config.cli.api.command;
+
+import org.terracotta.diagnostic.model.LogicalServerState;
+import org.terracotta.dynamic_config.api.model.Cluster;
+import org.terracotta.dynamic_config.api.model.Node.Endpoint;
+import org.terracotta.inet.HostPort;
+
+import java.util.Map;
+
+public abstract class LockUnlockConfigAction extends RemoteAction {
+
+  protected HostPort node;
+  protected boolean force;
+
+  protected Map<Endpoint, LogicalServerState> onlineNodes;
+  protected Cluster cluster;
+
+  public void setForce(boolean force) {
+    this.force = force;
+  }
+
+  public void setNode(HostPort node) {
+    this.node = node;
+  }
+
+  @Override
+  public void run() {
+    cluster = getRuntimeCluster(node);
+    onlineNodes = findOnlineRuntimePeers(node);
+
+    boolean allOnlineNodesActivated = areAllNodesActivated(onlineNodes.keySet());
+    if (!allOnlineNodesActivated) {
+      throw new IllegalStateException("The cluster is not fully activated: a lock or unlock operation can only be performed on an activated cluster.");
+    }
+
+    // validate that all the online nodes are either actives or passives
+    ensureNodesAreEitherActiveOrPassive(onlineNodes);
+
+    ensureActivesAreAllOnline(cluster, onlineNodes);
+
+    if (requiresAllNodesAlive()) {
+      // Check passive nodes as well if the setting requires all nodes to be online
+      ensurePassivesAreAllOnline(cluster, onlineNodes);
+    }
+  }
+
+  protected boolean requiresAllNodesAlive() {
+    // By default, we require all nodes to be up to allow a DC change: the consensus system needs all nodes to be able to vote and prepare the change.
+    // This will also solve any potential partitioned config to happen (or loosing a change if a failover happens just after),
+    // in the case a nomad change is prepared at the same time a passive is synchronizing the DC configuration from active server.
+    // This can be overridden by the user if he knows what he is doing with: -force
+    return !force;
+  }
+}
--- a/dynamic-config/cli/api/src/main/java/org/terracotta/dynamic_config/cli/api/command/TopologyAction.java
+++ b/dynamic-config/cli/api/src/main/java/org/terracotta/dynamic_config/cli/api/command/TopologyAction.java
@@ -1,6 +1,6 @@
 /*
  * Copyright Terracotta, Inc.
- * Copyright IBM Corp. 2024, 2025
+ * Copyright IBM Corp. 2024, 2026
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -117,7 +117,6 @@ public abstract class TopologyAction extends RemoteAction {
     if (mustLock) {
       // if we need to lock the operation, lock it first
       createdLockToken = lock(destinationCluster, destinationOnlineNodes, LockTag.OWNER_PLATFORM, buildLockTag());
-      output.out("Config lock with token: " + createdLockToken);
       destinationCluster = getUpcomingCluster(destination);
     }
     // then we run the scale task. In case of error, we unlock, but only if we placed a lock
--- a/dynamic-config/cli/config-tool/src/main/java/org/terracotta/dynamic_config/cli/config_tool/parsing/LockConfigCommand.java
+++ b/dynamic-config/cli/config-tool/src/main/java/org/terracotta/dynamic_config/cli/config_tool/parsing/LockConfigCommand.java
@@ -1,6 +1,6 @@
 /*
  * Copyright Terracotta, Inc.
- * Copyright IBM Corp. 2024, 2025
+ * Copyright IBM Corp. 2024, 2026
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -35,6 +35,9 @@ public class LockConfigCommand extends Command {
   @Parameter(names = {"-lock-context"}, description = "Lock context", required = true)
   String lockContext;
 
+  @Parameter(names = {"-force"}, description = "Force the operation", hidden = true)
+  boolean force;
+
   @Inject
   public LockConfigAction action;
 
@@ -50,6 +53,7 @@ public class LockConfigCommand extends Command {
   public void run() {
     action.setNode(node);
     action.setLockContext(lockContext);
+    action.setForce(force);
 
     action.run();
   }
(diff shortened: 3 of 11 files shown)


dynamic-config/cli/api/src/main/java/org/terracotta/dynamic_config/cli/api/command/LockUnlockConfigAction.java
```diff
@@ -0,0 +1,71 @@
...
+      throw new IllegalStateException("The cluster is not fully activated: a lock or unlock operation can only be performed on an activated cluster.");
+    }
+
+    // validate that all the online nodes are either actives or passives
+    ensureNodesAreEitherActiveOrPassive(onlineNodes);
+
+    if (requiresAllNodesAlive()) {
+      // Check passive nodes as well if the setting requires all nodes to be online
+      ensurePassivesAreAllOnline(cluster, onlineNodes);
+    }
```
guptakshit, line 60: If PASSIVE goes offline after the above check and before the actual lock command is executed, would the behaviour be similar to the old code, meaning the same problem can still occur or is there always a window between checks?
or it can't happen since now PASSIVE will always be part of onlineNodes and prepare itself won't succeed

mathieucarbou-ibm: > If PASSIVE goes offline after the above check and before the actual lock command is executed, would the behaviour be similar to the old code, meaning the same problem can still occur or is there always a window between checks? or it can't happen since now PASSIVE will always be part of onlineNodes and prepare itself won't succeed

You are right, thanks!

I forgot to replace:

```
Map<Endpoint, LogicalServerState> allNodes = findRuntimePeersStatus(node);
    onlineNodes = filterOnlineNodes(allNodes);
```

with

```
onlineNodes = findOnlineRuntimePeers(node);
```

mathieucarbou-ibm: I also added : `ensureActivesAreAllOnline(cluster, onlineNodes);` which is also there is set/unset validation.

dynamic-config/cli/api/src/main/java/org/terracotta/dynamic_config/cli/api/command/LockUnlockConfigAction.java
```diff
@@ -0,0 +1,70 @@
...
+    cluster = getRuntimeCluster(node);
+    onlineNodes = findOnlineRuntimePeers(node);
+
+    boolean allOnlineNodesActivated = areAllNodesActivated(onlineNodes.keySet());
+    if (!allOnlineNodesActivated) {
+      throw new IllegalStateException("The cluster is not fully activated: a lock or unlock operation can only be performed on an activated cluster.");
+    }
+
+    // validate that all the online nodes are either actives or passives
+    ensureNodesAreEitherActiveOrPassive(onlineNodes);
```
devendra-ibm, line 53: Even if we ensure here that passive should be in standby, before executing the lock or unlock change, 
there could be a window here, where it's possible that prepared msg succeeded, but when it sends commit message, in between passive restarts and becomes passive-syncing state, then again we might loose nomad change commit message on passive right.
` if (!act(results, decider, messageSender, data)) {
      results.done(decider.getConsistency());
      return;
    }
    commitOrRollback(results, decider, messageSender);`

So the real ques is when we have set `message -> ExecutionStrategy.Location.BOTH,` then why commit msg didn't reach to passive, even though passive was in passive-syncing state, are we knowingly not sending the message.?, because in normal scenario CRUD messages gets replicated to passive even when it's passive-syncing state, then why nomad commit msg didn't reach.
though it's very rare after all these checks, but still it's possible.

mathieucarbou-ibm: Yes this fix does not prevent the problem from occurring but decreases its likelihood to happen.
The same behavior is implemented in set / unset.
The way nomad works, the prepare phase is going through diagnosis channel so there is nothing reliable about that.
So what we can do is some pre-validations and checks to display the right message to the user the best we can, fail early the best we can.

But if something wrong happens during the nomad process, in any case, there will be a repair to do and this is documented in the troubleshooting section in the TCDB doc of config-tool.

About the passive sync, there are 2 phases: config sync and data sync. When a passive starts it will first sync its config: that is where the append log is synced from active and where currently a message could be lost:

**with previous behavior**

- passive starts
- active agrees to prepare a lock or unlock change
- active see the passive and sends its configuration to sync
- passive received its configuration and applies the prepared change in its append log
- while the active receives the commit for the tx (the passive is not there in data sync)
- then passive syncs (data messages)
- then passive becomes standby with a prepared change that is not committed

**with new behavior in this PR**

The fixes in PR here will prevent any lock/unlock change to be prepared if servers are not active or passive.

If a server crashes just after the change is prepared, there are 2 cases:
- either a failover happens, in that case the commit phase will continue and the node which crashed will sync its config after restart
- or there is a disconnection and the nomad change is left prepared but not committed (In that case this is OK, all the servers have the same config state and a repair command can repair hat)
- or the commit phase can proceed because this was a passive that crashed and it will sync later on restart

devendra-ibm (approved):

dynamic-config/cli/api/src/main/java/org/terracotta/dynamic_config/cli/api/command/LockUnlockConfigAction.java
```diff
@@ -0,0 +1,70 @@
...
+import org.terracotta.inet.HostPort;
+
+import java.util.Map;
+
+public abstract class LockUnlockConfigAction extends RemoteAction {
+
+  protected HostPort node;
+  protected boolean force;
+
+  protected Map<Endpoint, LogicalServerState> onlineNodes;
```
chrisdennis, line 31: It feels wrong that these are instance variables. I know we only use actions once, but it seems they should be "stateless" and in principle be able to be run multiple times. There's also some small issues with the `super.run()` calls being required to check the preconditions - it's brittle, and too easily skipped in a subclass. Could we move toward something like this:
```java
@Override
public final void run() {
  Cluster cluster = getRuntimeCluster(node);
  Map<Endpoint, LogicalServerState> onlineNodes = findOnlineRuntimePeers(node);

  //preconditions here

  lockingOperation(cluster, onlineNodes);
}

protected abstract void lockingOperation(Cluster cluster, Map<Endpoint, LogicalServerState> onlineNodes);
```

dynamic-config/cli/api/src/main/java/org/terracotta/dynamic_config/cli/api/command/LockUnlockConfigAction.java
```diff
@@ -0,0 +1,70 @@
...
+import org.terracotta.diagnostic.model.LogicalServerState;
+import org.terracotta.dynamic_config.api.model.Cluster;
+import org.terracotta.dynamic_config.api.model.Node.Endpoint;
+import org.terracotta.inet.HostPort;
+
+import java.util.Map;
+
+public abstract class LockUnlockConfigAction extends RemoteAction {
+
+  protected HostPort node;
```
chrisdennis, line 28: I think these could both be private.

dynamic-config/cli/api/src/main/java/org/terracotta/dynamic_config/cli/api/command/TopologyAction.java
```diff
@@ -117,7 +117,7 @@ public final void run() {
     if (mustLock) {
       // if we need to lock the operation, lock it first
       createdLockToken = lock(destinationCluster, destinationOnlineNodes, LockTag.OWNER_PLATFORM, buildLockTag());
-      output.out("Config lock with token: " + createdLockToken);
+      output.out("Config locked with token: " + createdLockToken);
```
chrisdennis: The fact that we're manually aligning the logging here makes me wonder if this locking should be being achieved by instantiating and configuring a lock action somehow? The lack of that right now makes me wonder whether the preconditions here (when locking is required) match the necessary preconditions in the locking action - and what it means/whether it matters if they don't.

dynamic-config/cli/config-tool/src/main/java/org/terracotta/dynamic_config/cli/config_tool/parsing/LockConfigCommand.java
```diff
@@ -35,6 +35,9 @@ public class LockConfigCommand extends Command {
   @Parameter(names = {"-lock-context"}, description = "Lock context", required = true)
   String lockContext;
 
+  @Parameter(names = {"-force"}, description = "Force the operation", hidden = true)
+  protected boolean force;
```
chrisdennis: Is there a reason this one is protected, when the others are package-private/default?

chrisdennis (dismissed):

dynamic-config/cli/api/src/main/java/org/terracotta/dynamic_config/cli/api/command/LockUnlockConfigAction.java
mathieucarbou-ibm: That's how all the commands are made and run: they are statefull and subclasses inherit the states and behavior and need to call super();

That's the command pattern applied were commands could be inherited and reused.

They can be instanciated and used only once. **They are not stateless services.**

kept the same pattern.

dynamic-config/cli/api/src/main/java/org/terracotta/dynamic_config/cli/api/command/LockUnlockConfigAction.java
mathieucarbou-ibm: Same answer as before.

protected because these fields could be reused by inherite commands.

Same pattern applied in all other abstract commands also.

dynamic-config/cli/api/src/main/java/org/terracotta/dynamic_config/cli/api/command/TopologyAction.java
mathieucarbou-ibm: Indeed: fixed the duplication by moving the output line in the lock method

dynamic-config/cli/config-tool/src/main/java/org/terracotta/dynamic_config/cli/config_tool/parsing/LockConfigCommand.java
mathieucarbou-ibm: I think it depends who implemented the CLI: some fields are marked protected some not and when adding the force boolean I just copied it from another command to keep consistency and protected was there.

So there is no technical reason at all.

So I have aligned with the other fields.

guptakshit (approved):

PR #1279, opened 2026-04-21
Fix lock/unlock commands to consider relay nodes

mathieucarbou-ibm:
This PR filters out the relay nodes from the online nodes to contact for the Nomad TX when doing a lock and unlock.

This change fixed LockConfigWithRelayIT.

But this change also leads to a question... Should relay nodes be restarted after a lock and unlock command ? And if yes, what is the impact of having some nodes currently restarting when the lock/unlock was part of a sequence of DC operations ?

I think we can leave them as-is, since they are not part of any Nomad TX, they do not need to be locked.

### DIFF
--- a/dynamic-config/cli/api/src/main/java/org/terracotta/dynamic_config/cli/api/command/RemoteAction.java
+++ b/dynamic-config/cli/api/src/main/java/org/terracotta/dynamic_config/cli/api/command/RemoteAction.java
@@ -67,6 +67,7 @@ import java.time.Duration;
 import java.util.Collection;
 import java.util.Comparator;
 import java.util.EnumSet;
+import java.util.HashSet;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
@@ -671,6 +672,18 @@ public abstract class RemoteAction implements Runnable {
 
   protected final LinkedHashMap<Endpoint, LogicalServerState> filterOnlineNodes(Map<Endpoint, LogicalServerState> nodes) {
     return filter(nodes, (addr, state) -> !state.isUnknown() && !state.isUnreacheable());
+  }
+
+  protected final Collection<Endpoint> removeRelayNodes(Map<Endpoint, LogicalServerState> onlineNodes) {
+    final Collection<Endpoint> onlineRelayNodes = new HashSet<>(onlineNodes.size() / 3);
+    onlineNodes.entrySet().removeIf(entry -> {
+      if (entry.getValue().isRelay()) {
+        onlineRelayNodes.add(entry.getKey());
+        return true;
+      }
+      return false;
+    });
+    return onlineRelayNodes;
   }
 
   protected final <K> LinkedHashMap<K, LogicalServerState> filter(Map<K, LogicalServerState> nodes, BiPredicate<K, LogicalServerState> predicate) {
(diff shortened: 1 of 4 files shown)


mathieucarbou-ibm:
This PR is a fix for a test in https://github.com/Terracotta-OSS/terracotta-platform/pull/1251

devendra-ibm:
> This PR filters out the relay nodes from the online nodes to contact for the Nomad TX when doing a lock and unlock.
> 
> This change fixed LockConfigWithRelayIT.
> 
> But this change also leads to a question... Should relay nodes be restarted after a lock and unlock command ? And if yes, what is the impact of having some nodes currently restarting when the lock/unlock was part of a sequence of DC operations ?
> 
> I think we can leave them as-is, since they are not part of any Nomad TX, they do not need to be locked.

we're restarting relay nodes only after the unlock, and not after lock,  since relay nodes also need to sync it's append log.
as you said, even though relay nodes are restarting, since they are excluded from nomad txn, then doesn't matter.

dynamic-config/cli/api/src/main/java/org/terracotta/dynamic_config/cli/api/command/RemoteAction.java
```diff
@@ -673,6 +674,18 @@ protected final LinkedHashMap<Endpoint, LogicalServerState> filterOnlineNodes(Ma
     return filter(nodes, (addr, state) -> !state.isUnknown() && !state.isUnreacheable());
   }
 
+  protected final Collection<Endpoint> removeRelayNodes(Map<Endpoint, LogicalServerState> onlineNodes) {
+    final Collection<Endpoint> onlineRelayNodes = new HashSet<>(onlineNodes.size() / 3);
+    onlineNodes.entrySet().removeIf(entry -> {
+      if (entry.getValue().isRelay()) {
```
guptakshit, line 680: do we need onlineNodes.size()/3? because if there are multiple DR sites then relay nodes are dynamic and can vary in number

mathieucarbou-ibm: Yes I agree: this is just an optimization to come up with an initial potential number of relay nodes from start that will probably be enough 80% of the time, the logic being that to have a relay node we probably already have a stripe with 3 nodes at least: active, passive and relay so 1/3 is a relay.
The map will be resized automatically if more slots are needed.

devendra-ibm (approved):

mathieucarbou-ibm:
> > This PR filters out the relay nodes from the online nodes to contact for the Nomad TX when doing a lock and unlock.
> > This change fixed LockConfigWithRelayIT.
> > But this change also leads to a question... Should relay nodes be restarted after a lock and unlock command ? And if yes, what is the impact of having some nodes currently restarting when the lock/unlock was part of a sequence of DC operations ?
> > I think we can leave them as-is, since they are not part of any Nomad TX, they do not need to be locked.
> 
> we're restarting relay nodes only after the unlock, and not after lock, since relay nodes also need to sync it's append log. as you said, even though relay nodes are restarting, since they are excluded from nomad txn, then doesn't matter.

Ah yes you are right the restart is already handled as part of `unlockInternal`. So that's perfect! There is no more question then.

~~So I should add the followRestart code in the unlock command then!~~

followRestart si also done as part of the restartNodes, so all good 👍

guptakshit (approved):

mathieucarbou-ibm:
@guptakshit : will merge the PR but there seem to be a timing issue with one of your test: SetCommand2x3IT line 88.
Got 2 jobs / 4 failing on this same test: https://github.com/Terracotta-OSS/terracotta-platform/actions/runs/24724733138/job/72322561102?pr=1279#step:7:635.

PR #1283, opened 2026-04-29
TDB-20388: Fix misleading error message when no SecurityLogger instance is found

guptakshit:


mathieucarbou-ibm (approved):

devendra-ibm (approved):

dynamic-config/testing/support/src/main/java/org/terracotta/dynamic_config/test_support/DynamicConfigIT.java
```diff
@@ -859,6 +859,8 @@ protected TerracottaServer waitForNewActive(TerracottaServer... servers) {
 
   public static class InlineToolExecutionResult extends ToolExecutionResult implements OutputService {
 
+    private static final long serialVersionUID = 1L;
```
guptakshit, line 862: the parent class ToolExecutionResult has become Serializable

dynamic-config/testing/support/src/main/java/org/terracotta/dynamic_config/test_support/DynamicConfigIT.java
```diff
@@ -859,6 +859,8 @@ protected TerracottaServer waitForNewActive(TerracottaServer... servers) {
...
+
     private final InMemoryOutputService delegate = new InMemoryOutputService();
```
mathieucarbou-ibm, line 864: The class now being serializable is causing a spotbug failure because InMemoryOutputService cannot be serializable.... Maybe it should be transient ? Or we allow it to be null ? Or we create a default public constructor always initializing it ?

mathieucarbou-ibm (changes requested):

mathieucarbou-ibm (approved):

PR #1285, opened 2026-05-26
DC change proposal to support activating the Nomad System at runtime without restart

mathieucarbou-ibm:


dynamic-config/cli/api/src/main/java/org/terracotta/dynamic_config/cli/api/command/RemoteAction.java
```diff
@@ -178,11 +178,16 @@ protected final void restartNodes(Collection<Endpoint> newNodes) {
   }
 
   protected final void activateNodes(Collection<Endpoint> newNodes, Cluster cluster, Path licenseFile) {
+    activateNodes(newNodes, cluster, licenseFile, true);
+  }
+  protected final void activateNodes(Collection<Endpoint> newNodes, Cluster cluster, Path licenseFile, boolean restart) {
```
mathieucarbou-ibm, line 183: CLI for a replica node would call that with `restart == false`

dynamic-config/server/config-provider/src/main/java/org/terracotta/dynamic_config/server/configuration/startup/ConfigurationGeneratorVisitor.java
```diff
@@ -95,19 +96,26 @@ public StartupConfiguration generateConfiguration() {
...
+      // - so "nomadServerManager.getConfiguration()" is not empty anymore
+      Optional<NodeContext> config = nomadServerManager.getConfiguration();
+      if (unConfiguredMode || repairMode || (replicaMode && config.isEmpty())) {
+        return nodeContext.alone();
+      } else {
+        return nomadServerManager.getConfiguration().orElseThrow(() -> new IllegalStateException("Node has not been activated or migrated properly: unable find any committed configuration to use at startup. Please delete the configuration directory and try again."));
+      }
+    };
+
+    return new StartupConfiguration(nodeContextSupplier, unConfiguredMode, repairMode, replicaMode, classLoader, pathResolver, parameterSubstitutor, jsonFactory, server);
```
mathieucarbou-ibm, line 118: this refactoring is more correct since each time the topology is queried by core, the supplier is then called and returns the right topology based on either the config on disk or the topology that was used to start the server.

So in case of a replica becoming active, core will query settings and they will be returned from the startup topology as long as the node is in replica state and once activated at runtime, nomad system is boostrapped and config on disk is written, then `nomadServerManager.getConfiguration()` would return the activated topo but without the replication setting.

PR #1311, opened 2026-07-10
TDB-21425: Replace TimeBudget with ConnectTimeout for Voltron-aware timeout tracking

mathieucarbou-ibm:
Introduce `ConnectTimeout` as a replacement for `TimeBudget`.
The new class models the three states that Voltron understands for connection timeouts:

- `empty`   → "-1"   (try exactly once, no blocking wait)
- `infinite` → 0 ms (effectively infinite)
- `finite`  → positive millis; but if the budget has depleted to 0,
            returns "-1" rather than "0" to prevent Voltron from
            silently treating an exhausted budget as an infinite wait.

The class captures its start time at construction and exposes isDepleted() / remaining() so that a shrinking budget can be threaded through multiple connection attempts without being restarted.

TimeBudget is removed.  All callsites (DiagnosticServiceProvider and its implementations, NomadEntityProvider, DynamicTopologyEntityProvider, LeasedConnectionServiceImpl, voter Options, and integration-test support) are updated to carry ConnectTimeout end-to-end.


common/structures/src/main/java/org/terracotta/common/struct/ConnectTimeout.java
mathieucarbou-ibm, line 1: Please read the javadoc and intent of this class. It represents a connection timeout and embed the conversion voltron semantic (-1, 0 positive numbers) with the ability to call remain() to get a new ConnectTimeout. object with the remaining time.

dynamic-config/testing/support/src/main/java/org/terracotta/dynamic_config/test_support/DynamicConfigIT.java
```diff
@@ -124,7 +125,7 @@ public class DynamicConfigIT {
   // can be modified by subclasses to update the timeouts
 
   // We intentionally set a huge timeout for connection
-  protected Duration connectionTimeout = Duration.ofDays(1);
+  protected ConnectTimeout connectionTimeout = ConnectTimeout.infinite();
```
mathieucarbou-ibm, line 128: @chrisdennis : That is nice for CLI and testing. We can now pass infinite (which translates to `-connect-timeout 0ms`), which is really what we want by default for our test. 1 day was to simulate that.

lease/client/src/main/java/org/terracotta/lease/connection/BasicLeasedConnection.java
```diff
@@ -18,33 +18,36 @@
...
     LeaseMaintainer leaseMaintainer = LeaseMaintainerFactory.createLeaseMaintainer(connection, timeSource);
 
     Exception exception = null;
     try {
-      boolean leased = leaseMaintainer.waitForLease(timeBudget.remaining(TimeUnit.MILLISECONDS), TimeUnit.MILLISECONDS);
+      // ConnectTimeout is a positive number ? return it.
+      // ConnectTimeout is empty (-1) ? we don't wait for the lease - we expect it to be there already
+      // ConnectTimeout is infinite ? => Long.MAX_VALUE since this is a Java wait on a latch
+      long remaining = timeBudget.remaining().getValue().orElseGet(() -> timeBudget.isInfinite() ? Long.MAX_VALUE : 0);
+      boolean leased = leaseMaintainer.waitForLease(remaining, TimeUnit.MILLISECONDS);
```
mathieucarbou-ibm, line 50: That's the translation for the lease wait.
Before we had a bug: if someone passed 0 for connect timeout (which means infinite wait in core semantic, we did't wait for the latch).

lease/client/src/main/java/org/terracotta/lease/connection/BasicLeasedConnection.java
```diff
@@ -18,33 +18,31 @@
...
-
 
 public class BasicLeasedConnection implements LeasedConnection {
   private final static Logger LOGGER = LoggerFactory.getLogger(BasicLeasedConnection.class);
 
   private final Connection base;
   private final LeaseMaintainer leaseMaintainer;
 
-  public static BasicLeasedConnection create(Connection connection, TimeBudget timeBudget, TimeSource timeSource) throws ConnectionException {
+  public static BasicLeasedConnection create(Connection connection, long timeoutMillis, TimeSource timeSource) throws ConnectionException {
```
mathieucarbou-ibm, line 40: API was wrong: this should be the caller responsibility to manage the time budget and call this method with the reduced timeout value when it needs to call it.

guptakshit (commented):
I think using `Long.compareUnsigned(now, deadlineNanos) >= 0` to check for expiration is unsafe for System.nanoTime()

System.nanoTime() can return negative values

`The value returned represents nanoseconds since some fixed but arbitrary origin time (perhaps in the future, so values may be negative).`

If
`when` is -10s
`now` = -5s (is still negative after 5 seconds elapsed)
`deadLineNanos` = 10s is positive (e.g after a 20s timeout), 

Long.compareUnsigned will cause the budget to be expired instantly on the first check

```java
@Test
  public void finiteShouldNotExpireInstantly() {
    long when = -10_000_000_000L;
    TimeBudget budget = TimeBudget.finite(when, 20, TimeUnit.SECONDS);
    long now = when + TimeUnit.SECONDS.toNanos(5);
    assertThat("Budget should not be depleted", budget.isDepleted(now), is(false));
  }
  ```

File: common/structures/src/main/java/org/terracotta/common/struct/TimeBudget.java [hunk 4]
```diff
@@ -16,28 +16,487 @@
...
+      return this;
+    }
+    // deadlineNanos - now is a signed subtraction.  When the budget is not yet depleted
+    // (now < deadline unsigned), the signed difference is a correct positive value because
+    // the unsigned distance fits within Long.MAX_VALUE for any practical timeout duration.
+    // remainingTimeout will be truncated to the timeUnit
+    long remainingTimeout = timeUnit.convert(Math.max(0, deadlineNanos - now), TimeUnit.NANOSECONDS);
+    // but deadlineNanos keeps nano precision
+    long correctedDeadlineNanos = Math.max(now, deadlineNanos);
+    return new TimeBudget(remainingTimeout, timeUnit, Math.max(now, correctedDeadlineNanos));
```
prasanta-ibm: `Math.max()` is not necessary, can return `new TimeBudget(remainingTimeout, timeUnit, correctedDeadlineNanos)`

mathieucarbou-ibm:
> I think using `Long.compareUnsigned(now, deadlineNanos) >= 0` to check for expiration is unsafe for System.nanoTime()
> 
> System.nanoTime() can return negative values
> 
> `The value returned represents nanoseconds since some fixed but arbitrary origin time (perhaps in the future, so values may be negative).`
> 
> If `when` is -10 `now` = -5 (is still negative after 5 seconds elapsed) `deadLineNanos` = 10 is positive (e.g after a 20s timeout),
> 
> Long.compareUnsigned will cause the budget to be expired instantly on the first check
> 
> ```java
> @Test
>   public void finiteShouldNotExpireInstantly() {
>     long when = -10_000_000_000L;
>     TimeBudget budget = TimeBudget.finite(when, 20, TimeUnit.SECONDS);
>     long now = when + TimeUnit.SECONDS.toNanos(5);
>     assertThat("Budget should not be depleted", budget.isDepleted(now), is(false));
>   }
> ```

Ahah ;-) when is the java time in nano seconds as per the javadoc. It cannot be negative. I will add a validation for this field.

mathieucarbou-ibm:
@guptakshit @prasanta-ibm : thank you for your review. PR updated.

common/structures/src/main/java/org/terracotta/common/struct/TimeBudget.java
```diff
@@ -16,28 +16,495 @@
...
+   * @param timeout  the duration; must be {@code >= 0}
+   * @param timeUnit the unit of {@code timeout}; must not be {@code null}
+   * @return a new finite timeout
+   * @throws IllegalArgumentException if {@code when} or {@code timeout} is negative
+   * @throws NullPointerException     if {@code timeUnit} is {@code null}
+   */
+  public static TimeBudget finite(long when, long timeout, TimeUnit timeUnit) throws IllegalArgumentException {
+    if (when < 0) {
+      throw new IllegalArgumentException("when must be >= 0 (System.nanoTime() value)");
+    }
```
guptakshit: System.nanoTime can be negative from java doc, does it mean different in above?


public static long nanoTime()
Returns the current value of the running Java Virtual Machine's high-resolution time source, in nanoseconds. This method can only be used to measure elapsed time and is not related to any other notion of system or wall-clock time. The value returned represents nanoseconds since some fixed but arbitrary origin time (perhaps in the future, so values may be negative). The same origin is used by all invocations of this method in an instance of a Java virtual machine; other virtual machine instances are likely to use a different origin.

common/structures/src/main/java/org/terracotta/common/struct/TimeBudget.java
```diff
@@ -16,28 +16,495 @@
...
+   * @throws IllegalArgumentException if {@code now} is negative
+   */
+  public boolean isDepleted(long now) {
+    if (now < 0) {
+      throw new IllegalArgumentException("now must be >= 0 (System.nanoTime() value)");
+    }
+    if (isEmpty()) return true;
+    if (isInfinite()) return false;
+    return now >= deadlineNanos;
+  }
```
guptakshit, line 161: should it be ` return now - deadlineNanos >= 0;` ?

common/structures/src/main/java/org/terracotta/common/struct/TimeBudget.java
mathieucarbou-ibm: > System.nanoTime can be negative from java doc, does it mean different in above?

Crap I missed that!

common/structures/src/main/java/org/terracotta/common/struct/TimeBudget.java
mathieucarbou-ibm: > should it be ` return now - deadlineNanos >= 0;` ?

well, now, yes indeed.

common/structures/src/main/java/org/terracotta/common/struct/TimeBudget.java
mathieucarbou-ibm: @guptakshit : fixed. See here is the diff between your first review and now FYI: https://github.com/Terracotta-OSS/terracotta-platform/compare/d8db195836e028c042be222b80597c537f9df277..3525bdb

I added your test case.

I was wrongly using unsigned operations because I didn't pay attention to the fact that nanoTime was possibly a negative value.

mathieucarbou-ibm: @guptakshit : fixed and included your test case in the test class.

common/structures/src/main/java/org/terracotta/common/struct/TimeBudget.java
```diff
@@ -16,28 +16,490 @@
...
+   *       {@code max(0, deadline - now)} converted to this timeout's unit.</li>
+   * </ul>
+   *
+   * @param now the current time in nanoseconds, as returned by {@link System#nanoTime()}
+   * @return a {@link TimeBudget} holding the remaining time at {@code now}
+   */
+  public TimeBudget remaining(long now) {
+    if (isEmpty() || isInfinite()) {
+      return this;
+    }
```
guptakshit, line 199: It seems we can accidentally lose time precision when creating the remaining budget.

If this happens inside a retry loop, the budget can drain much faster. When it converts remaining time from nanos to units like seconds, it can round down the values.

For example 4.9 seconds left on a budget gets round down and looses 900 millis

```java
@Test
public void remaining_shouldNotDrainBudget() {
  long start = System.nanoTime();
  TimeBudget budget = TimeBudget.finite(start, 5, TimeUnit.SECONDS);
  long now = start + TimeUnit.MILLISECONDS.toNanos(100);
  TimeBudget remaining = budget.remaining(now);
  OptionalLong value = remaining.getValue(TimeUnit.NANOSECONDS);
  assertThat("Remaining budget shouldn't loose time", value.getAsLong(), is(4_900_000_000L));
}
```

mathieucarbou-ibm: Yes this is an "expected" design: it is because the unit that the user chose is kept: `TimeBudget.finite(start, 5, TimeUnit.SECONDS);` => means that the user wants unit in SECONDS.

The original unit is kept across objects but maybe this is a bad design and the unit should only be used for conversion at input and output.

mathieucarbou-ibm: I am changing the API... Conserving the timeunit is wrong and causes this truncation.

mathieucarbou-ibm: @guptakshit : thanks! rewrote the implementation to use nanos internally and added `finite_remaining_shouldNotDrainBudget`

Diff: https://github.com/Terracotta-OSS/terracotta-platform/compare/3525bdb1f8af0e8ee7473dd3757e40786d8c63dd..0380d08f06a636e862179405fe53bb4db8555507

common/structures/src/main/java/org/terracotta/common/struct/TimeBudget.java
```diff
@@ -16,28 +16,490 @@
...
+    // deadlineNanos - now: safe signed subtraction using the JDK-recommended nanoTime idiom,
+    // correct even when deadlineNanos has wrapped the signed long range.
+    // remainingNanos < 0 means depleted (consistent with isDepleted: now - deadlineNanos >= 0).
+    long remainingNanos = deadlineNanos - now;
+    // Store full nanosecond precision — no unit truncation.
+    long remainingTimeout = Math.max(0, remainingNanos);
+    // When depleted, set the corrected deadline to now so that isDepleted() on the returned
+    // budget returns true immediately (now - now == 0 >= 0).
+    long correctedDeadlineNanos = remainingNanos < 0 ? now : deadlineNanos;
+    return new TimeBudget(remainingTimeout, correctedDeadlineNanos);
```
guptakshit, line 209: when the `remainingNanos` become 0, should it return `EMPTY` (`if (remainingNanos <= 0) return EMPTY`) instead of a finite budget as checking `budget.isEmpty()` will evaluate false even though it's depleted?

common/structures/src/main/java/org/terracotta/common/struct/TimeBudget.java
```diff
@@ -16,28 +16,490 @@
...
+   * </ul>
+   *
+   * @param unit the unit to express the timeout value in; must not be {@code null}
+   * @return the timeout value in {@code unit}, or empty for non-finite timeouts
+   */
+  public OptionalLong getValue(TimeUnit unit) {
+    if (isEmpty()) return OptionalLong.empty();
+    if (isInfinite()) return OptionalLong.empty();
+    return OptionalLong.of(unit.convert(timeoutNanos, TimeUnit.NANOSECONDS));
   }
```
guptakshit, line 251: `getValue` currently returns fixed `timeOutNanos` set at construction and calculated when remaining is called, should it be dynamically calculated rather than returning a fixed/older value? or maybe we remove timeoutNanos and only keep deadlineNanos

common/structures/src/main/java/org/terracotta/common/struct/TimeBudget.java
```diff
@@ -16,28 +16,490 @@
...
+   * @return {@code true} for a finite timeout
+   */
+  public boolean isFinite() {
+    return !isEmpty() && !isInfinite();
+  }
+
+  /**
+   * Checks whether this budget is now depleted, using the current wall-clock time.
+   * <ul>
+   *   <li>If the budget is <em>empty</em>   &rarr; always returns {@code true}
```
guptakshit, line 125: seems java-doc mismatch we're using system.nanoTime but it refers wall-clock time

common/structures/src/main/java/org/terracotta/common/struct/TimeBudget.java
mathieucarbou-ibm: No — this would be wrong. EMPTY means "try once, don't check time at all" (Voltron -1). A depleted finite budget means "budget ran out" — it should still be finite with value 0, so toVoltronString() returns "-1" correctly and getValue() returns OptionalLong.of(0). Returning EMPTY would conflate two semantically different states. The Javadoc just needs to clarify this distinction.

common/structures/src/main/java/org/terracotta/common/struct/TimeBudget.java
mathieucarbou-ibm: No — getValue() is intentionally the configured/remaining snapshot value, not a live query. isDepleted() is the live query. The confusion is because timeoutNanos on a budget returned by remaining() already is the remaining value at the time remaining() was called. The Javadoc just needs to make this clearer.

=> I will rename getValue() in getTimeout() to make is clearer.

mathieucarbou-ibm:
@guptakshit : method names and jaavdoc cleaned up ;-)

guptakshit (approved):

jhouserizer (approved):

PR #1321, opened 2026-08-10
TDB-22586:Redesign and Implement DR Replica Pre-Configured Cluster in Dynamic Configuration.

devendra-ibm:


dynamic-config/testing/entity/src/main/java/org/terracotta/dynamic_config/test_support/provider/ReplicaSimulationServiceProvider.java
mathieucarbou-ibm, line 1: Why you didn't use the existing SimulationHandler which does exactly that and you can even improve it to handle more specific use cases ? Was it because of the missing rollback ? If so, you do not need to test it I think ? Prepare failures and commit failures are different cases that could lead to different resolution strategies. A rollback failure is similar to a commit failure and it is unit tested (can be repaired). A system test showing how you handle a prepare failure and commit failure is enough because a rollback failure is just like a commit failure: the repair command will automatically continue to process the rollback.

dynamic-config/cli/api/src/main/java/org/terracotta/dynamic_config/cli/api/command/ConfigurationMutationAction.java
```diff
@@ -164,20 +164,28 @@ public void run() {
       final Collection<Endpoint> onlineRelayNodes = removeRelayNodes(onlineNodes);
 
       // cluster is active, we need to run a nomad change and eventually a restart
+      MultiSettingNomadChange changes = getNomadChanges(updatedCluster);
+      if (isReplicaCluster(onlineNodes)) {
```
mathieucarbou-ibm: I would do the check `if (!changes.getChanges().isEmpty())` just after because I don't think it is useful to go further of the changeset is empty ? I don't even understand why we have this check and how could a changeset be empty...
If empty, we should even do nothing at all maybe ?

dynamic-config/cli/api/src/main/java/org/terracotta/dynamic_config/cli/api/command/LockUnlockConfigAction.java
```diff
@@ -53,11 +53,13 @@ public void run() {
...
 
-    ensureActivesAreAllOnline(cluster, onlineNodes);
+      ensureActivesAreAllOnline(cluster, onlineNodes);
+    }
 
-    if (requiresAllNodesAlive()) {
+    if (requiresAllNodesAlive() && !isReplicaCluster(onlineNodes)) {
       // Check passive nodes as well if the setting requires all nodes to be online
       ensurePassivesAreAllOnline(cluster, onlineNodes);
     }
```
mathieucarbou-ibm: lock/unlock being a normal dc op, why we don't treat replica nodes like active and passives  ? With `ensureNodesAreEitherActiveOrPassiveOrReplica()` for example ? If we allow DC changes over nodes in REPLICA and REPLICA-START states, then we can validate them like active and passive ones ?

dynamic-config/cli/api/src/main/java/org/terracotta/dynamic_config/cli/api/command/RemoteAction.java
```diff
@@ -220,6 +220,10 @@ private void syncNomadChangesTo(Collection<Endpoint> newNodes, NomadChangeInfo[]
     }
   }
 
+  protected final boolean isReplicaCluster(Map<Endpoint, LogicalServerState> onlineNodes) {
+    return !onlineNodes.isEmpty() && onlineNodes.values().stream().allMatch(s -> s == REPLICA || s == REPLICA_SUSPENDED);
```
mathieucarbou-ibm, line 224: what is a replica became and active server but not the other ones yet ? Because one was offline, was restarting, or whatever reason ?

What I mean is what is the rules to consider a cluster a replica ?

- when all online nodes are in replica or replica-start state ?
- when at least one online node is ?
- or something else ?

dynamic-config/cli/api/src/main/java/org/terracotta/dynamic_config/cli/api/command/RemoteAction.java
```diff
@@ -369,7 +384,11 @@ protected final void forceUnlock(Cluster destinationCluster, Map<Endpoint, Logic
 
   private void unlockInternal(Cluster destinationCluster, Map<Endpoint, LogicalServerState> onlineNodes, boolean force) {
     output.info("Trying to unlock the config...");
-    runConfigurationChange(destinationCluster, filter(onlineNodes, (endpoint, state) -> !state.isRelay()), new UnlockConfigNomadChange(force));
+    if (isReplicaCluster(onlineNodes)) {
+      runConfigurationChangeViaDiagnostic(onlineNodes, new UnlockConfigNomadChange(force));
+    } else {
+      runConfigurationChange(destinationCluster, filter(onlineNodes, (endpoint, state) -> !state.isRelay()), new UnlockConfigNomadChange(force));
+    }
     output.info("Config unlocked.");
```
mathieucarbou-ibm: I see this code duplication at many places (if ... else ...).
Wouldn't it be easier to just keep `runConfigurationChange()` like before and within `runConfigurationChange()` do the check ? 

Having to change so many caller sites like that is error-prone and harder to maintain. It would be cleaner to group that in `runConfigurationChange` or, if not possible, on an overloaded `runConfigurationChange()` method maybe ?

One reason more is that callers don't need to know the internal details: for example the activate or config change actions just want to run their nomad change, that's all. How it is run, through which channel, this is up to the com layers to decide. Actions should not have the responsibility to choose that.

dynamic-config/cli/api/src/main/java/org/terracotta/dynamic_config/cli/api/command/RemoteAction.java
```diff
@@ -750,6 +769,19 @@ protected final void ensureActivesAreAllOnline(Cluster cluster, Map<Endpoint, Lo
...
+    }
+    // replicas == list of current replica nodes in the runtime topology
+    List<String> replicas = onlineNodes.entrySet().stream().filter(e -> e.getValue().isReplica() || e.getValue().isReplicaSuspended()).map(Map.Entry::getKey).map(Endpoint::getNodeName).collect(toList());
+    // Check for stripe count. Whether there is a pending dynamic config change or not, the stripe count is not changing.
+    // The stripe count only changes in case of a runtime topology change, which is another case.
+    if (cluster.getStripeCount() != replicas.size()) {
+      throw new IllegalStateException("Expected 1 replica per stripe, but only these nodes are in replica state: " + toString(replicas));
+    }
+  }
+
```
mathieucarbou-ibm: We can have more than 1 replica per stripe from our last discussion right ? 

If so, the check is wrong and you'd better build a Map<String, Integer> for example that is counting the number of online replicas per stripe to check that all stripes have at least 1 replica, and also that the online server count is >= total map count.

Also, what happens when a node has transitioned but not yet other ones (because of any possible reasons) ?

dynamic-config/cli/api/src/main/java/org/terracotta/dynamic_config/cli/api/command/TopologyAction.java
```diff
@@ -99,7 +99,7 @@ protected void validate() {
       throw new IllegalArgumentException("Wrong destination endpoint: " + destination + ". It does not match any node in destination cluster: " + destinationCluster.toShapeString());
     }
 
-    if (destinationClusterActivated) {
+    if (destinationClusterActivated && !isReplicaCluster(destinationOnlineNodes)) {
       ensureNodesAreEitherActiveOrPassive(filter(destinationOnlineNodes, (endpoint, state) -> !state.isRelay()));
       ensureActivesAreAllOnline(destinationCluster, destinationOnlineNodes);
     }
```
mathieucarbou-ibm: same comment as for the lock/unlock: if we allow this for a replica cluster, we should validate that nodes are `ensureNodesAreEitherActiveOrPassiveOrReplica`, something like this.

Or:

if not replica cluster: `ensureNodesAreEitherActiveOrPassive()`
if replica cluster: `ensureNodesAreAllReplica`

Depending what the definition of a replica cluster is (all replica or at least one)

dynamic-config/cli/api/src/main/java/org/terracotta/dynamic_config/cli/api/command/TopologyAction.java
```diff
@@ -150,7 +150,11 @@ protected void execute() {
       onNomadChangeReady(nomadChange);
       output.info("Sending the topology change");
       try {
-        runTopologyChange(destinationCluster, filter(destinationOnlineNodes, (endpoint, state) -> !state.isRelay()), nomadChange);
+        if (isReplicaCluster(destinationOnlineNodes)) {
+          runConfigurationChangeViaDiagnostic(destinationOnlineNodes, nomadChange);
+        } else {
+          runTopologyChange(destinationCluster, filter(destinationOnlineNodes, (endpoint, state) -> !state.isRelay()), nomadChange);
+        }
       } catch (RuntimeException e) {
```
mathieucarbou-ibm: same comment - caller sites should not have the responsibility to switch the com : they should just call: `runTopologyChange`

dynamic-config/testing/entity/src/main/java/org/terracotta/dynamic_config/test_support/provider/ReplicaSimulationServiceProvider.java
devendra-ibm: We can’t use existing SimulationHandler as it’s based on entity and we can’t directly create entities on replica, as it’s only possible with replication, that’s why I did it this way, it’s not because of rollback, 
I tested rollback  just because it was straight forward after I intercepted NomadServer.

dynamic-config/cli/api/src/main/java/org/terracotta/dynamic_config/cli/api/command/ConfigurationMutationAction.java
devendra-ibm: This empty changeset was added because when we do license set then configuration can be empty and so the changeset, because we’re explicitly clearing the configuration in validate method and allowing the license upgrade via diagnostic in validate method itself. So now if now `ConfigurationMutationAction#run` method executed then changeset will be empty. to protect this `if (!changes.getChanges().isEmpty())` check was added.
But this is not needed now, as we have made the explicit check if configuration is empty then return immediately via below code,
```
 (configurations.isEmpty()) {
  // no remaining configuration changes left to process
  output.info("Command successful!");
  return;
}
```
 this configuration check was added later, So previously it made sense, but not now. if So, I’m removing this.

dynamic-config/cli/api/src/main/java/org/terracotta/dynamic_config/cli/api/command/RemoteAction.java
devendra-ibm: This can only possible when user is trying  to do replica-failover after disaster and one of the server became Active and others not. In this case neither it will be a replica cluster nor it will be normal cluster with Actives. So we wouldn’t allow DC change at this stage before fixing this. So basically it means cluster is in broken state now, and requires manual intervention or repair command  to fix this. (all replica server should transition to Active)

devendra-ibm: So there could be these scenarios possible during `replica-failover`.
**Case1:** One of the replica was down, when primary cluster is gone. 
then `replica-failover` will throw exception saying please restart that server and try again, So now all replica server will be either in `PASSIVE-REPLICA-START` or `PASSIVE-REPLICA`, and unset of the replica flag will be done. Unset of the servers which are in `PASSIVE-REPLICA-START` is completely fine, as after restart if it transition to `ACTIVE`, then no issue as current data-dir has the valid data, but if it didn’t, then current data-dir data is not valid, and it will stuck in `START-STATE` and manual intervention will be needed for this, as correct data resides in one of the terracotta.backup folder 
**Case2:** all Replica was in `PASSIVE-REPLICA` state when primary went down.
Run `replica-failover `command and unset replica will be done and replica cluster will promote to `ACTIVE` immediately. No restart required.
**Case 3:** when primary went down, Replica was in `PASSIVE-REPLICA` state, and then restarted before `replica-failover` command was run.
After restart, the node comes back in `PASSIVE-REPLICA-START` state because the relay (primary) is gone. However, the data directory is still valid and complete. Running `replica-failover` followed by a server restart will successfully promote it to `ACTIVE`.
**Case4:** Replica was in `PASSIVE-SYNCING` state when the primary went down.
If the relay was actively syncing data to the replica when the primary crashed, the sync is aborted mid-way. So that replica will be in `PASSIVE-SYNCING` state. then `replica-failover` will throw exception saying please restart the server and try again, The replica restarts into `PASSIVE-REPLICA-START` state but its current data directory contains partial, incomplete data So `replica-failover` will unset the replica flag, and when restarted this server again, this will be stuck in `START-STATE ` as current data-dir is dirty, So it require manual intervention to identify which `terracotta.backup` folder contains valid data and then restart the server with correct data, Now it will become `ACTIVE`.
 
So, for `replica-failover` command, every replica server will be either in PASSIVE-REPLICA-START or `PASSIVE-REPLICA` and if succeeded it will unset replica flag for all servers, and transition servers to Active which are in `PASSIVE-REPLICA`.

But if `replica-failover` command itself has partial failure then we can fix this using repair command.
1.  If prepare failed on one of the nodes, then rollback will be triggered and nodes will be again in accepting state , and failover command can be retried as replica flag will be untouched in case of partial prepare. 
2. if commit failed for few replica nodes, then repair command will trigger commit on failed nodes, and eventually their replica flag will be unset. and in this case some of the nodes will be in `ACTIVE `and for some, we need to identify the correct data and restart to transition to `ACTIVE` as explained above.

dynamic-config/cli/api/src/main/java/org/terracotta/dynamic_config/cli/api/command/RemoteAction.java
devendra-ibm: yeah, right.
I have moved replica check inside `runConfigurationChange`.

dynamic-config/cli/api/src/main/java/org/terracotta/dynamic_config/cli/api/command/LockUnlockConfigAction.java
devendra-ibm: Now, I am checking for replica as well via this method.
`ensureReplicasAreAllOnline(cluster, onlineNodes)`

dynamic-config/cli/api/src/main/java/org/terracotta/dynamic_config/cli/api/command/TopologyAction.java
devendra-ibm: I am validating nodes for replica and non-replica cluster using this method.
`ensureClusterIsReadyForChange(Cluster cluster, Map<Endpoint, LogicalServerState> onlineNodes, boolean requireAllNodesAlive)`

dynamic-config/cli/api/src/main/java/org/terracotta/dynamic_config/cli/api/command/TopologyAction.java
devendra-ibm: I have moved replica related config change method inside `runConfigurationChange`.

dynamic-config/cli/api/src/main/java/org/terracotta/dynamic_config/cli/api/command/RemoteAction.java
guptakshit: with pre-configured cluster we can only have at-most 1 replica per stripe (similar to 1 Active per stripe in a normal cluster). 
In future from core side there can be non-replica nodes (like passive-uninitialized) alongside a replica node in a stripe

dynamic-config/testing/entity/src/main/java/org/terracotta/dynamic_config/test_support/provider/ReplicaSimulationServiceProvider.java
mathieucarbou-ibm: The DC services are not bound to the entity system.
If you look at the entities, they are not used at all: it was wrongly implemented as entities to use the entity service provider as a way to wire system services instead of using a normal service provider.

I raised a PR to cleanup that and you should be able to reuse / adapt in this PR.

See: https://github.com/Terracotta-OSS/terracotta-platform/pull/1326

dynamic-config/cli/api/src/main/java/org/terracotta/dynamic_config/cli/api/command/ConfigurationMutationAction.java
mathieucarbou-ibm: I forgot about the license set indeed!

dynamic-config/cli/api/src/main/java/org/terracotta/dynamic_config/cli/api/command/RemoteAction.java
mathieucarbou-ibm: That's good to have all these use cases in this PR ! Do not forget to add them in the doc (DC troubleshooting section) : I think a lot of warnings have to be put in writing to make sure the user does not delete its data folder by mistake. 

The risk here is that since its source cluster is done, the only data he has know lies in the replica cluster, which, if in START-STATE or PASSIVE-REPLICA-START, requires a manual recovery. 

So extra doc / care must be put in order to make sure the user does not delete this server data by mistake.

devendra-ibm: yeah, we have a plan to put this as a troubleshooting doc for faillover-command.

dynamic-config/testing/entity/src/main/java/org/terracotta/dynamic_config/test_support/provider/ReplicaSimulationServiceProvider.java
devendra-ibm: As it's existing test entity issue, So can it be taken care of separately than this PR.
and I saw it's designed for map based Settings, and I found issues with non map based settings like unset the replica, even after this PR #1326 PR gets merged, I woudn't be able to use as is.

devendra-ibm: I have changed the replica simulation based on #1326 and removed ReplicaSimulationServiceProvider.
Here is the PR #1328

mathieucarbou-ibm (approved):

devendra-ibm:
@myronkscott please have a look once as we need to merge this PR to create DR tests PR in tc-enterprise.
as we haven't received any validation for tests yet, So want to have DR tests PR reviewed and validated if any change needed, we can address it now. also we don't know at this stage whether tests are good enough in tc-enterprise or we need to move some in TCDB, as long back it was discussed if tests suit are taking more time, then we might need to move some in TCDB repo as well.

devendra-ibm:
We're merging this, as we want to open DR tests for review.
if Myron wants to have a look, he can look still, we will address this at later point in time as any way we might need to change things when he will bring other servers along with replica in a stripe.

